Issue #6, opened 2022-03-26
Replace active TX list with high water mark


thorfour:
Instead of maintaining a list of active transactions since we only want to prevent read tearing, we should replace it with a high watermark instead. This will also obviate the need for a garbage collector on the active list

metalmatze:
Do you have a good resource on learning about the high watermark concept?

PR #8, opened 2022-03-27
6 high water mark

thorfour:
Addresses issue #6 

Replaces the active list with a high watermark.

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -68,6 +68,9 @@ func TestAggregate(t *testing.T) {
 	err = table.Insert(buf)
 	require.NoError(t, err)
 
+	// Ensure all transactions are completed
+	table.Sync()
+
 	engine := query.NewEngine(
 		memory.NewGoAllocator(),
 		db.TableProvider(),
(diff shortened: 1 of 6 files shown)


brancz (approved):
This is awesome! I think there are two things that would be useful additions (best as follow up PRs I think)

* Allow syncing until a certain transaction. Maybe `.SyncTx(tx)`, as then a writer can after submitting its write transaction wait to see if it succeeeds. This is important in order to tell a client whether their write succeeded or not.
* Allow timing out transactions (and maybe enforce an upper bound). As it stands, a transaction can block all following transactions. This is by design and good for throughput, but we also want to make sure we continue to make progress, so timing a transaction is a good tool for that.

With the above two it might start to make sense to introduce a struct for transactions and reference that struct from Granule parts as opposed to just the tx id, as then we can also set transactions to be aborted (just an implementation I thought of off the top of my head, there are probably many ways to implement this).

metalmatze:
Currently taking a look at these :)

metalmatze:
I've started adding `context.Context` to handle timeouts on a branch: https://github.com/polarsignals/arcticdb/compare/context-timeout
Feel free to pick it up from there. Further discussion is better in another Issue/PR.

PR #9, opened 2022-03-28
Add golangci-lint and fix issues

metalmatze:


metalmatze:
Not even sure how to fix the last two once right now. Will look at it again another day.

PR #10, opened 2022-03-28
Add distinct push down query optimization

### LINKED ISSUE #5: Distinct push down to use dictionaries
This is a pretty specific query optimization but one that happens to occur quite a bit, which is querying all available values of a column where the column is dictionary encoded. In the Parca case, that's listing the `__name__` label values. In these cases instead of returning the entire decoded column, we can just return the dictionary directly.

brancz:
Closes #5

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -1,6 +1,7 @@
 package pqarrow
 
 import (
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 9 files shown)


pqarrow/arrow.go
```diff
@@ -94,15 +104,54 @@ func contiguousParquetRowGroupToArrowRecord(
...
 ) (arrow.Record, error) {
 	s := rg.Schema()
-
 	children := s.ChildNames()
+
+	if len(distinctColumns) == 1 && filterExpr == nil {
+		// We can use the faster path for a single distinct column by just
+		// returning its dictionary.
+		fields := make([]arrow.Field, 0, 1)
+		cols := make([]array.Interface, 0, 1)
```
metalmatze, line 117:
```suggestion
		cols := make([]arrow.Array, 0, 1)
```
The Interface is just an alias to `arrow.Array`

PR #17, opened 2022-04-04
Add schema scans

brancz:


query/engine.go
```diff
@@ -9,71 +9,91 @@ import (
 	"github.com/polarsignals/arcticdb/query/physicalplan"
 )
 
-type Engine struct {
+type Engine interface {
+	ScanTable(name string) QueryBuilder
+	ScanSchema(name string) QueryBuilder
+}
```
metalmatze, line 15: Better to move this to where it's actually called.

metalmatze (commented):
Great work! 🎉

query/engine.go
brancz: agreed. done.

PR #18, opened 2022-04-05
Swap out append active table block at configurable size

brancz:


metalmatze (commented):
Verrry nice. It's easy to follow although quite a lot. I don't have anything major to comment on. I'd say let's go with it after you ran it for quite some time, and we'll see how it goes. 🚀

PR #19, opened 2022-04-05
Add golangci-lint config files and GitHub Actions

metalmatze:
Let's just add the configs for now and then improve one step at a time. 
Supersedes #9


metalmatze:
We agreed to merge this and have golangci-lint as an optional CI job in our heads until we fix it.

PR #24, opened 2022-04-05
Fix linting issues

metalmatze:


metalmatze:
I don't see why this is failing the Go test now... 🤔

PR #25, opened 2022-04-05
Move to go.uber.org/atomic for ease of use

metalmatze:
There are still some things left to finish and currently `TestAggregate` is panicking.


brancz:
I'm guessing this might have conflicts with #8, so best to coordinate with @thorfour

metalmatze:
Thanks for pointing it out. Yes, that's probably true!
Maybe if @thorfour starts using the go.uber.org/atomic package for the few occasions in the PR we might be fine? WDYT?

metalmatze:
Somewhere in the `func (l *PartList) Iterate(iterate func(*Part) bool)` loops endlessly now... I'll continue another other.
Although the library is nice, it's quite a PITA to construct all the types manually now... Guess that's due to introducing it later.

PR #27, opened 2022-04-06
query/physicalplan: Fix aggregations of nil values

brancz:
I believe this fixes the panic reported in https://github.com/polarsignals/arcticdb/issues/26


kakkoyun (approved):
LGTM. CI has something else to say though :)

brancz:
We are still in the process of fixing all lints, it appears to me that this didn't introduce any new ones so this is good to go then.

PR #29, opened 2022-04-06
Move table to use go.uber.org/atomic

metalmatze:


metalmatze:
Interesting. I don't see any failures on my end. Even without the cache.
```
go version go1.18 darwin/amd64
```

PR #30, opened 2022-04-06
Move part_list to use go.uber.org/atomic

metalmatze:


granule.go
```diff
@@ -37,8 +37,11 @@ type Granule struct {
 func NewGranule(granulesCreated prometheus.Counter, tableConfig *TableConfig, firstPart *Part) *Granule {
 	g := &Granule{
 		granulesCreated: granulesCreated,
-		parts:           &PartList{},
-		tableConfig:     tableConfig,
+		parts: &PartList{
```
brancz, line 40: This feels like it should have a constructor function now.

metalmatze:
Yes, agreed! 👍

PR #39, opened 2022-04-14
go.mod: Update apache/arrow to latest

brancz:
Now that https://github.com/apache/arrow/pull/12764 is merged, we can use the upstream again.


brancz:
Merging since this is a no-op, just changing from our fork to upstream that is now identical to the fork.

PR #40, opened 2022-04-15
Support AscendRange options

thorfour:
Usage of AscendRange during iteration if given pivot options.

Since we have a multi-column primary key, it performs an intersection of the given pivot row schema and the table schema. Using this intersection it performs a comparison using the overlapping sorted columns. If a sorted column doesn't exist in the pivot schema, it's considered equal since it isn't known if that granule is less than the pivot. This can be seen in use with the unit tests where it pivots around a sample type and only ends up iterating over a subset of the granules.

In Parca we'll want to rearrange our sorted columns to be able to leverage this better, with the timestamps being the second sorted column.

### DIFF
--- a/dynparquet/reader.go
+++ b/dynparquet/reader.go
@@ -30,9 +30,14 @@ func ReaderFromBytes(buf []byte) (*SerializedBuffer, error) {
 }
 
 func NewSerializedBuffer(f *parquet.File) (*SerializedBuffer, error) {
+	b := &SerializedBuffer{
+		Reader: parquet.NewReader(f),
+		f:      f,
+	}
+
 	dynColString, found := f.Lookup(DynamicColumnsKey)
 	if !found {
-		return nil, ErrNoDynamicColumns
+		return b, nil
 	}
 
 	dynCols, err := deserializeDynamicColumns(dynColString)
@@ -40,11 +45,8 @@ func NewSerializedBuffer(f *parquet.File) (*SerializedBuffer, error) {
 		return nil, fmt.Errorf("deserialize dynamic columns metadata %q: %w", dynColString, err)
 	}
 
-	return &SerializedBuffer{
-		Reader:  parquet.NewReader(f),
-		f:       f,
-		dynCols: dynCols,
-	}, nil
+	b.dynCols = dynCols
+	return b, nil
 }
 
 func (b *SerializedBuffer) ParquetFile() *parquet.File {
--- a/dynparquet/row.go
+++ b/dynparquet/row.go
@@ -1,6 +1,8 @@
 package dynparquet
 
 import (
+	"fmt"
+
 	"github.com/segmentio/parquet-go"
 )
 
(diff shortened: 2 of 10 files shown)


thorfour:
AscendGreaterOrEqual
image

Ascend
image

dynparquet/row.go
```diff
@@ -10,12 +12,53 @@ type DynamicRow struct {
 	DynamicColumns map[string][]string
 }
 
-func (s *Schema) RowLessThan(a, b *DynamicRow) bool {
+// Intersection returns the intersection of two schemas
+// This is useful when comparing two rows of different schemas to only compare the columns that intersect.
+func Intersection(a, b *parquet.Schema) *parquet.Schema {
```
brancz, line 17: This seems like a lot of trouble to go through to just use `.ChildNames()` directly on the result in the only place where this function is used. Why not just merge the two string slices?

table.go
```diff
@@ -123,6 +125,11 @@ func newTable(
 				Help:    "Size of batch inserts into table.",
 				Buckets: prometheus.ExponentialBuckets(1, 2, 10),
 			}),
+			insertDuration: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+				Name:    "row_insert_duration_ms",
```
brancz, line 129: metrics should be in base units, so this should be converted to seconds

dynparquet/reader.go
```diff
@@ -30,21 +30,23 @@ func ReaderFromBytes(buf []byte) (*SerializedBuffer, error) {
...
 
 func NewSerializedBuffer(f *parquet.File) (*SerializedBuffer, error) {
+	b := &SerializedBuffer{
+		Reader: parquet.NewReader(f),
+		f:      f,
+	}
+
 	dynColString, found := f.Lookup(DynamicColumnsKey)
 	if !found {
-		return nil, ErrNoDynamicColumns
```
brancz, line 35: I think I prefer empty dynamic columns than not having them at all. This has caught several bugs at this point, and I fear that gracefully accepting that there are no dynamic columns would hide bugs further.

brancz (commented):
The API of this doesn't seem right to me and I think this only gets us halfway to where we want to be.

If the caller needs to be able to figure out the pivot point, then the query engine should be able to do the same, the API of the query engine is for asking _what_ data the user is asking for, not how the query engine will do that.

> In Parca we'll want to rearrange our sorted columns to be able to leverage this better, with the timestamps being the second sorted column.

It's a trade-off. If we do this, then all of the memory savings that we get from the efficient encodings go out the window as RLE doesn't have effect anymore. The way I imagine this to work is that there needs to be an interaction between the filter physical plan and the table scan, where the table scan starts with the approach laid out in this PR and whenever there is a whole row-group that after filtering contains no interesting data we take an intersection of MAX of that row-group and the filters and start a new AscendLargerThanOrEqualTo, and repeat this until the end.

I think this interaction between the physical plans is fine to defer to a future PR, since this lays the groundwork, but the API should be fixed I believe.

thorfour:
> The API of this doesn't seem right to me and I think this only gets us halfway to where we want to be.
> 
> If the caller needs to be able to figure out the pivot point, then the query engine should be able to do the same, the API of the query engine is for asking _what_ data the user is asking for, not how the query engine will do that.
> 
> > In Parca we'll want to rearrange our sorted columns to be able to leverage this better, with the timestamps being the second sorted column.
> 
> It's a trade-off. If we do this, then all of the memory savings that we get from the efficient encodings go out the window as RLE doesn't have effect anymore. The way I imagine this to work with our current schema is that there needs to be an interaction between the filter physical plan and the table scan, where the table scan starts with the approach mostly laid out in this PR and whenever there is a whole row-group that after filtering contains no interesting data we take an intersection of MAX of that row-group and the filters and start a new AscendLargerThanOrEqualTo, and repeat this until the end.
> 
> I think this interaction between the physical plans is fine to defer to a future PR, since this lays the groundwork, but the API should be fixed I believe.

Yea I agree the API is garbage, I was less concerned about the API surface in this iterataion as I was the implementation. However you've pointed out the key flaw in this in that we'd need to sacrifice the storage savings. So I'll need to circle back on this and think about your description some more.

PR #45, opened 2022-04-23
Return result if left expr eval to false in AndExpr

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>

If the left expr evals to false then seems it is always false for `AndExpr`.
I am still new to the codebase so not sure if there is any edge case here...


brancz:
Absolutely correct!

brancz (approved):

PR #46, opened 2022-04-23
replace[dynparquet/scheme.go]:replace boolean map with empty structs …

mertakman:
https://dave.cheney.net/2014/03/25/the-empty-struct
The previous approach was generating allocating boolean values for creating string sets, a new approach is allocation free for map values.


brancz (approved):
Thank you so much for the awesome contribution!

PR #48, opened 2022-04-28
Filter whole granules during query

thorfour:
Moves column metadata up into the Granule layer. We maintain a list of min/max values for each column stored in a Granule. Then when iterating we can use the filter expr to determine if there's any valid data in the Granule, and skip reading it entirely if there isn't.

### DIFF
--- a/granule.go
+++ b/granule.go
@@ -258,7 +258,7 @@ func (g *Granule) minmaxes(p *Part) bool {
 				page := p.Values()
 				values := make([]parquet.Value, p.NumValues())
 				_, err = page.ReadValues(values)
-				if err != nil {
+				if err != nil && err != io.EOF {
 					return false
 				}
 
--- a/granule.go
+++ b/granule.go
@@ -252,54 +252,41 @@ func (g *Granule) minmaxes(p *Part) error {
 		for j := 0; j < numColumns; j++ {
 			columnChunk := rowGroup.Column(j)
 
-			pages := columnChunk.Pages()
-			for {
-				p, err := pages.ReadPage()
-				if err != nil {
-					if err == io.EOF {
-						break
-					}
-					return err
-				}
+			idx := columnChunk.ColumnIndex()
+			minvalues := make([]parquet.Value, 0, idx.NumPages())
+			maxvalues := make([]parquet.Value, 0, idx.NumPages())
+			for k := 0; k < idx.NumPages(); k++ {
+				minvalues = append(minvalues, idx.MinValue(k))
+				maxvalues = append(maxvalues, idx.MaxValue(k))
+			}
 
-				page := p.Values()
-				values := make([]parquet.Value, p.NumValues())
-				n, err := page.ReadValues(values)
-				if err != nil && err != io.EOF {
-					return err
-				}
-				if n == 0 {
-					break
-				}
-
-				// Check for min
-				min := findMin(values)
-				g.metadata.minlock.RLock()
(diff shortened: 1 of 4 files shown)


thorfour:
@brancz could use some eyes on how I'm using the filterexpr, because it's awful and I hate it, and there's got to be a better way.

metalmatze:
Very nice improvements. Are those metrics supposed to be gone?

thorfour:
> Very nice improvements. Are those metrics supposed to be gone?

What metrics?

granule.go
```diff
@@ -211,5 +238,130 @@ func (g *Granule) Less(than btree.Item) bool {
...
 // Least returns the least row in a Granule.
 func (g *Granule) Least() *dynparquet.DynamicRow {
-	return (*dynparquet.DynamicRow)(g.least.Load())
+	return (*dynparquet.DynamicRow)(g.metadata.least.Load())
+}
+
+// minmaxes finds the mins and maxes of every column in a part.
+func (g *Granule) minmaxes(p *Part) error {
+	f := p.Buf.ParquetFile()
+	numRowGroups := f.NumRowGroups()
```
brancz, line 247: The Parquet ColumnIndex seems like it already gets us most of the way here, no?

thorfour: Hmm I can't figure out how to actually extract all the correct values. I see the label values, but none of the non string values.

thorfour: Only problem I'm running into using the column index is we're truncating values. 
ex: `process_cpu_cpu_nanoseconds` is being stored as the index max as `process_cpu_cpu`

thorfour: Looks like it's a writer setting that we control. So I've changed it to use that setting to truncate during comparison.

brancz (approved):
Really nice work!

This needs a rebase, then lgtm (plus the compare comment)! (maybe you could squash the commits)

granule.go
```diff
@@ -211,5 +238,117 @@ func (g *Granule) Less(than btree.Item) bool {
...
+				minvalues = append(minvalues, idx.MinValue(k))
+				maxvalues = append(maxvalues, idx.MaxValue(k))
+			}
+
+			// Check for min
+			min := findMin(minvalues)
+			g.metadata.minlock.RLock()
+			val := g.metadata.min[rowGroup.Schema().Fields()[columnChunk.Column()].Name()]
+			g.metadata.minlock.RUnlock()
+			if val == nil || Compare(*val, *min) == 1 {
```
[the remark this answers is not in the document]
thorfour: Yea I see that method, but its completely not clear to me how I get from a Value to a Type that I could use Compare on?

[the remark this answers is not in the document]
brancz: I believe we get it from the schema, which we already access above anyways.

table.go
```diff
@@ -793,9 +800,140 @@ func addPartToGranule(granules []*Granule, p *Part) error {
...
+			case *scalar.String:
+				v = left.Value.(*scalar.String)
+			}
+		}
+
+		switch right := expr.Right.(type) {
+		case logicalplan.BinaryExpr:
+			switch expr.Op {
+			case logicalplan.AndOp:
+				rightresult := filterGranule(logger, right, g)
```
yeya24, line 848: This can be skipped if leftresult == false

brancz: You want to send a PR? (I know Thor is offline for the next few days)

yeya24: https://github.com/polarsignals/arcticdb/pull/52 Opened.

PR #52, opened 2022-05-08
check left expr result when filtering granule with binary and op

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>


brancz (approved):
Amazing! Thank you!

PR #53, opened 2022-05-11
Cancel queries

thorfour:
Thread context for queries to allow cancellation.


metalmatze (approved):
Very nice and straight forward! 🚀

PR #58, opened 2022-05-14
fix query schema not hard-coded

albertlockett:
fixes issue in query/engine.go where the schema passed to the physical plan builder was hard-coded to the sample schema.

It now gets the schema from the logical plan. A new method was added to the logcal plan to return the schema which traverses the tree to get the schema from the TableProvider on either TableScan or SchemaScan steps.

A new method was also added to the TableReader interface which returns the schema

### DIFF
--- a/query/engine.go
+++ b/query/engine.go
@@ -6,7 +6,6 @@ import (
 	"github.com/apache/arrow/go/v8/arrow"
 	"github.com/apache/arrow/go/v8/arrow/memory"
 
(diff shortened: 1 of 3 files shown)


brancz:
Nice! I think implementation looks good, just API wise I expected this method to return the schema of the result as opposed to the input schema. How about we call the method `InputSchema`?

albertlockett:
Makes sense @brancz  - method name has been updated

thorfour:
@albertlockett thank you so much for this contribution!

Issue #59, opened 2022-05-15
Problem when `splitGranule()` is aborted


ostafen:
Hi, all, while experimenting with the persistence issue (#35), I found a problem with the `splitGranule()` function defined in the `table.go` file.
To reproduce the problem, you can set the granule size to 4096 instead of 8192 inside function `Test_Table_Concurrency()`. The test will fail with an ouput similar to the following:

```bash
=== RUN   Test_Table_Concurrency
    table_test.go:495: 
        	Error Trace:	table_test.go:495
        	Error:      	Not equal: 
        	            	expected: 8000
        	            	actual  : 7880
        	Test:       	Test_Table_Concurrency
--- FAIL: Test_Table_Concurrency (6.08s)
```

The issue is caused by the fact that `splitGranule()' is called several time, and, because of the high number of concurrent transactions, the function is aborted. For example when reaching the following condition:
```go
if n < t.table.db.columnStore.granuleSize { // It's possible to have a Granule marked for compaction but all the parts in it aren't completed tx's yet
	t.abort(commit, granule)
	return
}
```
Now, when `splitGranule()` is called, a "Compacting" sentinel node is immediately prepended in the part list of the granule:
```golang
parts := granule.parts.Sentinel(Compacting)
```
If the function is aborted, this sentinel node is not cleared, and remains in the part list of the granule.
Because of this additional node being present, any subsequent iteration on the part list is stopped when reaching the sentinel node, because of the following condition inside function `Iterate()`:

```golang
// Iterate accesses every node in the list.
func (l *PartList) Iterate(iterate func(*Part) bool) {
	...
	if node.part == nil && node.sentinel != l.listType { // if we've encountererd a sentinel node from a different type of list we must exit
		return
	}
	...
}
```
From my understanding of the code, stopping the iteration when `node.sentinel != l.listType` makes sense only when `Iterate()` is used inside the `splitGranule()` function. However, transactions which want to iterate on the full part list should skip this condition.
If my guess is correct, then we could pass a flag to the `Iterate()` function to skip the condition under certain circumstances.

thorfour:
Thank you so much for this detailed bug report! I've opened a fix for this

PR #60, opened 2022-05-15
Fix 59 concurrency

thorfour:
This fixes the issue detailed in #59 

We now iterate over all parts in a part list when the listType is None which happens during queries of the table.
It also fixes the unit test to wait for the txPool to drain instead of assuming we know how many tx the system has.


brancz:
Nice! Just a side note, it would be nice if we could have a way to wait for a specific transaction ID to be released, so an inserting entity can simulate having “waited” for a transaction to have finished. 

Lgtm and please squash the last commit :)

Issue #54, opened 2022-05-11
Support aborting writes


thorfour:
Arcticdb will need to support aborting write transactions. 

Transactions either need to be rolled back or tombstoned such that reads don't contain aborted transactions.

brancz:
One thing that I would also add is that a writer needs to be able to observe somehow whether their transaction was aborted or not, as a writer might want to retry a transaction if it was aborted.

PR #65, opened 2022-05-16
Issue 54 abort writes

thorfour:
Write cancellation. Addresses #54 

On context cancellation we'll mark all parts that have been written with a tombstone transaction id (`math.MaxUint64`) which causes them to be ignored during reads. They'll eventually be reaped during compaction and not be copied into the next granules.


brancz:
My bad, I created a conflict.

PR #67, opened 2022-05-17
Write function to take 'rows' and convert them to parquet

thorfour:
Took a shot at simplifying the writing experience with Arctic. 

Instead of requiring the writer to define a buffer and a bunch of parquet primitives we instead infer all of that from the input and the table schema.

### DIFF
--- a/table.go
+++ b/table.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"sync"
 	"unsafe"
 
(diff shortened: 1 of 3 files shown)


table.go
```diff
@@ -183,6 +184,71 @@ func (t *Table) Sync() {
...
+
+// Write a set of rows as a single buffer into the table
+func (t *Table) Write(rows []Row) (uint64, error) {
+
+	dynCols := t.config.schema.DynamicColumns()
+
+	// Create the full list of dynamic columns
+	mapping := map[string][]string{}
+	for _, dc := range dynCols {
+		found := map[string]bool{}
```
yeya24, line 197: What about using struct{} here

PR #69, opened 2022-05-18
Replace array.Interface with arrow.Array

ohkinozomu:
`array.Interface` is deprecated and removed by [this commit](https://github.com/apache/arrow/commit/61655b34f14e23fff217c314f60bbee92fe7961f).


thorfour:
Awesome thank you!

PR #74, opened 2022-05-27
Upgrade parquet-go

brancz:
parquet-go recently broke its APIs to give users more control over memory management and allow batch reads and writes for more control of performance.

cc @thorfour


thorfour (approved):
LGTM!

PR #80, opened 2022-06-01
add validation of logical plan aggregations

albertlockett:


albertlockett:
I can also add some validation of the GroupExprs if desired.

It looks like whatever is added to GroupExprs is stringified and that string is used to match the column. E.g. Col(“column_name”) and Literal(“column_name”) will both work.

A BinaryExpr like, like for example ... 
```golang
logicalplan.BinaryExpr{
    Left:  logicalplan.Col("Column1"),
    Op:    logicalplan.RegExpOp,
    Right: logicalplan.Literal("hello.*"),
},
```
will be used to group by a column with name "Column1 =~ hello.*", which is a bit weird 🙃

Issue #71, opened 2022-05-23
Test_Table_Concurrency fails rarely


ohkinozomu:
##  OS
```
$ sw_vers
ProductName:	macOS
ProductVersion:	11.6.4
BuildVersion:	20G417
```
## Go version
```
$ go version
go version go1.18 darwin/amd64
```
## revision
latest(`ccf34f7bbb98fa1b6af19a0198b81f7b4cd1441e`)
## Output
```
$ go test -count 300 -run Test_Table_Concurrency -timeout 20h
--- FAIL: Test_Table_Concurrency (27.61s)
    --- FAIL: Test_Table_Concurrency/8192 (6.71s)
        table_test.go:516:
            	Error Trace:	table_test.go:516
            	Error:      	Not equal:
            	            	expected: 8000
            	            	actual  : 7990
            	Test:       	Test_Table_Concurrency/8192
--- FAIL: Test_Table_Concurrency (26.70s)
    --- FAIL: Test_Table_Concurrency/8192 (6.65s)
        table_test.go:516:
            	Error Trace:	table_test.go:516
            	Error:      	Not equal:
            	            	expected: 8000
            	            	actual  : 7990
            	Test:       	Test_Table_Concurrency/8192
FAIL
exit status 1
FAIL	github.com/polarsignals/arcticdb	8061.410s
```

brancz:
Nice find! Any clue where the culprit might be?

ohkinozomu:
Since the merge of [this PR](https://github.com/polarsignals/arcticdb/pull/66 ), this doesn't seem to reproduce anymore. I ran the same command as above several times and the test didn't fail.
I will continue to look into this for my interest, but I think this isssue can be closed.

thorfour:
I am still able to recreate this 
```
thor@thors-MacBook-Pro arcticdb % go test -count 300 -run Test_Table_Concurrency -timeout 20h
--- FAIL: Test_Table_Concurrency (3.92s)
    --- FAIL: Test_Table_Concurrency/1024 (0.83s)
        table_test.go:513:
            	Error Trace:	table_test.go:513
            	Error:      	Not equal:
            	            	expected: 8000
            	            	actual  : 6964
            	Test:       	Test_Table_Concurrency/1024
^Csignal: interrupt
FAIL	github.com/polarsignals/arcticdb	42.293s
thor@thors-MacBook-Pro arcticdb % git ls 1
[...]

PR #83, opened 2022-06-02
Fix race cond

thorfour:
This fixes the error with the `GranuleSplit` unit test reported in #71 
It was a matter of a non-deterministic split that was happening. If we just wait for the split before the final write we test the logic we're looking for.


brancz (approved):
Oooooo good catch!

PR #84, opened 2022-06-04
Fix for concurrency race

thorfour:
If we start a new transaction, reads that started after the last
committed write, but before the compaction wont be able to find those commited
reads. Instead copy the watermark as the tx id to allow those reads to
be able to read all the committed data.

This is should be the fix for #71

### DIFF
--- a/table.go
+++ b/table.go
@@ -490,8 +490,8 @@ func (t *TableBlock) splitGranule(granule *Granule) {
 		return
 	}
 
-	// Obtain a new tx for this compaction
-	tx, watermark, commit := t.table.db.begin()
+	// Use the latest watermark as the tx id
+	tx := t.table.db.tx.Load()
 
 	// Start compaction by adding sentinel node to parts list
 	parts := granule.parts.Sentinel(Compacting)
@@ -503,7 +503,7 @@ func (t *TableBlock) splitGranule(granule *Granule) {
 	// Convert all the parts into a set of rows
 	parts.Iterate(func(p *Part) bool {
 		// Don't merge uncompleted transactions
-		if p.tx > watermark {
+		if p.tx > tx {
 			if p.tx < math.MaxUint64 { // drop tombstoned parts
 				remain = append(remain, p)
 			}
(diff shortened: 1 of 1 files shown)


thorfour:
hmm I think this caused a different problem where we now read more data 
```
--- FAIL: Test_Table_Concurrency (4.62s)
    --- FAIL: Test_Table_Concurrency/1024 (1.10s)
        table_test.go:33: level=error msg="failed to delete granule during split"

        table_test.go:33: level=debug msg="read blocks" n=0

        table_test.go:523:
            	Error Trace:	table_test.go:523
            	Error:      	Not equal:
            	            	expected: 8000
            	            	actual  : 9663
            	Test:       	Test_Table_Concurrency/1024
            	```

**Comment by thorfour** [comment]
With the latest fix from @ostafen and this fix I'm no longer seeing any failures with the concurrency test.

**Comment by MadhavJivrajani** [comment]
Not sure if I'm missing something but I tried running the test again with the race detector enabled and I still get races detected:
```
gh pr checkout 84
```
`HEAD` is at `a3e6195e7beef1e77eff933d569555eb474772f7`
```
go test -race -run Test_Table_Concurrency
```
The logs can be found here: https://gist.github.com/MadhavJivrajani/fa69278be76cdf18143dcde533f0e193

**Comment by thorfour** [comment]
> Not sure if I'm missing something but I tried running the test again with the race detector enabled and I still get races detected:
> 
> ```
> gh pr checkout 84
> ```
> 
> `HEAD` is at `a3e6195e7beef1e77eff933d569555eb474772f7`
> 
> ```
> go test -race -run Test_Table_Concurrency
> ```
> 
> The logs can be found here: https://gist.github.com/MadhavJivrajani/fa69278be76cdf18143dcde533f0e193
> 
> Edit: This is probably unrelated to the bug being fixed by the PR tho

This is a new one, but definitely unrelated to this fix. Can you open a second issue with this information?

**Comment by MadhavJivrajani** [comment]
Created https://github.com/polarsignals/arcticdb/issues/86

File: table.go [hunk 1]
```diff
@@ -490,8 +490,8 @@ func (t *TableBlock) splitGranule(granule *Granule) {
 		return
 	}
 
-	// Obtain a new tx for this compaction
-	tx, watermark, commit := t.table.db.begin()
+	// Use the latest watermark as the tx id
```
brancz, line 493: This is safe to do because we're not actually adding new data, we're just compacting it, right?

Issue #86, opened 2022-06-05
Race condition in Test_Table_Concurrency


MadhavJivrajani:
I tried running the test again with the race detector enabled (on PR #84) and I get races detected:
```
gh pr checkout 84
```
`HEAD` is at `a3e6195e7beef1e77eff933d569555eb474772f7`
```
go test -race -run Test_Table_Concurrency
```
The logs can be found here: https://gist.github.com/MadhavJivrajani/fa69278be76cdf18143dcde533f0e193

This is probably unrelated to the bug being fixed by the PR #84 tho.

PR #87, opened 2022-06-05
.github/workflows: Run tests with race detection enabled

MadhavJivrajani:
Following https://github.com/polarsignals/arcticdb/issues/86, maybe it's a good idea to run unit tests with race detection enabled in the CI itself?


brancz:
I guess that failure is exactly why #86 was possible :P 

Thanks for finding this! I guess we'll merge this once we have a fix for #86.

brancz:
Awesome! Thank you for the original report and enabling this!

Issue #88, opened 2022-06-07
Parallelize query execution


brancz:
Currently, arcticDB's query execution is not parallelized, but we want to do that. There are well-known techniques such as [vulcano](https://paperhub.s3.amazonaws.com/dace52a42c07f7f8348b08dc2b186061.pdf) to generalize the parallelization of steps within a query. I think vulcano is a promising direction for arcticDB but I'd be happy for us to explore other possibilities as well.

albertlockett:
Hey @brancz I agree volcano seems like a good fit.

To add parallelization to articdb, I think we could add an implementation of the exchange operator described in the paper. We could have optimizers in the logicalplan to inject exchange operators in the appropriate places in the query tree.

I took a stab at implementing this. https://github.com/polarsignals/arcticdb/pull/90

The ExchangeOperator in this cases uses channels instead of IPC. The chan length is used to provide the backpressure that (whereas int he paper they describe using semaphores), and it spawns a number goroutines that are listening to provide the horizontal parallelism (instead of using multiple queues as the paper describes).

Does that sound like it's the right approach? Is it too simplistic?

Of course this code in the PR is still WIP, but it gives us something to iterate on at least. Looking forward to your feedback!

yeya24:
Compared to the Volcano's 1-tuple-a-time Iterator model, vectorwise batching iterator model used by MonetDB seems better for columnar storage https://www.cidrdb.org/cidr2005/papers/P19.pdf. This is used by TiDB and CRDB as well.

But since we are using Arrow underlying already maybe Volcano is good enough.

brancz:
I wasn't aware of this paper, thanks for sharing @yeya24! I've only cross-read the paper, but from what I can tell what @albertlockett proposed so far is actually in line with what's described in this paper, as it handles full `arrow.Record` frames, as opposed to 1-tuple-a-time like you, said.

About what Albert wrote:

> Does that sound like it's the right approach? Is it too simplistic?

I think what you proposed is very elegant and should work quite well. Only practice will truly tell, but I think the direction looks promising!

albertlockett:
Thanks @brancz, I'll move forward with this. I'll get the code cleaned up, add tests & error handling and do some profiling to verify if this approach improves performance and what level of parallelism makes sense.

albertlockett:
@brancz @metalmatze we've managed to get a good performance boost from this parallelism. I wrote a simple benchmark program (see below) and found out that this is about 50% faster with the new code path. 

[...]

PR #90, opened 2022-06-07
#88 execute queries in parallel

albertlockett:
https://github.com/polarsignals/arcticdb/issues/88#issuecomment-1148997165

This PR adds parallelism in the table iterator. After the rowgroups are collected, a number of goroutines are spawned to process the rowgroups in parallel by calling the passed iterator (e.g. the physical plan).

It might be the case that the iterator keeps internal state in a way that is not thread-safe (e.g. physical plan's HashAggregate). In this case, it makes sense for the iterator to be created per thread (i.e. one instance of physical plan per thread). To accommodate this, the interface to table iterator has changed to accept an interface `IteratorProvider` with a method `Iterator()` that generates the iterator.

For aggregation queries, the aggregations happen in parallel and a Finisher has been added to combine the results of the multiple aggregations.

It's also possible that the final query callback could now be called concurrently. A new parameter is added to the query builder called Options which has a field to control this behaviour.
```golang
query.NewEngine(memory.DefaultAllocator, db.TableProvider()).
  ScanTable("test_table").
  // NEW:
  Options(&query.Options{SyncCallback: false}) // default is 'true'
  Execute(context.Background(), func(ar arrow.Record) error {
    // controls if this function can execute is synchronized or concurrent
```


query/logicalplan/optimize.go
```diff
@@ -158,6 +163,36 @@ func (p *FilterPushDown) optimize(plan *LogicalPlan, exprs []Expr) {
...
+
+func (p *FilterExchange) optimize(plan *LogicalPlan, exprs []Expr) {
+	if plan.Filter != nil {
+		numWorkers := runtime.NumCPU() - 1
+		if numWorkers <= 0 {
+			numWorkers = 1
+		}
+		exchangePlan := LogicalPlan{
+			Exchange: &Exchange{
+				Parallelism:  numWorkers,
```
brancz, line 183: Should the degree of parallelism really be decided in the logical plan? To me, it feels like this should happen while building of the physical plan.

table_test.go
```diff
@@ -30,6 +30,14 @@ type testOutput struct {
 	t *testing.T
 }
 
+type tIteratorProvider struct {
```
brancz, line 33: I would have expected something like a typed function here instead. Something like:

```
type iteratorProviderFunc func(arrow.Record) error

func (f iteratorProviderFunc) Iterator() func(arrow.Record) error {
  return f
}
```

brancz (commented):
I think the general direction of this is good!

I'm not sure the `SyncCallback` option makes sense like this in the API. I think since the compilation of logical query plans to physical ones is deterministic, I think it is the callers responsibility to take care of synchronization if the generated physical query plan happens to call the callback concurrently. I think it would be ok for the physical plan to be self-describing whether that is the case, but I think that would be fine as a follow-up to this PR.

While I think the merging approach that's laid out works for hash aggregations, I don't think this approach is going to evolve well in the future. I think we need to make the merging part a first-class citizen in the physical query plan, not just part of the finisher.

Just a high-level thought, we want to be able to have queries like (this is just pseudo-code):

```
TableScan("table).
Filter(Col("xyz").Regex("abc.*")).
Aggregate("sum", "value", "xyz").
Filter(Col("sum(value)").LT(0))
```

Which physical plan should look something like this:

```
                                                                 ┌───────────┐  ┌───────────┐  ┌───────────┐
                                                                 │           │  │           │  │           │
                                                              ┌──┤ Hash      │◄─┤ Filter    │◄─┤ Table Scan│
                                                              │  │ Aggregate │  │           │  │           │
                                                              │  └───────────┘  └───────────┘  └───────────┘
                                                              │
┌───────────┐  ┌───────────┐  ┌───────────┐  ┌───────────┐    │  ┌───────────┐  ┌───────────┐  ┌───────────┐
│           │  │           │  │ Concurrent│  │           │    │  │           │  │           │  │           │
│ Finisher  │◄─┤ Filter    │◄─┤ Hash      │◄─┤ Merger    │◄───┼──┤ Hash      │◄─┤ Filter    │◄─┤ Table Scan│
│           │  │           │  │ Aggregate │  │           │    │  │ Aggregate │  │           │  │           │
└───────────┘  └───────────┘  └───────────┘  └───────────┘    │  └───────────┘  └───────────┘  └───────────┘
                                                              │
                                                              │  ┌───────────┐  ┌───────────┐  ┌───────────┐
                                                              │  │           │  │           │  │           │
                                                              └──┤ Hash      │◄─┤ Filter    │◄─┤ Table Scan│
                                                                 │ Aggregate │  │           │  │           │
                                                                 └───────────┘  └───────────┘  └───────────┘
```

In a system like this, we need to be able to distinguish which parts of the query plan can happen concurrently and which cannot, and build the physical query plan accordingly.

table_test.go
albertlockett: that would make it cleaner, I'll make that change

query/physicalplan/physicalplan.go
```diff
@@ -117,60 +143,114 @@ func (s *SchemaScan) Execute(ctx context.Context, pool memory.Allocator) error {
...
+				// insantiated, do it now
+				if _, ok := disticts[plan.Distinct]; !ok {
+					matchers := make([]logicalplan.ColumnMatcher, 0, len(plan.Distinct.Columns))
+					for _, col := range plan.Distinct.Columns {
+						matchers = append(matchers, col.Matcher())
+					}
+					disticts[plan.Distinct] = Distinct(pool, matchers)
+				}
+				// use same distinct in all threads
+				phyPlan = disticts[plan.Distinct]
```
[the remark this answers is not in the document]
albertlockett: I think given the implementation in this PR when the review was done, that does make sense. It would make it more consistent

table_test.go
albertlockett: @brancz I think that makes sense wrt to the options to sync the query, I'll remove that.

I think for the query that's posted above, could we not execute the filter of the aggregate results in parallel?
```golang
TableScan("table).
Filter(Col("xyz").Regex("abc.*")).
Aggregate("sum", "value", "xyz").
Filter(Col("sum(value)").LT(0))
```

```
       A           B               C               D             E                   F              G               H
=====================================================================================================================================     
┌───────────┐ ┌───────────┐                                                     ┌───────────┐  ┌───────────┐  ┌───────────┐
│           │ │           │                                                     │           │  │           │  │           │
│ Finisher  ├─┤ Filter    ├─┐                                                ┌──┤ Hash      │◄─┤ Filter    │◄─┤ Table Scan│
│           │ │           │ │                                                │  │ Aggregate │  │           │  │           │
└───────────┘ └───────────┘ │                                                │  └───────────┘  └───────────┘  └───────────┘
                            │                                                │
┌───────────┐ ┌───────────┐ │ ┌───────────┐  ┌───────────┐  ┌───────────┐    │  ┌───────────┐  ┌───────────┐  ┌───────────┐
│           │ │           │ │ │           │  │ Concurrent│  │           │    │  │           │  │           │  │           │
│ Finisher  ├─┤ Filter    ├─┼─┤ Exchange  │◄─┤ Hash      │◄─┤ Merger    │◄───┼──┤ Hash      │◄─┤ Filter    │◄─┤ Table Scan│
│           │ │           │ │ │           │  │ Aggregate │  │           │    │  │ Aggregate │  │           │  │           │
└───────────┘ └───────────┘ │ └───────────┘  └───────────┘  └───────────┘    │  └───────────┘  └───────────┘  └───────────┘
                            │                                                │
┌───────────┐ ┌───────────┐ │                                                │  ┌───────────┐  ┌───────────┐  ┌───────────┐
│           │ │           │ │                                                │  │           │  │           │  │           │
│ Finisher  ├─┤ Filter    ├─┘                                                └──┤ Hash      │◄─┤ Filter    │◄─┤ Table Scan│
│           │ │           │                                                     │ Aggregate │  │           │  │           │
└───────────┘ └───────────┘                                                     └───────────┘  └───────────┘  └───────────┘
```

I think the interesting challenge we'll have is coordinating when each step should call the next callback. For example, we wouldn't want to have step D (Concurrent Hash Aggregate), call it's next callback to pass the results on step C (Exchange) until it has seen every record that will be emitted by step E (Merger).

I think the concept of a Finish/Next Finish function on each step could be used for this. The Finish function can be called on physical plan when all records have been sent into the iterator. Steps that want to wait to emit results to the next step will call their nextCallback in their Finish function.

It might also be a strategy that's more maintainable than having the physical plan building one monolithic Finish.

Here's an example (pseudocode) of how this could work:

```golang
type PhysicalPlan interface {
    // NEW:
    Finish() error
    SetNextFinish(next func() error)
    // ...
}
```

```golang
Filter:
  Callback(record) error {
    // filter records
    nextCallback(record)
  }
  Finish() error {
    // filter doesn't have any special finish logic
    return nextFinish()
  }
```

```golang
HashAggregate:
    Callback(record) error {
        // same as current logic
    }
    Finish() error {
        // same as current logic.
        // then do ...
        return nextFinish()
    }

```

```golang
// Merger takes the results from multiple parallel streams and emits them serially
Merger:
    // waiting for all the previous steps to call Finish. Initialized in builder
    wg sync.WaitGroup{}
    
    // only one thread will call the next callback
    mtx sync.Mutex{} 

    Callback(record) error {
        // buffer record
        return nil
    }

    // this will get called from multiple threads, but only one thread waits
    Finish() error {
        wg.Done() // reduce wg w/ each call

        // if this is the waiting thread, wait for all records buffered then emit
        if ok := mtx.TryLock(); ok {
            defer mtx.Unlock();
            wg.Wait()
            for _, record := buffer {
                nextCallback(record) // return err if not nil etc. 
            }
        }
    }
```

```golang
// Exchange takes a serial stream of results and emits the concurrently
Exchange {
    eg ErrorGroup
    rgChan chan arrow.Record

    Callback(record) error {
        rgChan <- record
        return nil
    }

    SetNextCallback (cb) {
        eg.Go(func() error {
            for rec <- rgChan {
                cb(rec) // return err if not nil
            }
        })
    }

    Finish() error {
        close(rgChan)
        eg.Wait() // return error if not nil
        nextCallback()
    }
}
```

brancz: I think the overall strategy is sound here, I agree it's still pretty akin to the exchange operator, but essentially exchange operator in a map/reduce style system. Something we haven't talked about too much is that we should be able to produce the graphs we keep drawing by hand as a result of the physical planning, so we can inspect what our software decided to do. We don't have to do this as part of this PR, but it's something we need to keep in mind.

albertlockett: Sounds good @brancz. Given that we want to draw the plan, maybe it makes sense for each step in the plan to have a reference to the next step itself, as opposed just some of it's methods. What do you think about that? 

i.e. instead of doing something:
```golang
// in the builder
phyPlan.SetNextCallback(prev.Callback)
phyPlan.SetNextFinish(prev.Finish)
```
... we do something like this:
```golang
// change the interface so plans reference the next physical plan
type PhysicalPlan interface {
	Callback(r arrow.Record) error
	Finish() error
	SetNextPlan(PhyisicalPlan)
}

// in the builder:
phyPlan.SetNextPlan(prev)
```
What's nice about doing it like this is that if we extend the PhysicalPlan interface with some Draw() method, it could be called recursively as we traverse the plan and build up the picture.

brancz: Yeah, I like that!

albertlockett:
@brancz the changes discussed in the comments above have been made. Would you mind reviewing again?

query/physicalplan/aggregate_concurrent.go
```diff
@@ -0,0 +1,276 @@
...
+	if numRows > 0 {
+		results = append(results, buildRecord())
+	}
+
+	return results
+}
+
+var nilTreeNode = struct{}{}
+
+// buildMergeTree creates a tree where each level of the tree is a column in
```
brancz, line 113: I don't think I really understand what's going on here. Since https://github.com/polarsignals/frostdb/pull/92 we have a predictable arrow schema in every record, so we can treat each record equally in terms of its columns.

I think there might have been a mismatch in communication. When I said we could build a tree to merge things, I meant to merge two whole records with each other at once.

Overall though I think we shouldn't complicate in the first version, and just combine records when the `Callback` is called, instead of when calling `Finish`.

query/physicalplan/distinct.go
```diff
@@ -14,7 +14,7 @@ import (
 
 type Distinction struct {
 	pool     memory.Allocator
-	next     func(r arrow.Record) error
+	nextPlan PhysicalPlan
```
brancz, line 17: Changing the callbacks to just be exactly the same as any `PhysicalPlan` is very clever and makes a ton of sense!

query/physicalplan/exchange.go
```diff
@@ -0,0 +1,51 @@
...
+			if err != nil {
+				return err
+			}
+		}
+		return nextPlan.Finish()
+	})
+}
+
+func (e *ExchangeOperator) Finish() error {
+	close(e.recordChan)
```
brancz, line 49: I don't think this follows the "only the producer closes the channel" pattern. I think we can create race condition here where the channel has been closed but the callback may still attempt to write to the channel.

query/physicalplan/exchange.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (e *ExchangeOperator) Callback(record arrow.Record) error {
+	record.Retain()
+	e.recordChan <- record
+	return nil
+}
+
+func (e *ExchangeOperator) SetNextPlan(nextPlan PhysicalPlan) {
+	// each time a next plan is added invoke a new goroutine to handle the records
+	e.eg.Go(func() error {
```
brancz, line 36: This doesn't seem right. We shouldn't start part of the query execution during planning. We might still end up optimizing the physical plan and reshuffling things, which may further exacerbate the problem.

thorfour:
Closing this since https://github.com/polarsignals/frostdb/pull/202 has merged. Thank you @albertlockett for your contributions! All your initial work helped to inform the final design!

albertlockett:
Very cool thanks nice work! @thorfour @metalmatze

PR #91, opened 2022-06-08
store the least row in each part

thorfour:
This fixes the race condition that was reported in #86 I'm not thrilled with this solution as it seems like there should be a better way to read the first row from the part without causing a data race.

### DIFF
--- a/dynparquet/row.go
+++ b/dynparquet/row.go
@@ -38,6 +38,10 @@ type DynamicRow struct {
 }
 
 func (s *Schema) RowLessThan(a, b *DynamicRow) bool {
+	if b == nil {
+		return true
+	}
+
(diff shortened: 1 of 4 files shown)


brancz:
I don't totally understand what's happening here. The stacktrace reads to me that there is an insert racing with a compaction trying to read a partial insert, this shouldn't happen in the first place. While this may fix the race we're detecting I don't think this fixes the actual semantic problem happening here. Am I missing something?

thorfour:
parquet `ReadRows` causes a write with `runtime.slicecopy()` which is where the write conflict comes from. I definitely don't understand how the compaction is causing a race with an insertion on the same buffer though

thorfour:
closing in favor of #96

PR #92, opened 2022-06-08
TableReader returns ArrowSchema for request

metalmatze:
It's helpful to return the computed schema for a request. 
The `TableReader` now has an `ArrowSchema` method that under the hood uses the same `ParquetRowGroupToArrowRecord` function that's used to generate the Arrow Records that now additionally returns the schema belonging to those records.

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -25,7 +25,7 @@ func ParquetRowGroupToArrowRecord(
 	projections []logicalplan.ColumnMatcher,
 	filterExpr logicalplan.Expr,
 	distinctColumns []logicalplan.ColumnMatcher,
-) (arrow.Record, *arrow.Schema, error) {
+) (arrow.Record, error) {
 	switch rg.(type) {
 	case *dynparquet.MergedRowGroup:
 		return rowBasedParquetRowGroupToArrowRecord(ctx, pool, rg)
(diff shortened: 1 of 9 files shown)


pqarrow/arrow.go
```diff
@@ -25,7 +25,7 @@ func ParquetRowGroupToArrowRecord(
 	projections []logicalplan.ColumnMatcher,
 	filterExpr logicalplan.Expr,
 	distinctColumns []logicalplan.ColumnMatcher,
-) (arrow.Record, error) {
+) (arrow.Record, *arrow.Schema, error) {
```
brancz, line 28: Am I missing something or can we already read the schema from the `arrow.Record`? If so I think we should just return the `arrow.Record` by itself and let the caller decide what to do.

table.go
```diff
@@ -949,7 +1010,7 @@ func (t *TableBlock) Serialize() ([]byte, error) {
 
 	// Read all row groups
 	rowGroups := []dynparquet.DynamicRowGroup{}
-	err := t.RowGroupIterator(ctx, nil, &AlwaysTrueFilter{}, true, func(rg dynparquet.DynamicRowGroup) bool {
+	err := t.RowGroupIterator(ctx, math.MaxUint64, nil, &AlwaysTrueFilter{}, func(rg dynparquet.DynamicRowGroup) bool {
```
brancz, line 1013: maybe put a comment on the use of `math.MaxUint64` here

brancz (commented):
This looks great! I do wonder whether at this point introducing a first call `Txn` struct would make sense, but I don't feel strongly about introducing it now. It feels inevitable though.

metalmatze:
> This looks great! I do wonder whether at this point introducing a first call `Txn` struct would make sense, but I don't feel strongly about introducing it now. It feels inevitable though.

Agreed, I was thinking the same, for now, it should be ok and we can improve down the road.

brancz:
lgtm 👍 🎉

brancz:
Nice work!!

PR #93, opened 2022-06-08
query/logicalplan: Better support for propagating DynamicColumns

### LINKED ISSUE #89: PhysicalProjectionPushDown needs to return dynamic columns
The projections that are created from the filters aren't correct when filtering by dynamic columns.

As shown in the screenshot below, the column is a `label.` prefix column and in this case, a dynamic column should be returned. Not because of its name, but because of some attribute that needs to get propagated into the optimizer. 
Screenshot from 2022-06-07 19-50-57

metalmatze:
The logicalplan should better propagate columns that are dynamic such that they can be treated differently in the future.

This PR builds on top of #92 

Fixes #89

### DIFF
--- a/query/logicalplan/optimize.go
+++ b/query/logicalplan/optimize.go
@@ -1,7 +1,11 @@
 package logicalplan
 
+import (
+	"github.com/polarsignals/arcticdb/dynparquet"
+)
+
 type Optimizer interface {
-	Optimize(plan *LogicalPlan) *LogicalPlan
+	Optimize(schema *dynparquet.Schema, plan *LogicalPlan) *LogicalPlan
 }
 
 var DefaultOptimizers = []Optimizer{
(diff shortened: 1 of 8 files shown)


query/logicalplan/optimize.go
```diff
@@ -17,19 +21,31 @@ var DefaultOptimizers = []Optimizer{
...
 		plan.SchemaScan.Projection = columnsUsed
 	case plan.TableScan != nil:
 		plan.TableScan.Projection = columnsUsed
 	case plan.Filter != nil:
-		columnsUsed = append(columnsUsed, plan.Filter.Expr.ColumnsUsed()...)
+		columns := plan.Filter.Expr.ColumnsUsed()
+		for _, column := range columns {
+			if dc, ok := schema.ColumnByName(column.Name()); ok {
+				if dc.Dynamic {
+					columnsUsed = append(columnsUsed, DynamicColumnMatcher{ColumnName: column.Name()})
```
brancz, line 40: Is there a specific reason why we're doing this? I don't understand what a filter of a dynamic column could look like, as a `DynamicColumnMatcher` is intended to match any of the dynamically created columns.

query/logicalplan/optimize_test.go
```diff
@@ -22,20 +22,21 @@ func TestOptimizePhysicalProjectionPushDown(t *testing.T) {
...
+		// duplicates because the statements of separate query plans just add
+		// the matchers for the columns they access. The optimizer could
+		// potentially deduplicate or use a more efficient datastructures
+		// in the future.
 		Projection: []ColumnMatcher{
 			StaticColumnMatcher{ColumnName: "stacktrace"},
 			StaticColumnMatcher{ColumnName: "stacktrace"},
 			StaticColumnMatcher{ColumnName: "value"},
-			StaticColumnMatcher{ColumnName: "labels.test"},
+			DynamicColumnMatcher{ColumnName: "labels.test"},
```
brancz, line 39: This doesn't look right to me. A `StaticColumnMatcher` is correct here because in the materialized result we intend to match on the `labels.test` column. A `DynamicColumnMatcher` is intended to match all `labels.*` columns. Like in an aggregation where we aggregate by all unique label combinations.

table.go
```diff
@@ -406,6 +407,63 @@ func (t *Table) SchemaIterator(
 	return err
 }
 
+func (t *Table) ArrowSchema(
```
brancz, line 410: I don't think this interface is going to work out. I think we should change the interface to read to something like:

```
func (t *Table) View(func(txn *Txn) error) error
```

Then based on that transaction the schema could be read within the read-only transaction. Then a consistent view of schema and data could be guaranteed.

query/logicalplan/optimize_test.go
metalmatze: Had to dig into why this needed to be changed again.
We have a projection for the `labels` column in some queries and for the `ArrowSchema` to be returned correctly we want to match all columns in the arrow schema that start with `labels.`. So yes in some cases we need to propagate the column as DynamicColumn. 
This is needed for the `includedProjection` in `arrow.go`.
Screenshot from 2022-06-09 15-24-10
The column `label` needs to be dynamically matched, so all columns starting with `labels.` are returned. Otherwise, the graphs will look like one series as no different label sets are returned.
Screenshot from 2022-06-09 15-26-00

metalmatze: Addition: Once querying a specific profile we know all columns exactly and the dynamic columns can then be treated as if they were StaticColumns since we know their full name like `labels.job` or `labels.instance`.

brancz: That makes sense, but in that case `logicalplan.DynCol("labels")` should have been used which consequently correctly uses the DynamicColumnMatcher as opposed to the StaticColumnMatcher.

brancz: Column names are sorted so the receiving side should not need to care about the concrete columns when querying, and instead merge them transparently.

metalmatze:
This PR isn't needed anymore.

PR #96, opened 2022-06-11
part: added least func

thorfour:
Move least of a part into a function. Lifted out all the calls in table where we fetched the least to use this helper function.

Also changed add part calls in addparttogranule to use the internal form that takes the least row so it doesn't need to be called again.

This addresses #86


brancz:
So if I'm understanding this right, it all boils down to parquet readers not being safe for concurrent reading?

thorfour:
> So if I'm understanding this right, it all boils down to parquet readers not being safe for concurrent reading (because the values buffer is concurrently reused)?

That's my understanding

brancz:
That makes sense. Happy with that!

PR #98, opened 2022-06-16
Add type to ParquetNodeToTypeWithWriterFunc log

ohkinozomu:
I was playing with the sample code and thought it would be easier to debug if the log showed the type of parquet.Node.

I also thought that Int64 was appropriate for the StorageLayout of value in the sample code.


brancz:
Whops yeah you’re right. Would be neat to support float64/doubles though as that’s what promtheus actually uses. Do you want to contribute that?

I’ll still merge this so the example that we have works, but would be cool to actually support the “right” type.

Issue #99, opened 2022-06-17
Support float64/double


ohkinozomu:
https://github.com/polarsignals/frostdb/pull/98#issuecomment-1157879825

PR #100, opened 2022-06-17
Switch by Kind instead of LogicalType in ParquetNodeToTypeWithWriterFunc

ohkinozomu:
I believe this change will make it easier to add types to the ParquetNodeToTypeWithWriterFunc.
This is a very small step for https://github.com/polarsignals/frostdb/issues/99.


brancz:
This doesn't feel quite right. The kind essentially describes the type that ultimately physically represents the data, but deciding the arrow type based on that seems incorrect as we might do an incorrect decision when a higher-level type is actually the right one.

ohkinozomu:
You're right...this affects the behavior.
I thought that adding test cases would prevent such a change, so I created another PR.
https://github.com/polarsignals/frostdb/pull/101

PR #104, opened 2022-06-23
pqarrow: Decouple Arrow Schema retrieval from Arrow Iterator

metalmatze:
Basically, we're now always creating an `arrow.Schema` from the given query and then passing that schema into the iterators building the actual `arrow.Record`. For the latter operation, the mapping becomes a  whole lot simpler since the arrow schema has a little index inside of it to do a lookup if a column should be included or not.
Performance isn't the main reason for this change though.

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -2,7 +2,6 @@ package pqarrow
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 
@@ -17,20 +16,20 @@ import (
 	"github.com/polarsignals/frostdb/query/logicalplan"
 )
 
-// ParquetRowGroupToArrowRecord converts a parquet row group to an arrow record.
-func ParquetRowGroupToArrowRecord(
+// ParquetRowGroupToArrowSchema converts a parquet row group to an arrow schema.
+func ParquetRowGroupToArrowSchema(
 	ctx context.Context,
 	pool memory.Allocator,
 	rg parquet.RowGroup,
 	projections []logicalplan.ColumnMatcher,
 	filterExpr logicalplan.Expr,
 	distinctColumns []logicalplan.ColumnMatcher,
-) (arrow.Record, error) {
+) (*arrow.Schema, error) {
 	switch rg.(type) {
 	case *dynparquet.MergedRowGroup:
-		return rowBasedParquetRowGroupToArrowRecord(ctx, pool, rg)
+		return rowBasedParquetRowGroupToArrowSchema(ctx, rg)
 	default:
-		return contiguousParquetRowGroupToArrowRecord(
+		return contiguousParquetRowGroupToArrowSchema(
 			ctx,
 			pool,
 			rg,
@@ -41,17 +40,12 @@ func ParquetRowGroupToArrowRecord(
 	}
 }
 
-// rowBasedParquetRowGroupToArrowRecord converts a parquet row group to an arrow record row by row.
-func rowBasedParquetRowGroupToArrowRecord(
-	ctx context.Context,
-	pool memory.Allocator,
-	rg parquet.RowGroup,
-) (arrow.Record, error) {
+func rowBasedParquetRowGroupToArrowSchema(ctx context.Context, rg parquet.RowGroup) (*arrow.Schema, error) {
 	s := rg.Schema()
 
 	parquetFields := s.Fields()
 	fields := make([]arrow.Field, 0, len(parquetFields))
-	newWriterFuncs := make([]func(array.Builder, int) writer.ValueWriter, 0, len(parquetFields))
+
 	for _, parquetField := range parquetFields {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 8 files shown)


pqarrow/arrow.go
```diff
@@ -82,10 +73,142 @@ func rowBasedParquetRowGroupToArrowRecord(
...
+	}
+	return false
+}
+
+// ParquetRowGroupToArrowRecord converts a parquet row group to an arrow record.
+func ParquetRowGroupToArrowRecord(
+	ctx context.Context,
+	pool memory.Allocator,
+	rg parquet.RowGroup,
+	schema *arrow.Schema, // TODO!
```
brancz, line 170: TODO what?

table.go
```diff
@@ -448,7 +463,26 @@ func (t *Table) ArrowSchema(
...
+				rg,
+				projections,
+				filterExpr,
+				distinctColumns,
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			record, err := pqarrow.ParquetRowGroupToArrowRecord(
```
brancz, line 478: why is just using the above schema not enough here?

pqarrow/arrow.go
```diff
@@ -82,10 +73,142 @@ func rowBasedParquetRowGroupToArrowRecord(
...
+		// returning its dictionary.
+		fields := make([]arrow.Field, 0, 1)
+		for i, field := range parquetFields {
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+				name := field.Name()
+				if distinctColumns[0].Match(name) {
+					typ, nullable, _, err := parquetColumnToArrowArray(
```
brancz, line 102: doing the entire array conversion seems awfully expensive just to get the schema

pqarrow/arrow.go
```diff
@@ -82,10 +73,142 @@ func rowBasedParquetRowGroupToArrowRecord(
...
+
+	fields := make([]arrow.Field, 0, len(parquetFields))
+
+	for i, parquetField := range parquetFields {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+			if includedProjection(projections, parquetField.Name()) {
+				typ, nullable, _, err := parquetColumnToArrowArray(
```
brancz, line 131: ditto

pqarrow/arrow.go
```diff
@@ -82,10 +73,142 @@ func rowBasedParquetRowGroupToArrowRecord(
...
+	}
+
+	// Create arrow writers from arrow and parquet schema
 	writers := make([]writer.ValueWriter, len(parquetFields))
-	b := array.NewRecordBuilder(pool, arrow.NewSchema(fields, nil))
+	b := array.NewRecordBuilder(pool, schema)
 	for i, column := range b.Fields() {
-		writers[i] = newWriterFuncs[i](column, 0)
+		parquetField := parquetFields[i]
+		_, newValueWriter, err := convert.ParquetNodeToTypeWithWriterFunc(parquetField)
```
brancz, line 207: It feels like we should have a function other than `ParquetNodeToTypeWithWriterFunc` now, that only returns what we truly need

pqarrow/arrow.go
```diff
@@ -17,20 +16,20 @@ import (
...
 	rg parquet.RowGroup,
 	projections []logicalplan.ColumnMatcher,
 	filterExpr logicalplan.Expr,
 	distinctColumns []logicalplan.ColumnMatcher,
-) (arrow.Record, error) {
+) (*arrow.Schema, error) {
 	switch rg.(type) {
 	case *dynparquet.MergedRowGroup:
-		return rowBasedParquetRowGroupToArrowRecord(ctx, pool, rg)
+		return rowBasedParquetRowGroupToArrowSchema(ctx, rg)
```
brancz, line 30: I don't see how for the schema row-based or contiguous makes a difference. The reason it makes a difference for the array conversion is that with contiguous ones, we know that the data to convert is all contained in a single contiguous page which is why we can exploit that nature for the conversion. The schema conversion should be identical to whether the underlying pages of column data is contiguous or not.

pqarrow/arrow.go
```diff
@@ -146,7 +268,7 @@ func contiguousParquetRowGroupToArrowRecord(
 			default:
 				name := field.Name()
 				if distinctColumns[0].Match(name) {
-					typ, nullable, array, err := parquetColumnToArrowArray(
+					_, _, array, err := parquetColumnToArrowArray(
```
brancz, line 271: now that we don't need the type and nullable property anymore we should probably separate those parts out of that function

table.go
metalmatze: Good point. It is and was an oversight, I guess. Updated.

pqarrow/arrow.go
metalmatze: Looking at the rest of the file only here we don't use the returned datatype. Everywhere else we do, so I'm fine having it stay like that.

brancz (approved):
Amazing!! Turns out it was also a lot less code now :D 

lgtm 🎉 🎉 🎉

PR #108, opened 2022-06-29
Fixes empty column EQ/NEQ/RE/NORE matcher.

cyriltovena:
This fixes the matchers when a column is missing.
There's definitively more improvements that could be done as the logic is sparse across the codebase.

However I wanted to have a first pass that works then think about how we can refactor.

I've tried to apply the matching as soon as possible and left a todo for the granule.

I think this fixes #1191 as this error is gone now.


cyriltovena:
Will do tomorrow sorry I left on a hurry

Issue #114, opened 2022-07-01
Unable to filter on something that isn't projected


thorfour:
Right now we wont filter on anything if the column we're filtering on isn't included in the projections.

So we're unable to perform queries equivalent to `select a,b from table where c > 10`
and instead require the query to be `select a,b,c from table where c > 10`

PR #115, opened 2022-07-01
Contionally add projection optimization

thorfour:
If we have a projection that doesn't include all columns
in a filter, we skip the projection optimization to avoid
the projection removing those columns before the filter can
happen.

This addresses issue #114

### DIFF
--- a/query/logicalplan/optimize.go
+++ b/query/logicalplan/optimize.go
@@ -60,6 +60,25 @@ func (p *PhysicalProjectionPushDown) optimize(plan *LogicalPlan, columnsUsed []C
 type ProjectionPushDown struct{}
 
 func (p *ProjectionPushDown) Optimize(plan *LogicalPlan) *LogicalPlan {
+
+	// Don't perform the optimization if the intersection of the columns affected is less than the number of columns filtered.
+	// Otherwise we'll removed the columns we're filtering on before we filter.
+	projectColumns := projectionColumns(plan)
+	projectMap := map[string]bool{}
+	filterColumns := filterColumns(plan)
+	intersection := map[string]struct{}{}
+	for _, m := range projectColumns {
+		projectMap[m.Name()] = true
+	}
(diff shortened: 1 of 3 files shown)


query/logicalplan/optimize.go
```diff
@@ -60,6 +60,25 @@ func (p *PhysicalProjectionPushDown) optimize(plan *LogicalPlan, columnsUsed []C
...
+		projectMap[m.Name()] = true
+	}
+	for _, m := range filterColumns {
+		if projectMap[m.Name()] {
+			intersection[m.Name()] = struct{}{}
+		}
+	}
+	if len(intersection) < len(filterColumns) {
+		return plan
+	}
```
MadhavJivrajani, line 80: If I understand correctly, we are trying to answer the question: Is it "safe" to push down a projection in the presence of a filter?

Instead of `len(intersection) < len(filterColumns)`, could we frame the logic to be "if the set difference `set(filterColumns) - set(projectColumns) > 0` we won't optimize? 
This could allow us to simplify the logic to something like:
```go
projectColumns := projectionColumns(plan)
filterColumns := filterColumns(plan)
projectMap := map[string]struct{}{}
for _, m := range projectColumns {
    projectMap[m.Name()] = struct{}{}
}

for _, m := range filterColumns {
    if _, ok := projectMap[m.Name()]; !ok {
        // There is something that is in filter but not in
        // project => non empty set difference, short circuit
        // and exit.
        return plan
    }
}
...
```
Maybe I've misunderstood something, please lmk if that's the case!

thorfour: No you're right. I overcomplicated that. Thanks!

brancz: I'll hold off on reviewing until after this is resolved then.

brancz (approved):
lgtm

Side note, I think we can still push the filter down and then the projection right after, no?

PR #118, opened 2022-07-04
Improves memory usage of ingest path.

cyriltovena:
```
❯ benchstat before.txt after.txt                                                                                                                                         
name                                       old time/op    new time/op    delta
_Table_Insert_10Rows_10Iters_10Writers-16     457ms ±11%     550ms ± 8%  +20.56%  (p=0.008 n=5+5)

name                                       old alloc/op   new alloc/op   delta
_Table_Insert_10Rows_10Iters_10Writers-16     577MB ±25%     169MB ±33%  -70.78%  (p=0.008 n=5+5)

name                                       old allocs/op  new allocs/op  delta
_Table_Insert_10Rows_10Iters_10Writers-16     1.06M ±36%     0.55M ±50%  -47.59%  (p=0.032 n=5+5)
```

Not yet sure why we're 20% slower, could be the sync.Map, nothing terrible is showing in  cpu profiling, but never the less, the memory usage is way better.


dynparquet/schema.go
```diff
@@ -431,6 +445,31 @@ func (s *Schema) NewWriter(w io.Writer, dynamicColumns map[string][]string) (*pa
 	), nil
 }
 
+type PooledWriter struct {
+	key string
+	*parquet.Writer
+}
+
+func (s *Schema) GetWriter(w io.Writer, dynamicColumns map[string][]string) (*PooledWriter, error) {
+	key := serializeDynamicColumns(dynamicColumns)
```
cyriltovena, line 454: Technically I need to also sort each suffix too at the risk of the cache not working, now sure if that's required or not that could be a big to do each time.

cyriltovena:
Latest commit seems to have improve CPU situation.

```
name                                       old time/op    new time/op    delta
_Table_Insert_10Rows_10Iters_10Writers-16     479ms ±12%     523ms ± 5%     ~     (p=0.095 n=5+5)

name                                       old alloc/op   new alloc/op   delta
_Table_Insert_10Rows_10Iters_10Writers-16     545MB ±19%     152MB ±31%  -72.01%  (p=0.008 n=5+5)

name                                       old allocs/op  new allocs/op  delta
_Table_Insert_10Rows_10Iters_10Writers-16     1.12M ±59%     0.37M ±15%  -67.42%  (p=0.016 n=5+4)
```

dynparquet/schema.go
```diff
@@ -363,14 +367,21 @@ func (s *Schema) NewBuffer(dynamicColumns map[string][]string) (*Buffer, error)
 	}, nil
 }
 
+var rowBufPool = &sync.Pool{
+	New: func() interface{} {
+		return make([]parquet.Row, 64) // Random guess.
+	},
+}
+
```
metalmatze, line 375: Does it make sense to have this initialized when calling `NewSchema` rather than having it globally in the package?

metalmatze:
Very nice! :tada:

dynparquet/schema.go
brancz: Not sure exactly if this is what you mean but keys are sorted, and values for keys are sorted, so there _should_ be no different keys with the same dynamic columns.

cyriltovena:
I used the sync.Map because it performs very well when you write only once, better than a RWMutex.

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) 

My only wish would be to avoid the empty pool creation required by the LoadOrStore.

PR #120, opened 2022-07-05
Avoid allocations when asking for schema fields.

cyriltovena:
[`RowLessThan`](https://github.com/polarsignals/frostdb/blob/450025ff363cb7241d9a038dcaf74b8f2779f2ac/dynparquet/row.go#L40) is part of the hot path for ingestion.

This function was using `schema.Fields()` from parquet for each iterations which creates a new slice every-time but it's always the same.

This PR actually set it once for the schema and flows it down into the dynamic rows.

image

I've been running this in a dev cluster and it's gone from the profile, however there's still improvement require for this function, we need to find a way to cache sorting columns per dynamic rows.


brancz (approved):
Nice!!

PR #126, opened 2022-07-14
dynparquet: Allow serializing/deserializing schemas with json

brancz:
@metalmatze @thorfour

### DIFF
--- a/dynparquet/schema/schema.go
+++ b/dynparquet/schema/schema.go
@@ -1,0 +1,86 @@
+package schema
+
+import (
+	"fmt"
+
+	"github.com/segmentio/parquet-go"
+	"github.com/segmentio/parquet-go/compress"
+	"github.com/segmentio/parquet-go/encoding"
+)
+
+type Definition struct {
+	Name           string             `json:"name"`
+	Columns        []ColumnDefinition `json:"columns"`
+	SortingColumns []SortingColumn    `json:"sorting_columns"`
+}
+
+type ColumnDefinition struct {
+	Name          string        `json:"name"`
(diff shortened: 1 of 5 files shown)


dynparquet/schema/schema.go
```diff
@@ -0,0 +1,86 @@
...
+type SortingColumn struct {
+	Name       string `json:"name"`
+	Order      string `json:"ordering"`
+	NullsFirst bool   `json:"nulls_first"`
+}
+
+func StorageLayoutToParquetNode(l StorageLayout) (parquet.Node, error) {
+	var node parquet.Node
+	switch l.Type {
+	case "string":
```
metalmatze, line 39: Can we reuse some sort of constants from other packages for these?

metalmatze (commented):
One question about using constants rather than strings, otherwise this is looking great!

dynparquet/schema/schema.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		node = parquet.Compressed(node, comp)
+	}
+
+	return node, nil
+}
+
+func encodingFromString(enc string) (encoding.Encoding, error) {
+	switch enc {
+	case "RLE_DICTIONARY":
```
thorfour, line 74: We have a lot of magic strings in this, 
RLE_DICTIONARY
ascending
descending
and then the types. Can we move all of these to a consts file?

brancz:
I think I've actually decided to go a different route and instead make this protobuf and it ends up being more intertwined with other work than I initially expected. Closing for now.

PR #128, opened 2022-07-17
Fix dynamic schema scan

thorfour:
What if we had schema scans actually build a schema instead of returning a single column of `name`?

This allows us to do things like schema scans with dynamic projections since our query behavior is based on column names. Otherwise to support something like this we'd need to have each expr know if it's apart of a schema scan to behave differently.

### DIFF
--- a/distinct_test.go
+++ b/distinct_test.go
@@ -198,3 +198,81 @@ func TestDistinctProjection(t *testing.T) {
 	t.Log(r)
 	require.Equal(t, int64(3), r.NumCols())
 }
+
+func Test_ScanSchema_DynamicProjection(t *testing.T) {
+	config := NewTableConfig(
+		dynparquet.NewSampleSchema(),
+	)
+
+	c := New(
+		nil,
+		8192,
+		512*1024*1024,
+	)
+	db, err := c.DB("test")
+	require.NoError(t, err)
+	table, err := db.Table("test", config, newTestLogger(t))
+	require.NoError(t, err)
+
+	samples := dynparquet.Samples{{
+		Labels: []dynparquet.Label{
+			{Name: "label1", Value: "value1"},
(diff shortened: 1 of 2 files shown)


metalmatze:
Very much agree. 
The schema and schema iterator are currently in a weird state. We still have the SchemaIterator but use the ArrowSchema a lot more. Maybe we should more broadly speaking look at these again and combine the interfaces?

brancz:
I was pretty sparse on the initial implementation but the idea was that the rows in the results are the columns of the schema, so an additional column of this result could for example be the type of the column. The whole point was that we could do things such as what you can do in postgres with `select * from information_schema.columns where table_name = 'stud1';`. The problem is though that we don't actually have a global schema so we will always need a layer that aggregates the potentially different sets of schemas, which is exactly what the distinct query plan does.

The idea was that the interface is that everything produces an `arrow.Record`, and schema scans are like accessing a different (meta-)table, therefore the expressions should not need to know anything special in regards to wether it's a tablescan or a schemascan.

thorfour:
> I was pretty sparse on the initial implementation but the idea was that the rows in the results are the columns of the schema, so an additional column of this result could for example be the type of the column. The whole point was that we could do things such as what you can do in postgres with `select * from information_schema.columns where table_name = 'stud1';`. The problem is though that we don't actually have a global schema so we will always need a layer that aggregates the potentially different sets of schemas, which is exactly what the distinct query plan does.
> 
> The idea was that the interface is that everything produces an `arrow.Record`, and schema scans are like accessing a different (meta-)table, therefore the expressions should not need to know anything special in regards to whether it's a table-scan or a schema-scan.

Yea I understood the intent, it just made more sense to me to actually return a schema, the types can be added to the actual schema (I just skipped that for now). Otherwise you need to know that there is a magic column returned called `name`.

PR #129, opened 2022-07-18
Add a parquet tool showing size of columns

cyriltovena:
see https://github.com/parca-dev/parca/issues/1309 for an example of the output.


brancz:
Awesome! Thank you!

PR #131, opened 2022-07-19
query: Allow recursive And expressions to be nil

brancz:
An extension of #130


brancz:
This is incorrect, the filter actually needs to be in the other function.

thorfour (approved):

PR #132, opened 2022-07-21
Add write-ahead log

brancz:
This adds a write-ahead log based on [tidwall/wal](). The whole design can be found in `design/wal.md`. As a side effect of having to replay schemas, we can now serialize and deserialize schemas using protobuf (and jsonpb).

The durability guarantees we get through this outweigh the minuscule performance overhead I think:

```
$ benchstat -delta-test=none ../old.txt ../new.txt
name                                          old time/op  new time/op  delta
_Table_Insert_1000Rows_10Iters_10Writers-10    603ms ± 0%   626ms ± 0%  +3.77%
_Table_Insert_100Rows_1000Iters_1Writers-10    569ms ± 0%   595ms ± 0%  +4.64%
_Table_Insert_100Rows_100Iters_100Writers-10   63.6s ± 0%   65.3s ± 0%  +2.75%
```

@metalmatze @thorfour


metalmatze (commented):
Veeery nice!

table.go
```diff
@@ -122,48 +145,45 @@ func newTable(
 		name:   name,
 		logger: logger,
 		mtx:    &sync.RWMutex{},
+		wal:    wal,
 		metrics: &tableMetrics{
 			blockRotated: promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
brancz: Not yet, but I was also thinking of doing it via the registry wrappers.

db.go
```diff
@@ -82,15 +140,60 @@ func (s *ColumnStore) Close() error {
 	return nil
 }
 
+func (s *ColumnStore) DatabasesDir() string {
+	return filepath.Join(s.storagePath, "databases")
+}
+
+func (s *ColumnStore) ReplayWALs(ctx context.Context) error {
```
thorfour, line 147: this function should have a comment

db.go
```diff
@@ -82,15 +140,60 @@ func (s *ColumnStore) Close() error {
...
+	dir := s.DatabasesDir()
+	if _, err := os.Stat(dir); err != nil {
+		if os.IsNotExist(err) {
+			level.Debug(s.logger).Log("msg", "WAL directory does not exist, no WAL to replay")
+			return nil
+		}
+		return err
+	}
+
+	files, err := ioutil.ReadDir(dir)
```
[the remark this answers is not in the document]
thorfour: What is dead may never die

db.go
```diff
@@ -82,15 +140,60 @@ func (s *ColumnStore) Close() error {
...
 	columnStore *ColumnStore
+	logger      log.Logger
 	name        string
 
 	mtx    *sync.RWMutex
 	tables map[string]*Table
 	reg    prometheus.Registerer
 
-	bucket objstore.Bucket
+	storagePath string
```
thorfour, line 194: `nit` would be nice to have these all in some settings/config child struct

brancz:
Ok so I made a mistake with the benchmarks, they didn't actually turn on the WAL.

Here are the benchmarks with the first commit of this PR (it couldn't even run all benchmarks):

```
$ benchstat -delta-test=none ../old.txt ../new.txt
name                                         old time/op  new time/op   delta
_Table_Insert_1000Rows_10Iters_10Writers-10   609ms ± 0%   2119ms ± 0%   +247.72%
_Table_Insert_100Rows_1000Iters_1Writers-10   540ms ± 0%  24732ms ± 0%  +4475.96%
```

The culprit is ultimately always the synchronization and actual writing of WAL records. Therefore I decided to relax the durability guarantee a bit and have records be written asynchronously in batches. Doing that makes the situation pretty much identical to what we had previously:

```
$ benchstat -delta-test=none ../old.txt ../new.txt
name                                          old time/op  new time/op  delta
_Table_Insert_1000Rows_10Iters_10Writers-10    609ms ± 0%   602ms ± 0%  -1.19%
_Table_Insert_100Rows_1000Iters_1Writers-10    540ms ± 0%   550ms ± 0%  +1.70%
_Table_Insert_100Rows_100Iters_100Writers-10   64.5s ± 0%   63.2s ± 0%  -2.09%
```

table.go
```diff
@@ -255,26 +354,55 @@ func (t *Table) InsertBuffer(ctx context.Context, buf *dynparquet.Buffer) (uint6
...
+	// Using active write block is important because it ensures that we don't
+	// miss pending writers when synchronizing the block.
+	block, close := t.ActiveWriteBlock()
+	if block.Size() > t.db.columnStore.activeMemorySize {
+		if err := t.RotateBlock(block); err != nil {
+			close()
+			return 0, fmt.Errorf("rotate block: %w", err)
+		}
+		close()
+		goto start
```
thorfour, line 388: I'd like to refactor this out, gotos create spaghetti

thorfour:
> 

If it's not going to really be a WAL then these should be consistency settings that users can tweak such that they could have a fully guarnteed wal

table.go
brancz: After playing with this code a bit, I think we can actually do this without the goto.

brancz:
> If it's not going to really be a WAL then these should be consistency settings that users can tweak such that they could have a fully guarnteed wal

Pretty much all WALs are best-effort, especially those that rely on POSIX fsync (which this one does as well), meaning the underlying file is flushed by the kernel, which can be up to 30s after the write was issued.

I would agree with allowing tweaking if the result of tweaking was in any shape or form reasonable, but exponential synchronization cost is never going to work, so I wouldn't want to allow that.

metalmatze (approved):

PR #135, opened 2022-07-24
Optimize combined distinct queries

brancz:
1) [Optimize distinct queries with single value per rowgroup](https://github.com/polarsignals/frostdb/commit/ce1959ed7e64d06d240400dd256315b49a546090):
Gathering distinct values of (non optional) dictionary encoded columns
is a constant operation per rowgroup when the dictionary only has a
single value. Additionally distinct projections using binary expressions
such as `timestamp > 0` can be determined without loading the row if the
column index has sufficient data. For example when the index notes that
all timestamps are between 5 and 8, then all values for `timestamp > 0`
are true, and therefore true for any distinct values in the rowgroup.

2) [pqarrow: Optimize distinct where all but one column are single value](https://github.com/polarsignals/frostdb/commit/872f3fcffed60a17989f579203c7a617e576e168):
When all but one column have only a single value, then the whole
rowgroup does not have to be scanned as the possible combination of
distinct values are the non-single column values combined with all
single value columns.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -158,7 +158,7 @@ func TestDBWithWAL(t *testing.T) {
 
 	pool := memory.NewGoAllocator()
 	err = table.View(func(tx uint64) error {
-		as, err := table.ArrowSchema(ctx, tx, pool, nil, nil, nil)
+		as, err := table.ArrowSchema(ctx, tx, pool, nil, nil, nil, nil)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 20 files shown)


brancz:
With these two optimizations profile type queries in [Parca](https://parca.dev) (which make heavy use of distinct queries) are about 94% better:

```
$ benchstat old.txt new.txt
name              old time/op  new time/op  delta
_ProfileTypes-10   760ms ± 7%    49ms ± 8%  -93.60%  (p=0.008 n=5+5)
```

This was a pretty extreme benchmark with _lots_ of data. A typical query is now <10ms, and Parca servers only ingesting a single type of profile (eg. only using the [Parca Agent](https://github.com/parca-dev/parca-agent) as it only ingests CPU profiles) even faster.

PR #136, opened 2022-07-25
Add parquet-reencode tool to switch encodings and compressions

brancz:
This tool is useful in order to try and identify whether a different
encoding or compression might be useful for a certain column. It allows
for a quick cycle to experiment without having to change the ingesting
application and wait for data to accumulate.


metalmatze:
Nifty :sunglasses:

PR #141, opened 2022-07-26
ReadAt storage

thorfour:
Uses GetRange to create an `io.ReaderAt` to not have to read the full file into memory to process it.


brancz:
Love it!

PR #144, opened 2022-07-28
Add support for Delta Byte Array and Delta Length Byte Array encoding

Pryz:
As I was working on a POC to use FrostDB as a backend to store otel traces, I had the need to use two encoding not supported by FrostDB yet. This PR adds support for both Delta Byte Array and Delta Length Byte Array encoding.


brancz (approved):
We'll just need you to sign the CLA (which is really just a one time DCO), then this is ready to merge!

thorfour:
Awesome! Thanks for the contribution!

PR #145, opened 2022-07-31
Cold start

thorfour:
Allows us to open up a database against a bucket storage and start querying data with the database reconstructing table definitions from the parquet file schemas.

Implements schema functionality to generate a `schemapb.Definition` from a parquet file.


dynparquet/schema.go
```diff
@@ -99,6 +101,15 @@ type Schema struct {
 	writers *sync.Map
 }
 
+// SortingColumns returns a slice of parquet.SortingColumn.
```
thorfour, line 104: I think this is unused

query/physicalplan/physicalplan.go
```diff
@@ -67,7 +67,7 @@ type TableScan struct {
 func (s *TableScan) Execute(ctx context.Context, pool memory.Allocator) error {
 	table := s.options.TableProvider.GetTable(s.options.TableName)
 	if table == nil {
-		return errors.New("table not found")
+		return fmt.Errorf("table not found")
```
thorfour, line 70: These fmt changes no longer necessary

db.go
```diff
@@ -261,6 +263,50 @@ func (s *ColumnStore) DB(name string) (*DB, error) {
 
 	db.txPool = NewTxPool(db.highWatermark)
 
+	ctx := context.TODO()
```
brancz, line 266: It doesn't seem unreasonable to me to add the context to the `DB` function parameters.

db.go
```diff
@@ -411,6 +457,16 @@ func (db *DB) Close() error {
 			return err
 		}
 	}
+
+	// If no persistence is enabled; we're done
+	if db.bucket == nil {
+		return nil
+	}
+
+	// Persist blocks to storage
```
brancz, line 466: I like where we're going with this, but it needs more thought I think. Essentially we're getting rid of all state at shutdown here, which is amazing! But that also means that if this is successful we can (and should) drop the write-ahead log if it's enabled, otherwise, we are very likely to duplicate all data in the WAL into object storage causing ourselves a ton of unnecessary deduplication work.

Additionally (this is more a follow-up), we should coordinate with the insertion layer that on shutdown we stop accepting writes, then wait for all active writers to finish, and then perform all of this. That way we maintain the highest chance for correctness and durability.

brancz (commented):
Very cool!

db.go
thorfour: Done

db.go
thorfour: Addressed dropping the WAL upon successful closing of tables

brancz (approved):

PR #148, opened 2022-08-02
update parquet-go: modify nilPage Data method to return encoding.Values

achille-roussel:
This PR adapts FrostDB to the changes introduced in https://github.com/segmentio/parquet-go/pull/289


brancz (approved):
Thank you so much!

PR #153, opened 2022-08-09
User metafile for persistence

thorfour:
I made a mistake trying to use the `data.parquet` files to determine the table schema. In the event we don't write all of the columns in a schema to the `data.parquet` files we'll be unable to retrieve the entire schema on reload which causes errors to be thrown when dynamic columns are written that don't exist in that reloaded schema.

~Instead this now writes a `schema.json` file into the tables directory, and loads that file on DB open instead.~

I've updated the PR to store schema files under a `schema` prefix. 
Ex: 
```
   db/table_name/schemas/ab2342cdwo912.json
    db/table_name/schemas/123940bvdaqfd.json
    db/table_name/01GA1F7HMQ4EWFC3VYRRB0JYEY/data.parquet
    db/table_name/01GA1F7HMQ4EWFC3VYRRB0EFCH/data.parquet
    db/table_name/01GA1F7HMQ4EWFC3VYRRB0ABCD/data.parquet
```
This prevents us from dropping a schema file every time we write a block. It names the schema files with a hash of the contents of the file to prevent duplicate schemas from being written. On startup a database scans all the tables `schema` prefixes for existing schema files and generates a table with those schemas, using the latest schema as the one utilized for writes.
This addresses the problem defined above, where we would potentially lose the schema of blocks already written.

This however does not address table migrations from one schema to an incompatible schema, but does set up the tables in such a way that we could read historical blocks with their historical schema.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2,16 +2,17 @@ package frostdb
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/oklog/ulid"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -221,6 +222,8 @@ type DB struct {
 }
 
 func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
+	name = SanitizeName(name)
+
 	s.mtx.RLock()
 	db, ok := s.dbs[name]
 	s.mtx.RUnlock()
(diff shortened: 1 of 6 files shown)


db.go
```diff
@@ -278,30 +279,27 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
...
+			if err != nil {
+				return fmt.Errorf("failed to get schema.json: %w", err)
 			}
+			defer r.Close()
 
-			f, err := parquet.OpenFile(b, attr.Size)
-			if err != nil {
-				return err
+			definition := &schemapb.Schema{}
+			if err := json.NewDecoder(r).Decode(definition); err != nil {
```
brancz, line 293: Should we be using the `jsonpb` encoder/decoder maybe? The advantage we get is that we can have the strinified version of enums that way as opposed to a plain number.

thorfour: yes, I always forget that's a thing

thorfour:
Thinking about this some more I'm realizing that the way things currently are; we can never actually migrate to a new schema, because the `Table` function in the db ignores the passed in schema if the table already exists. So if we've ever persisted a schema we'll need a different method to allow it to migrate to a new one. 

Maybe a migrate function that's intended to be called before any other routines act on the db/table

brancz (approved):

brancz:
If we do migrations they probably need a whole lot more thought anyways. So let's rather make progress at this point.

thorfour:
Closing in favor of #155

PR #155, opened 2022-08-10
Use block schemas

thorfour:
This obviates the need for the changes in #153 

- On `db.DB()` we'll now scan storage for table names, and open up read only tables for the discovered databases, allowing us to perform reads against those tables. 
- No longer use the table schema during queries, but instead use the row group schema
- on `db.Table()` if discovered we'll move a read only table into an active writable table

### DIFF
--- a/query/logicalplan/expr.go
+++ b/query/logicalplan/expr.go
@@ -258,8 +258,10 @@ func DynCol(name string) *DynamicColumn {
 
 func (c *DynamicColumn) DataType(s *parquet.Schema) (arrow.DataType, error) {
 	for _, field := range s.Fields() {
-		if field.Name() == c.ColumnName {
-			return convert.ParquetNodeToType(field)
+		if names := strings.Split(field.Name(), "."); len(names) == 2 {
+			if names[0] == c.ColumnName {
+				return convert.ParquetNodeToType(field)
+			}
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


query/logicalplan/expr.go
```diff
@@ -255,13 +256,14 @@ func DynCol(name string) *DynamicColumn {
 	return &DynamicColumn{ColumnName: name}
 }
 
-func (c *DynamicColumn) DataType(s *dynparquet.Schema) (arrow.DataType, error) {
-	colDef, found := s.ColumnByName(c.ColumnName)
-	if !found {
-		return nil, errors.New("column not found")
+func (c *DynamicColumn) DataType(s *parquet.Schema) (arrow.DataType, error) {
```
brancz, line 259: This doesn’t seem quite right. We won’t find an exact match, but we should still be able to use the parquet schema by just checking the prefix.

I’m not sure though what will happen if no column exists for a dynamic column, which can happen. Can we add a test for that?

brancz (commented):
I _think_ it should be possible to not error on the thing I commented on and just continue execution until there is a dynamic column instance. We would need to introduce a `nil,nil` case that callers would need to handle.

query/logicalplan/expr.go
thorfour: I fixed it to correctly match the dynamic schema columns. However, I don't think this code is actually reachable currently. We only call this in `pqarrow/arrow.go:72` which only happens if the distinct expr is computed which `DynamicColumn` isn't.

brancz (approved):
lgtm

Issue #156, opened 2022-08-10
Write table blocks directly to bucket


thorfour:
Right now in frostdb when we persist a table block into memory, we first read the entirety of the table block into a set of merged row groups, and then into a buffer. This causes our memory to explode every time we persist a block. This is seen the screenshot from Parca below.

![image](https://user-images.githubusercontent.com/8681572/184017079-e3f2e541-1249-41a2-9690-d13cc60d0694.png)
![](https://media.giphy.com/media/HP7mtfNa1E4CEqNbNL/giphy.gif)

It would be ideal if we could instead have the `TableBlock` implement the `io.Reader` interface that the `objstore.Bucket` interface requires so we could write it directly there instead of the memory spikes.

brancz:
It's kind of a shame that `objstore` requires an `io.Reader` when all it does is copy that `io.Reader` to an `io.Writer`. Parquet files can be written in a single contiguous write, so if we could instead get the writer from objstore we could refactor [`Serialize`](https://github.com/polarsignals/frostdb/blob/2f68e10c0065d36b9dc2b73dc32a824834b3c6c2/table.go#L1118) to take an `io.Writer` instead of writing to this in-memory buffer first:

https://github.com/polarsignals/frostdb/blob/2f68e10c0065d36b9dc2b73dc32a824834b3c6c2/table.go#L1137

brancz:
Ok, after looking ever so slightly into different providers, it seems that really only the GCS provider works the way I expected. I suspect the best way forward is to write the parquet file to disk and upload it from there. Feels really unnecessary, but I think it's the best we can do given all the circumstances.

brancz:
If we want to go the extra mile and avoid writing to disk at all cost we could propose an addition to the objstore API to get an `UploadWriter` which returns an `io.WriteCloser` and if a provider doesn't implement it we fall back to the filesystem based approach from above, but if they do support it we write directly to object storage.

thorfour:
So we can actually refactor the Serialize function today to take an `io.Writer` using the `io.Pipe`
```
    r, w := io.Pipe()
      var err error
      go func() {
          defer w.Close()
          err = t.Serialize(w)
      }()
      defer r.Close()

      fileName := filepath.Join(t.table.name, t.ulid.String(), "data.parquet")
      err2 := t.table.db.bucket.Upload(context.Background(), fileName, r)
      if err2 != nil {
          return err2
      }
```

[...]

PR #161, opened 2022-08-15
Unbuffered storage

thorfour:
This addresses the memory spikes seen in #156 

Instead of accumulating the block in a memory buffer we write it directly to objstore using the io.Pipe.
In addition instead of performing `readRows` on all merged rowgroups which allocates a large buffer of memory to do so, we instead only merged subslices of the rowgroups.

Can see in the screenshot there's no longer a spike in memory as we write to storage
image

### DIFF
--- a/table.go
+++ b/table.go
@@ -1144,7 +1144,7 @@ func (t *TableBlock) Serialize(writer io.Writer) error {
 
 	// Iterate over all the row groups, and write them to storage
 	count := 0
-	maxRows := 8096
+	maxRows := 8192
 	j := 0
 	for i, rg := range rowGroups {
 		count += int(rg.NumRows())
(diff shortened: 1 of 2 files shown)


table.go
```diff
@@ -1115,55 +1115,88 @@ func (t *TableBlock) abort(granule *Granule) {
...
 	if err != nil {
-		return nil, err
+		return err
 	}
 	defer t.table.config.schema.PutWriter(w)
+	defer w.Close()
+
+	// Iterate over all the row groups, and write them to storage
+	count := 0
+	maxRows := 8096
```
brancz, line 1147: I assume you meant 8192 (2pow13)

thorfour: lmao I have refactored enough times that it somewhere got typed wrong

brancz (approved):
lgtm, nice work!

PR #163, opened 2022-08-16
WAL replay fails when WAL dir is empty

arajkumar:
Signed-off-by: Arunprasad Rajkumar <ar.arunprasad@gmail.com>

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -273,6 +273,11 @@ func (w *FileWAL) Replay(handler func(tx uint64, record *walpb.Record) error) er
 		return fmt.Errorf("read last index: %w", err)
 	}
 
+	// FirstIndex and LastIndex returns zero when there is no WAL files.
+	if firstIndex == 0 || lastIndex == 0 {
+		return nil
+	}
+
 	level.Debug(w.logger).Log("msg", "replaying WAL", "first_index", firstIndex, "last_index", lastIndex)
 
 	for tx := firstIndex; tx <= lastIndex; tx++ {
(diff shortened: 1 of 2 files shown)


arajkumar:
cc @thorfour

wal/wal.go
```diff
@@ -273,6 +273,11 @@ func (w *FileWAL) Replay(handler func(tx uint64, record *walpb.Record) error) er
 		return fmt.Errorf("read last index: %w", err)
 	}
 
+	// FirstIndex and LastIndex returns zero when there is no WAL files.
+	if firstIndex == 0 || lastIndex == 0 {
```
thorfour, line 277: There's already a directory check in `db.go:166` I think we should just need to update that instead no?

arajkumar: db.go:166 actually checks for the existence of database directory not the wal. Also we may not able to check the existence of wal dir because creating a WAL instance[1] would always create a `wal` directory[2].

[1] https://github.com/polarsignals/frostdb/blob/363a6ad4c48da19af87f94ca0a6dcb1283e0e01b/db.go#L303
[1]https://github.com/tidwall/wal/blob/0fa32977a196086591e6eace7bdf13f67bc305aa/wal.go#L148

thorfour: Great catch! Okay then this seems like a reasonable approach to me!

thorfour (approved):

brancz:
Nice!!

brancz:
Thank you so much for the awesome and thoughtful contribution!

PR #164, opened 2022-08-18
pqarrow: use ColumnIndex to answer distinct queries at the scan level

### LINKED ISSUE #158: Attempt to use ColumnIndex to answer distinct queries
While we already attempt to optimize distinct queries by checking if they are dictionary encoded and only contain a single value, that still causes us to read the page, using the ColumnIndex however, we could avoid reading the page altogether.

asubiotto:
This commit uses a column chunk's column index to optimize distinct queries
where there is a single value in the chunk (across all pages) of a
dictionary-encoded column. In this case, a dictionary can be constructed with
the value stored in the index, avoiding an expensive page decompression step.

```
QueryTypes-8     169ms ±11%     133ms ±18%  -21.14%  (p=0.049 n=10+3)

QueryTypes-8     443MB ± 0%     436MB ± 0%   -1.54%  (p=0.007 n=10+3)

QueryTypes-8     1.13M ± 0%     1.05M ± 0%   -6.78%  (p=0.007 n=10+3)
```

Closes #158 

cc @metalmatze @thorfour

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -346,6 +346,7 @@ func parquetColumnToArrowArray(
 	defer b.Release()
 
 	err = writeColumnToArray(
+		n.Type(),
 		c,
 		optional,
 		repeated,
@@ -375,11 +376,14 @@ func parquetColumnToArrowArray(
 // column. Possilibities why it might not read the whole column:
 //
 // * If it has been requested to only read the dictionary it will only do that
-// (provided it's not a repeated type).
+// (provided it's not a repeated type). Additionally, decompression of all pages
+// are avoided if the column index indicates that there is only one value in the
+// column.
 //
 // If the type is a repeated type it will also write the starting offsets of
 // lists to the list builder.
 func writeColumnToArray(
+	t parquet.Type,
 	columnChunk parquet.ColumnChunk,
 	optional bool,
 	repeated bool,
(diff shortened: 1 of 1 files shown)


asubiotto:
Looks like I missed a test. Will fix

pqarrow/arrow.go
```diff
@@ -387,6 +389,56 @@ func writeColumnToArray(
...
+
+		globalMinValue := columnIndex.MinValue(0)
+		readPages := false
+		for pageIdx := 0; pageIdx < columnIndex.NumPages(); pageIdx++ {
+			if columnType.Length() == 0 {
+				// Variable-length datatype. The index can only be trusted if
+				// the size of the values is less than the column index size,
+				// since we cannot otherwise know if the index values are
+				// truncated.
+				if len(columnIndex.MinValue(pageIdx).Bytes()) >= dynparquet.ColumnIndexSize ||
```
brancz, line 410: This is pretty confusing, but `.Bytes()` only works on values that are "binary" in parquet. Otherwise, this will panic. We should only perform this if that's the primitive type of the column. Otherwise we should just check once that the size of the type (eg. a uint64) fits into `ColumnIndexSize`.

brancz (approved):
The overall approach looks good!

asubiotto:
I took a look at what's failing and it's a type assertion in parquet when I attempt to create a dictionary from a string in the new code path. The underlying `columnType` that I call `NewDictionary` on is a `string` type wrapped by `indexedType`, but `indexedType` overloads its wrapped type's `NewValues` method with a call to `Int32ValuesFromBytes`, which means that the values passed in to `NewDictionary` are int32s. I'm guessing that these are the indexes of the indexedType so not a bug (which indicates to me that I need to create a dictionary in another way), but this behavior seems weird.

metalmatze (approved):
Verrry nice! :tada:

pqarrow/arrow.go
asubiotto: Gotcha, but these `Bytes` calls are within a `Length() == 0` check. My understanding is that all types that return `Length() == 0` are variable-width and so implicitly will return something meaningful from the `Bytes` call. In the `uint64` example, `Length` is 64 so we wouldn't get to this `Bytes` call. Is this a wrong assumption?

thorfour (approved):
Great work!

asubiotto:
Thanks all! I would say to skip the failing test, merge, and unskip once the upstream bug is fixed, but it looks like there are a couple more tests failing due to the same reason. I'll wait to merge this until https://github.com/segmentio/parquet-go/issues/312 is fixed.

asubiotto:
Found a way to bypass the parquet panic (using a passed in explicit type instead of determining it from the column chunk). Will merge on Monday.

metalmatze (approved):
LGTM :tada:

asubiotto:
TFTRs

PR #165, opened 2022-08-20
WAL replay with bucket storage causes panic while registering table metrics

arajkumar:
For a given DB and Table, a read only Table will be created on DB initialisation. During WAL replay for NewTableBlock entry, another Table instance is being created with the same name which causes panic while registering table metrics. This commit fixes the problem by converting existing read only table into normal table.

Signed-off-by: Arunprasad Rajkumar <ar.arunprasad@gmail.com>


thorfour (approved):
Great catch! Thank you

PR #166, opened 2022-08-22
*: add end-to-end benchmarks that represent polar signals query path

### LINKED ISSUE #151: Query path benchmarks
Right now we have a small set of benchmarks that really only benchmark concurrent writers.
It would be good to get some benchmarks around the query path, as well as large database benchmarks etc.

asubiotto:
This commit adds three benchmarks that represent the queries run by Polar
Signals. The benchmarks are skipped by default and require a pre-generated
data directory for now.

Closes #151

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -1,0 +1,209 @@
+package frostdb
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/apache/arrow/go/v8/arrow"
+	"github.com/apache/arrow/go/v8/arrow/array"
+	"github.com/apache/arrow/go/v8/arrow/memory"
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+
+	"github.com/polarsignals/frostdb/query"
+	"github.com/polarsignals/frostdb/query/logicalplan"
+)
+
+const (
+	dbName      = "parca"
+	tableName   = "stacktraces"
+	storagePath = "testdata/data"
+	skipReason  = "requires data directory"
+)
+
+// This file runs end-to-end benchmarks on a single FrostDB instance using
+// Polar Signals-specific data and schema. These benchmarks are skipped by
+// default since they require a manually-generated data directory to run. To run
+// these benchmarks, unskip the benchmarks and create a data directory in
+// testdata/data.
+
+func newDBForBenchmarks(ctx context.Context, b testing.TB) (*DB, error) {
+	b.Helper()
+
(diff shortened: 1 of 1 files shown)


bench_test.go
```diff
@@ -0,0 +1,109 @@
...
+				return nil
+			}); err != nil {
+			b.Fatalf("query returned error: %v", err)
+		}
+	}
+}
+
+// BenchmarkMerge executes a merge of profiles over a 15-minute time window.
+func BenchmarkMerge(b *testing.B) {
+	b.Skip()
```
thorfour, line 97: `nit` prefer to put the comment below into the skip reason unless you plan to wait to merge until the below is satisfied

bench_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	b.Logf("db initialized and WAL replayed, starting benchmark %s", b.Name())
+	return colDB, nil
+}
+
+// Remove unused warning.
+var _ = newDBForBenchmarks
+
+func BenchmarkTypes(b *testing.B) {
+	b.Skip()
```
thorfour, line 67: `nit` add the skip comment of "require data directory" or something along those lines

thorfour (approved):
Nice!

asubiotto:
The first commit I just added is #167 in order to use the max function.

I've written the remaining two benchmarks. Here are the results:
```
name          time/op
QueryTypes-8   136ms ± 0%
QueryMerge-8  24.7ms ± 0%
QueryRange-8  12.1ms ± 0%
```

These are run on a WAL generated overnight. Do these numbers make intuitive sense to you? I want to make sure I've written the queries properly. I basically chose some random filter values for `Merge` and `Range` based on the data we have.

metalmatze:
Makes a lot of sense.
Usually I like to also add `b.ReportAllocs()` to get some insight into memory allocations too.

asubiotto:
Definitely! I won't check that in, since I think the `-benchmem` flag has the same effect

asubiotto:
Oof, half a gig for `Types`:
```
name          time/op
QueryTypes-8   159ms ± 0%
QueryMerge-8  18.9ms ± 0%
QueryRange-8  14.1ms ± 0%

name          alloc/op
QueryTypes-8   443MB ± 0%
QueryMerge-8  39.9MB ± 0%
QueryRange-8  18.6MB ± 0%

name          allocs/op
QueryTypes-8   1.13M ± 0%
QueryMerge-8    104k ± 0%
QueryRange-8   6.61k ± 0%
```

brancz (approved):
Very nice!

PR #169, opened 2022-08-23
Timestamp hints

thorfour:
This adds a new concept called `Hints` to the query engine. Hints are an optional setting you can provide to the query engine for it to potentially make optimizations to it's queries. 

This adds a single hint called `ColAsTimestamp` that informs the query engine that a specific column is being used as a timestamp. This hint is then used as an optimization when performing historic queries of blocks. If a query is provided that filters on that timestamp column, it shall use that expr against the blocks ulid timestamp to determine if the block could contain any timestamps that satisfy that portion of the query, if not it'll skip that block entirely.

This is useful for queries that are being performed over say the last ~15m but have historic blocks much older than that. Instead of reading those blocks the query engine can skip over all of them and only access blocks that were written in the last 15m. This is most notable when build with Parca that has had persistence enabled.

### DIFF
--- a/query/engine.go
+++ b/query/engine.go
@@ -38,7 +38,7 @@ func ColAsTimestamp(columnName string) func(*LocalEngine) {
 func NewEngine(
 	pool memory.Allocator,
 	tableProvider logicalplan.TableProvider,
-	hints ...Hint,
+	hints ...Hint, // TODO THOR rename this to optimizer instead
 ) *LocalEngine {
 	return &LocalEngine{
 		pool:          pool,
--- a/query/engine.go
+++ b/query/engine.go
@@ -22,23 +22,23 @@ type LocalEngine struct {
 	pool          memory.Allocator
 	tableProvider logicalplan.TableProvider
 
-	timestampColHint string
+	optimizations []logicalplan.Optimizer
 }
 
-// Hint is a suggestion to be made to the query engine about how it might more effectively query
+// Hint is a suggestion to be made to the query engine about how it might optimize the query
 type Hint func(*LocalEngine)
 
 // ColAsTimestamp is a query engine hint that informs the engine which column to use as a timestamp during historical queries
 func ColAsTimestamp(columnName string) func(*LocalEngine) {
 	return func(l *LocalEngine) {
-		l.timestampColHint = columnName
+		l.optimizations = append(l.optimizations, logicalplan.TimestampColumnOptimization(columnName))
 	}
 }
 
 func NewEngine(
 	pool memory.Allocator,
 	tableProvider logicalplan.TableProvider,
-	hints ...Hint, // TODO THOR rename this to optimizer instead
+	hints ...Hint,
 ) *LocalEngine {
 	e := &LocalEngine{
 		pool:          pool,
@@ -60,14 +60,14 @@ type LocalQueryBuilder struct {
 func (e *LocalEngine) ScanTable(name string) Builder {
 	return LocalQueryBuilder{
 		pool:        e.pool,
-		planBuilder: (&logicalplan.Builder{}).Scan(e.tableProvider, name, logicalplan.TableScanColAsTimestamp(e.timestampColHint)),
+		planBuilder: (&logicalplan.Builder{}).Scan(e.tableProvider, name),
 	}
 }
 
 func (e *LocalEngine) ScanSchema(name string) Builder {
 	return LocalQueryBuilder{
 		pool:        e.pool,
-		planBuilder: (&logicalplan.Builder{}).ScanSchema(e.tableProvider, name, logicalplan.SchemaScanColAsTimestamp(e.timestampColHint)),
+		planBuilder: (&logicalplan.Builder{}).ScanSchema(e.tableProvider, name),
 	}
 }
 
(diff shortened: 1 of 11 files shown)


brancz:
I expect this won't be the last thing that we'll want to disregard full blocks on, it's the same thing as what we do on a rowgroup/columnchunk level, but one step above. I think we should rather use metadata of a block (if we can't read it from a block then we should write a separate metadata file next to it) that describes the content of the block. We can then put bloomfilters, min/max, etc. of each column into that and ignore full blocks on arbitrary column contents. I think what's proposed here is too tightly coupled to implementation details that we could introduce new optimizations.

thorfour:
> 

The intent was to avoid having to perform any reads of the metadata. I feel like the block filter function is generic enough that all of what you've described above should still be possible using that function.

query/logicalplan/logicalplan.go
```diff
@@ -27,6 +27,16 @@ type LogicalPlan struct {
 	Aggregation *Aggregation
 }
 
+// IterOptions are a set of options for the TableReader Iterators
+// TODO: should we instead use the option pattern? Is that possible with the way the Iterator functions work?
```
metalmatze, line 31: Not sure if it's possible. Would need to try locally. It would be very neat for sure.

query/logicalplan/logicalplan.go
```diff
@@ -27,6 +27,16 @@ type LogicalPlan struct {
...
+type IterOptions struct {
+	PhysicalProjection []Expr
+	Projection         []Expr
+	Filter             Expr
+	DistinctColumns    []Expr
+	TimestampCol       string
```
metalmatze, line 37: The `DistinctColumns` is fully written. Why not this column too?
```suggestion
	TimestampColumn       string
```

store.go
```diff
@@ -118,3 +177,130 @@ func (b *BucketReaderAt) ReadAt(p []byte, off int64) (n int, err error) {
...
+
+			right, righterr := compareTimestamp(col, start, end, e.Right)
+			if righterr != nil && righterr != ErrNonTimestampCol {
+				return false, righterr
+			}
+
+			return left || right, nil
+		default:
+			// unsupported op; return true. Better to have false positives
+			return true, nil
```
metalmatze, line 301: If possible we should at least log this. This actually bit us before, where we silently ignore things and it was super hard to find in the end...

query/logicalplan/logicalplan.go
thorfour: Yea I don't think it is without a serious rewrite

brancz:
> The intent was to avoid having to perform any reads of the metadata. I feel like the block filter function is generic enough that all of what you've described above should still be possible using that function.

We already read which blocks exist, so we're already reading metadata, but implicitly and only the timestamp that's in the ULID. I just don't think that we should make this a special case, I don't think there is anything special about this that can't be optimized using knowledge about blocks and a query being executed.

I do think that it could be useful to have a faster "cache" that stores block metadata. That's a fancy way of saying we read the metadata once at startup or when uploaded.

query/engine.go
```diff
@@ -99,6 +125,10 @@ func (b LocalQueryBuilder) Execute(ctx context.Context, callback func(r arrow.Re
 		logicalPlan = optimizer.Optimize(logicalPlan)
 	}
 
+	for _, optimizer := range b.optimizations {
```
asubiotto, line 128: nit: add a comment about the difference between these optimizations and the ones above

query/logicalplan/logicalplan.go
```diff
@@ -108,30 +118,21 @@ type TableReader interface {
 		tx uint64,
 		pool memory.Allocator,
 		schema *arrow.Schema,
-		physicalProjection []Expr,
-		projection []Expr,
-		filter Expr,
-		distinctColumns []Expr,
+		options *IterOptions,
```
asubiotto, line 121: I think this argument can be a value (it's not huge) to avoid allocations.

store.go
```diff
@@ -118,3 +177,130 @@ func (b *BucketReaderAt) ReadAt(p []byte, off int64) (n int, err error) {
...
+// ErrNonTimestampCol is returned by compareTimestamp when the expression is not for the given timestamp column.
+var ErrNonTimestampCol = errors.New("non timestamp column")
+
+// compareTimestamp returns true if the block contains timestamps covered in expr.
+func compareTimestamp(col string, start, end uint64, expr logicalplan.Expr) (bool, error) {
+	if expr == nil {
+		return false, nil
+	}
+
+	parseExpr := func(e *logicalplan.BinaryExpr) (uint64, error) {
```
asubiotto, line 190: I feel like having this expression evaluation code feels wrong in `store.go`. Is there some way to extract this and the switch below to into some sort of `expr.Eval` function to be reused?

store.go
brancz: I actually don't think we should do anything here. This is a safe fallback, where the worst thing that can happen is that we look at the block. The only thing that the place where this bit us had in common was a type switch, but that's where the similarity ends in my opinion. I don't see a need for anything other than tests for the cases that we want to ensure work, which this has.

thorfour:
Closing this in favor of meta files

PR #170, opened 2022-08-24
Filter blocks based on min/max values against query expr

thorfour:
In place of my previous PR with query hints.

This no longer has query hints, it instead using the query expr against the min/max values in a block from the parquet file metadata to determine if the block may contain data that satisfies the query.

This appeared to have the same effect as my previous PR but performs this in a more generic way as requested.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -20,6 +20,9 @@ func (f PreExprVisitorFunc) PostVisit(expr logicalplan.Expr) bool {
 	return false
 }
 
+// Particulate is an abstraction of something that can be filtered.
+// A parquet.RowGroup is a particulate that is able to be filtered, and wrapping a parquet.File with
+// ParquetFileParticulate allows a file to be filtered.
 type Particulate interface {
 	Schema() *parquet.Schema
 	ColumnChunks() []parquet.ColumnChunk
--- a/filter.go
+++ b/filter.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/segmentio/parquet-go"
 
-	"github.com/polarsignals/frostdb/dynparquet"
 	"github.com/polarsignals/frostdb/pqarrow"
 	"github.com/polarsignals/frostdb/query/logicalplan"
 )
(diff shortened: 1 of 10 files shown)


store.go
```diff
@@ -40,7 +45,91 @@ func (t *TableBlock) Persist() error {
...
+}
+
+func (b *BlockFilter) MinMaxFilter(filter logicalplan.Expr) *BlockFilter {
+	return &BlockFilter{
+		BlockFilterFunc: func(block ulid.ULID, file *parquet.File) bool {
+			mins := map[string]*parquet.Value{}
+			maxes := map[string]*parquet.Value{}
+			for i, cidx := range file.ColumnIndexes() {
+				var c parquet.ColumnIndex
+				switch *file.Metadata().Schema[i+1].Type { // TODO THOR this is probably not the correct way to do this
```
thorfour, line 87: Feels like there has to be a better way than what I'm doing here...

metalmatze: Do you mean the way you're retrieving the metadata or this switch statement for the types themselves?

store.go
```diff
@@ -118,3 +207,168 @@ func (b *BucketReaderAt) ReadAt(p []byte, off int64) (n int, err error) {
...
+			// unsupported op; return true. Better to have false positives
+			return true, nil
+		}
+	default:
+		return false, fmt.Errorf("unsupported expr %T", e)
+	}
+}
+
+// compares two parquet values. 0 if they are equal, -1 if v1 < v2, 1 if v1 > v2.
+func compare(v1, v2 *parquet.Value) int {
```
asubiotto, line 331: `parquet.Type` has a `Compare` method, I think you should be able to use that instead of this function

query/logicalplan/logicalplan.go
```diff
@@ -108,30 +117,21 @@ type TableReader interface {
 		tx uint64,
 		pool memory.Allocator,
 		schema *arrow.Schema,
-		physicalProjection []Expr,
-		projection []Expr,
-		filter Expr,
-		distinctColumns []Expr,
+		options *IterOptions,
```
asubiotto, line 120: Repeating the same comment as on the other PR: I think we can change this pointer argument to a value to avoid allocations

store.go
```diff
@@ -118,3 +207,168 @@ func (b *BucketReaderAt) ReadAt(p []byte, off int64) (n int, err error) {
 
 	return rc.Read(p)
 }
+
+// ErrColNotExist is returned when the column described in the filter does not exist in the columns.
+var ErrColNotExist = errors.New("column does not exist")
+
+func compareMinsMaxes(mins, maxes map[string]*parquet.Value, expr logicalplan.Expr) (bool, error) {
```
asubiotto, line 214: Same comment as other PR: this seems to be an expression evaluation with a parquet value at first glance (might be missing some complications). Ideally, we would reuse expression code here/write this in a more general way. I might be missing something though

asubiotto (commented):
Nice!

store.go
thorfour: ah ha! I knew there was something like that, thanks for pointing me to it!

metalmatze (approved):

filter.go
```diff
@@ -21,18 +20,33 @@ func (f PreExprVisitorFunc) PostVisit(expr logicalplan.Expr) bool {
 	return false
 }
 
+type Particulate interface {
```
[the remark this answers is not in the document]
thorfour: Added!

PR #172, opened 2022-08-25
query/physicalplan: PhysicalPlans are passed rather than just function callbacks

metalmatze:
Adding the `Finish(context.Context) error` callbacks next.


metalmatze:
I should mention: Most of the code was written by @albertlockett in #90. 
This PR takes parts of that PR and thus most of the code was written by @albertlockett originally and he's a co-author.

asubiotto (approved):

PR #175, opened 2022-08-26
pqarrow/frostdb: accumulate results in Iterator

asubiotto:
Previously, in some cases, the Table.Iterator method would push Records with a
single row one at a time to the next operator. This would happen especially in
cases where a distinct optimization is executed and a single value is written
per parquet page. This commit changes the scan process to write scan results to
a provided RecordBuilder in order to buffer results up to a specified
builderBufferSize (default=1024) and take advantage of processing these results
in tight loops downstream.

The results are especially noticeable in QueryTypes, which takes advantage of
the aforementioned distinct optimizations:

```
QueryTypes-8     169ms ±11%      68ms ±13%  -59.43%  (p=0.000 n=10+9)
QueryMerge-8    27.0ms ±34%    23.4ms ±12%     ~     (p=0.095 n=10+9)
QueryRange-8    13.8ms ±15%    13.3ms ± 9%     ~     (p=0.356 n=9+10)

QueryTypes-8     443MB ± 0%      88MB ± 0%  -80.21%  (p=0.000 n=10+9)
QueryMerge-8    39.9MB ± 0%    39.9MB ± 0%     ~     (p=0.497 n=10+9)
QueryRange-8    18.6MB ± 0%    18.6MB ± 0%   -0.05%  (p=0.000 n=10+10)

QueryTypes-8     1.13M ± 0%     0.52M ± 0%  -53.55%  (p=0.000 n=10+9)
QueryMerge-8      104k ± 0%      104k ± 0%   -0.10%  (p=0.000 n=10+9)
QueryRange-8     6.61k ± 0%     6.49k ± 0%   -1.69%  (p=0.000 n=8+10)
```

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -271,6 +271,10 @@ func Test_DB_WithStorage(t *testing.T) {
 		dynparquet.NewSampleSchema(),
 	)
 
+	// TODO(asubiotto): These buckets are not deleted from disk, resulting in
+	// test failures if run multiple times. There seems to be an attempt below
+	// to delete them, but it doesn't have an effect. What's the magic
+	// incantation?
 	bucket, err := filesystem.NewBucket(".")
 	require.NoError(t, err)
 
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -98,6 +98,7 @@ func includedProjection(projections []logicalplan.Expr, name string) bool {
 }
 
 // ParquetRowGroupToArrowRecord converts a parquet row group to an arrow record.
+// The result is appended to builder.
 func ParquetRowGroupToArrowRecord(
 	ctx context.Context,
 	pool memory.Allocator,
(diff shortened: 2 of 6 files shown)


asubiotto:
I think I would still like to add a test or two (even if the test suite caught a bunch of things). I'll leave exploring improving the intermediate result cache on top of the record builder (to avoid having to recreate a builder if an optimization cannot be applied) to a future PR.

asubiotto:
Cherry-picking #164 on to this PR gives us an addition 10% improvement:
```
QueryTypes-8     169ms ±11%      50ms ±13%  -70.19%  (p=0.030 n=10+2)

QueryTypes-8     443MB ± 0%      81MB ± 0%  -81.80%  (p=0.030 n=10+2)

QueryTypes-8     1.13M ± 0%     0.45M ± 0%  -60.34%  (p=0.000 n=10+2)
```

thorfour (approved):
Love this change

asubiotto:
Added a test for list types and refactored the list building code while I was at it. I'll let this PR sit for a bit in case there are further comments, but planning on merging this soon.

PR #178, opened 2022-08-30
*: update parquet-go to include goroutine leak fix

asubiotto:
This should alleviate our CI flakes where too many goroutines are alive at one
time.

There is also a fix pulled in for buffering bloom filter reads.

### DIFF
--- a/dynparquet/nil_chunk.go
+++ b/dynparquet/nil_chunk.go
@@ -198,21 +198,21 @@ func (p *nilValueReader) ReadValues(values []parquet.Value) (int, error) {
 
 // DefinitionLevels returns the definition levels of the page. Since the page
 // contains only null values, all of them are 0. Implements the
-// parquet.BufferedPage interface.
+// parquet.Page interface.
 func (p *nilPage) DefinitionLevels() []byte {
 	return nil
 }
 
 // RepetitionLevels returns the definition levels of the page. Since the page
 // contains only null values, all of them are 0. Implements the
-// parquet.BufferedPage interface.
+// parquet.Page interface.
 func (p *nilPage) RepetitionLevels() []byte {
 	return nil
 }
 
 // Data is unimplemented, since the page is virtual and does not need to be
 // written in its current usage in this package. If that changes this method
-// needs to be implemented. Implements the parquet.BufferedPage interface.
+// needs to be implemented. Implements the parquet.Page interface.
 func (p *nilPage) Data() encoding.Values {
 	panic("not implemented")
 }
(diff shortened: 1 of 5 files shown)


asubiotto:
Interesting failure mode on first CI run. Looks like an unrelated flake:
```
=== RUN   Test_Table_Concurrency/8192
race: limit on 8128 simultaneously alive goroutines is exceeded, dying
FAIL	github.com/polarsignals/frostdb	12.551s
```
The test only seems to spawn 8 goroutines. Trying to reproduce the failure locally.

asubiotto:
Looks like the cause if a bunch of leaked goroutines in `asyncPage.init`:
```
         Goroutine 105824 in state select, with github.com/segmentio/parquet-go.readPages on top of the stack:
        goroutine 105824 [select]:
        github.com/segmentio/parquet-go.readPages({0x105d992a0, 0xc06f880198}, 0xc06f23e180, 0xc06daf9ab0, 0xc06f23e0c0)
                /Users/asubiotto/go/pkg/mod/github.com/segmentio/parquet-go@v0.0.0-20220830163417-b03c0471ebb0/page.go:222 +0x168
        created by github.com/segmentio/parquet-go.(*asyncPages).init
                /Users/asubiotto/go/pkg/mod/github.com/segmentio/parquet-go@v0.0.0-20220830163417-b03c0471ebb0/page.go:161 +0x270
        ]
```

Looks like we're spawning 100k goroutines in a single test run. I think the likeliest thing is that this is a latent bug in our page lifecycle management that has only surfaced due to https://github.com/segmentio/parquet-go/pull/297.

asubiotto:
The last commit fixes the `asyncPage.init` leak but CI is still failing. We are still leaking some goroutines (see https://github.com/polarsignals/frostdb/issues/180) but this is an issue on `main` as well. I'm guessing that we're still spawning a ton of goroutines simultaneously when writing rows. I'm wondering if this is expected.

asubiotto:
Apologies for the excess notifications. I'm going to try peppering the code with some missing `Close` calls to see if we can at least alleviate the number of simultaneously alive goroutines.

asubiotto:
I found a potential goroutine leak in parquet (fixed in https://github.com/segmentio/parquet-go/pull/337) which I think fixes our problem (at least makes the leak detector not scream at me when running without finalizers). I'll merge the upstream fix, update the version on this PR and then retry.

asubiotto:
RFAL, I pulled in the upstream goroutine leak fix as well as the bloom filter read buffering @thorfour

brancz (approved):

PR #179, opened 2022-08-30
Reduce mem overhead 2 electric boogaloo

thorfour:
This is an attempt to address [#301](https://github.com/parca-dev/parca/issues/301).

This change does 2 main things to reduce the memory overhead. 

1) Instead of opening multiple writers for each granule during an insert, we instead only open a single writer per insert and use that writer for all granule writes, resetting it after each granule has been written to. Of note, I initially tried to have the `splitRowsByGranule` function return `map[*Granule][]parquet.Row` and have the writers write those rows directly (can see in commit `41b8914`). However this resulted in bad profiles being produced, so I'm not entirely sure that didn't work, but if anyone knows that would be an easy win to make this code cleaner. 

2) Instead of holding the open `parquet.File` inside the `SerializedBuffer` object we instead just hold the `[]byte` of the file object. 

With these changes we see about an order of magnitude reduction in memory overhead. 
image

This does slow down queries due to it needing to read the bytes as the file instead of just accessing them directly, however the slow down doesn't make Parca unusable, and is likely worth the trade off here, and we can make query improvements after this to address the lag. Also noticeably there's a large memory spike during persistence, this will be something to investigate as well.

### DIFF
--- a/dynparquet/reader.go
+++ b/dynparquet/reader.go
@@ -15,24 +15,12 @@ var ErrNoDynamicColumns = errors.New("no dynamic columns metadata found, it must
 
 type SerializedBuffer struct {
 	file []byte
-
-	/*
-		dynCols map[string][]string
-		fields  []parquet.Field
-	*/
 }
 
 func ReaderFromBytes(buf []byte) (*SerializedBuffer, error) {
 	return &SerializedBuffer{
 		file: buf,
 	}, nil
-}
-
-func NewSerializedBuffer(file *parquet.File) (*SerializedBuffer, error) {
-	if true {
-		panic("shit")
-	}
-	return &SerializedBuffer{}, nil
 }
 
 func (b *SerializedBuffer) Reader() *parquet.Reader {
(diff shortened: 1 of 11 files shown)


thorfour:
Latest commit reduces the memory spike by about 1/2

image

thorfour:
image

dynparquet/reader.go
```diff
@@ -52,24 +28,30 @@ func (b *SerializedBuffer) Reader() *parquet.Reader {
 }
 
 func (b *SerializedBuffer) ParquetFile() *parquet.File {
-	return b.f
+	f, err := parquet.OpenFile(bytes.NewReader(b.file), int64(len(b.file)))
+	if err != nil {
+		panic("at the disco")
```
metalmatze, line 33: haha.

Is there no way to return the error though?

thorfour: hah! Thanks, yea I'll replace those with error propagation if this is something we want to merge. Just didn't want to go through all that if this wasn't an approach we liked.

metalmatze: Always rather refactor and return errors than panic. If it can be somewhat ignored then maybe logging it is fine. 

Either way:
> [Don't panic.](https://github.com/golang/go/wiki/CodeReviewComments#dont-panic)

thorfour:
I wonder if we should consider not useing parquet format but the arrow record format instead for in memory

dynparquet/reader.go
thorfour: Panics have been removed

thorfour:
Closing in favor of #189

PR #183, opened 2022-09-01
Obsv as options

thorfour:
Makes logging, tracing, and metrics options instead of required params.

### DIFF
--- a/query/engine.go
+++ b/query/engine.go
@@ -25,16 +25,30 @@ type LocalEngine struct {
 	tableProvider logicalplan.TableProvider
 }
 
+type Option func(*LocalEngine)
+
+func WithTracer(tracer trace.Tracer) Option {
+	return func(e *LocalEngine) {
+		e.tracer = tracer
+	}
+}
+
 func NewEngine(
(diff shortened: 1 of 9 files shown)


query/engine.go
```diff
@@ -25,16 +25,30 @@ type LocalEngine struct {
...
+type Option func(*LocalEngine)
+
+func WithTracer(tracer trace.Tracer) Option {
+	return func(e *LocalEngine) {
+		e.tracer = tracer
+	}
+}
+
 func NewEngine(
 	pool memory.Allocator,
```
metalmatze, line 37: I always wonder if we should do the same with these `memory.Allocator`.
Totally out of scope for this PR, maybe creating an issue for outside contributions. That would be a nice one, I imagine.

asubiotto (approved):
Thank you!

metalmatze (approved):
Awsome. Thanks for cleaning this up!

query/engine.go
thorfour: Totally agree, I strongly considered removing it in this PR as well but decided to leave it for now

PR #185, opened 2022-09-01
query: Introduce array of callbacks to be called concurrently

metalmatze:
For now, we only ever call the first callback, since not all operators of the PhysicalPlan are concurrency-safe yet.

Overall the plan is already concurrency ready as per our "draw" output that only works based on the amount of callbacks. 
```
TableScan(24x) - Projection(24x) (labels.label1) - Distinction(24x) (labels.label1)
```

An easier-to-understand version of this idea is available in #184

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -416,8 +416,6 @@ github.com/sean-/seed v0.0.0-20170313163322-e2103e2c3529/go.mod h1:DxrIzT+xaE7yg
 github.com/segmentio/asm v1.1.3/go.mod h1:Ld3L4ZXGNcSLRg4JBsZ3//1+f/TjYl0Mzen/DQy1EJg=
 github.com/segmentio/encoding v0.3.5 h1:UZEiaZ55nlXGDL92scoVuw00RmiRCazIEmvPSbSvt8Y=
 github.com/segmentio/encoding v0.3.5/go.mod h1:n0JeuIqEQrQoPDGsjo8UNd1iA0U8d8+oHAA4E3G3OxM=
-github.com/segmentio/parquet-go v0.0.0-20220725200142-047e5979dfcf h1:5yLngjz3nAomA1BbuHoRdPaCSsmy2RP2n+KF2Ew+VUs=
-github.com/segmentio/parquet-go v0.0.0-20220725200142-047e5979dfcf/go.mod h1:BuMbRhCCg3gFchup9zucJaUjQ4m6RxX+iVci37CoMPQ=
 github.com/segmentio/parquet-go v0.0.0-20220802221544-d84ed320251d h1:IxYmTiYCMaR3B0fjD9igXDr1AE5M8KRFbhEhEJ1WYx4=
 github.com/segmentio/parquet-go v0.0.0-20220802221544-d84ed320251d/go.mod h1:BuMbRhCCg3gFchup9zucJaUjQ4m6RxX+iVci37CoMPQ=
 github.com/shurcooL/sanitized_anchor_name v1.0.0/go.mod h1:1NzhyTcUVG4SuEtjjoZeVRXNmyL/1OwPU0+IJeTBvfc=
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -115,12 +115,18 @@ type HashAggregate struct {
 	columnToAggregate     logicalplan.Expr
 	aggregationFunction   AggregationFunction
 	hashSeed              maphash.Seed
-	next                  PhysicalPlan
+
+	next      PhysicalPlan
+	callbacks []func(ctx context.Context, r arrow.Record) error
 
 	// Buffers that are reused across callback calls.
 	groupByFields      []arrow.Field
 	groupByFieldHashes []uint64
 	groupByArrays      []arrow.Array
+}
+
+func (a *HashAggregate) Callbacks() []func(ctx context.Context, r arrow.Record) error {
+	return a.callbacks
 }
 
 func NewHashAggregate(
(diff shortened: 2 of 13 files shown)


query/physicalplan/physicalplan.go
```diff
@@ -14,10 +15,12 @@ import (
 	"github.com/polarsignals/frostdb/query/logicalplan"
 )
 
+var concurrency = runtime.NumCPU()
```
thorfour, line 18: I think this could be the default, but I think concurrency should be a query engine option

query/physicalplan/physicalplan.go
```diff
@@ -14,10 +15,12 @@ import (
...
+
 type PhysicalPlan interface {
-	Callback(ctx context.Context, r arrow.Record) error
-	Finish(ctx context.Context) error
+	Callbacks() []func(ctx context.Context, r arrow.Record) error
```
thorfour, line 21: Maybe we should consider making callback a type so that we don't have to keep typing it out

table.go
```diff
@@ -509,14 +512,17 @@ func (t *Table) SchemaIterator(
...
 	ctx, span := t.tracer.Start(ctx, "Table/SchemaIterator")
 	span.SetAttributes(attribute.Int("physicalProjections", len(iterOpts.PhysicalProjection)))
 	span.SetAttributes(attribute.Int("projections", len(iterOpts.Projection)))
 	span.SetAttributes(attribute.Int("distinct", len(iterOpts.DistinctColumns)))
 	defer span.End()
 
+	// TODO(metalmatze): Wrap the following in an errgroup to introduce concurrency
+	iterator := iterators[0]
+
 	rowGroups, err := t.collectRowGroups(ctx, tx, iterOpts.Filter)
```
thorfour, line 526: If we're adding concurrency it probably makes sense to have the row group collection be writers into a channel instead of in-memory buffering all of them.

thorfour (commented):
I like the direction of this so far

query/physicalplan/physicalplan.go
asubiotto: +1, or an interface

query/physicalplan/physicalplan.go
metalmatze: In another branch I had exactly that TODO comment. Forgot to adding it here. :grimacing:

table.go
metalmatze: Yes, that would be perfect. We could then probably use channels instead of the errgroup altogether, right?

query/physicalplan/aggregate.go
```diff
@@ -236,103 +238,110 @@ func hashInt64Array(arr *array.Int64) []uint64 {
...
-		}
+func (a *HashAggregate) SetNext(next PhysicalPlan) {
+	a.next = next
 
-		if a.columnToAggregate.MatchColumn(field.Name) {
-			columnToAggregate = r.Column(i)
-			aggregateFieldFound = true
-		}
-	}
+	a.callbacks = make([]logicalplan.Callback, 0, len(next.Callbacks()))
```
asubiotto, line 244: It's confusing to me that the operators (here HashAggregate, but Distinction as well below) need to know about the multiple callbacks. I think that these operators should stay as they are (only aware of a single execution flow), but many copies of this operator should be instantiated at the physical planning stage for a single logical plan stage.

metalmatze: Yes, you're right. 
What we really want should be a ConcurrentAggregator that's composed of one Aggregator, one Synchronizer, and n-Aggregators internally. 
I'll rework it to that, however, I think for the scope of just introducing multiple callbacks and for now only using the first callback, this should be a nice path into the component described above.

query/physicalplan/physicalplan.go
metalmatze: Actually, this should be just `1` in this PR. We don't want concurrency yet, just multiple callbacks support.

query/physicalplan/aggregate.go
asubiotto: I wouldn't make it so specific (i.e. there is no need for a new component ConcurrentAggregator). In the plan you described above, you would just pass the same synchronizer callback to each of the n aggregators (i.e. they will each call the synchronizer) and the synchronizer is the only component aware that it needs to handle concurrent calls to its callback method.

go.sum
```diff
@@ -416,8 +416,6 @@ github.com/sean-/seed v0.0.0-20170313163322-e2103e2c3529/go.mod h1:DxrIzT+xaE7yg
 github.com/segmentio/asm v1.1.3/go.mod h1:Ld3L4ZXGNcSLRg4JBsZ3//1+f/TjYl0Mzen/DQy1EJg=
 github.com/segmentio/encoding v0.3.5 h1:UZEiaZ55nlXGDL92scoVuw00RmiRCazIEmvPSbSvt8Y=
 github.com/segmentio/encoding v0.3.5/go.mod h1:n0JeuIqEQrQoPDGsjo8UNd1iA0U8d8+oHAA4E3G3OxM=
-github.com/segmentio/parquet-go v0.0.0-20220725200142-047e5979dfcf h1:5yLngjz3nAomA1BbuHoRdPaCSsmy2RP2n+KF2Ew+VUs=
```
asubiotto, line 419: OOC, how come this is removed? I assumed go.sum was only ever added to (accompanied with a go.mod change)

query/physicalplan/aggregate.go
```diff
@@ -236,103 +238,110 @@ func hashInt64Array(arr *array.Int64) []uint64 {
...
+	for range next.Callbacks() {
```
asubiotto, line 245: To be clear: is your idea to merge callback awareness in operators as a first step and then move this to the physical plan stage? I'm very uncomfortable with having planning code in execution but ok if it's temporary

go.sum
metalmatze: For some reason, my setup did this. :man_shrugging: I might have run `go mod tidy` but do not remember. Not sure. Don't think it breaks anything.

PR #187, opened 2022-09-06
Create array of PhysicalPlans to prepare for concurrency

metalmatze:
All slices only have 1 entry which means no concurrency.

The Diagrams are currently broken, but I'll update them afterward. 
I think it makes sense to build a tree for those and use some sort of ASCI lib to print the tree.

### DIFF
--- a/table.go
+++ b/table.go
@@ -474,7 +474,7 @@ func (t *Table) Iterator(
 	pool memory.Allocator,
 	schema *arrow.Schema,
 	iterOpts logicalplan.IterOptions,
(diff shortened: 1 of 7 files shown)


brancz:
lgtm

table.go
```diff
@@ -481,6 +481,8 @@ func (t *Table) Iterator(
 		return err
 	}
 
+	callback := callbacks[0]
```
thorfour, line 484: Can we add that this uses index 0 because we don't support concurrency currently?

thorfour (approved):
lgtm

PR #190, opened 2022-09-09
Support average aggregation

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

One thing I am not sure about is how to deal with the float64 type array in the tests.

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -81,24 +81,30 @@ func TestAggregate(t *testing.T) {
 	)
 
 	for _, testCase := range []struct {
-		fn     func(logicalplan.Expr) *logicalplan.AggregationFunction
-		alias  string
-		expVal int64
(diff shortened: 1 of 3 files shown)


metalmatze:
Thanks for starting with this feature. 
Since we've introduced the new possibility to have concurrent physical plans we need to split the calculation of averages into multiple steps. 
We need to calculate the count and sum of the value that should be average transparently and only at the very last aggregation step run `sum / count`. 

Now, we can merge it as is, but that would make the introduction of the concurrent physical plan be blocked on this feature until we refactored it to be concurrent. 
How critical is it for you to get the average into FrostDB right now?

yeya24:
Thanks @metalmatze.
This feature is not urgent. I can calculate the average on the application side as well.

thorfour:
I'm going to close this, since the query engine has change dramatically since opening we'll need a different approach for averages. 

Thanks!

PR #191, opened 2022-09-13
query/physicalplan: Check if returned rows match expected rows

metalmatze:
Previously we only checked that the expected number of rows are returned. However, it seems that the actually returned rows should be checked as well and are currently inconsistent. We're basically not handling NULL correctly if I'm not mistaken.


asubiotto:
Taking a look now

asubiotto:
Added a fix. Weirdly, NULLs are not represented in indexes or dictionary pages in parquet. Have to verify this is expected, but bypassing optimized path in these cases for now.

PR #198, opened 2022-09-15
pqarrow: AppendNull to adjust the offset

metalmatze:


asubiotto (approved):
LGTM, but maybe it deserves a comment along the lines of "We cannot use unsafe appends with the binary builder because offsets won't be appended"

asubiotto:
LGTM

PR #199, opened 2022-09-15
table: Write api to make FrostDB easier to use

thorfour:
This adds a layer on top of the `Insert` API to increase accessibility of the API. The new API function is `Write(context.Context, ...any)` where a user may write a normal Go struct and FrostDB will handle converting it into a parquet buffer.

I had taken a first pass at this some time ago in https://github.com/polarsignals/frostdb/pull/67/files but finally got around to cleaning that up and making it even more ergonomic.

While I'm certain this does not cover every type of struct that someone may want to write, it covers the use cases that exist in FrostDB today, and can be expanded overtime to support more fields as required.

### DIFF
--- a/examples/simple.go
+++ b/examples/simple.go
@@ -53,7 +53,7 @@ func main() {
 	thor := Simple{
 		Names: FirstLast{
 			FirstName: "Thor",
-			Surname:   "",
+			Surname:   "Hansen",
 		},
 		Value: 99,
 	}
--- a/examples/simple.go
+++ b/examples/simple.go
@@ -74,7 +74,7 @@ func main() {
 			Surname    string
 		}{
 			FirstName:  "Matthias",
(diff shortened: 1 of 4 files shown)


examples/simple.go
```diff
@@ -33,45 +32,61 @@ func main() {
...
+		}
+		Value int64
+	}{
+		Names: struct {
+			FirstName  string
+			MiddleName string
+			Surname    string
+		}{
+			FirstName:  "Matthias",
+			MiddleName: "Oliver",
```
metalmatze, line 77: lol, first time seeing this example 😄 
```suggestion
			MiddleName: "Oliver Rainer",
```
😝

metalmatze (approved):
I'm wondering if there's a possibility to come up with a minimal interface to not use `any`. Regardless, it looks good.

PR #201, opened 2022-09-19
physicalplan/pqarrow: handle partially pre-computed results in Project

asubiotto:
Previously, the scan layer would pre-compute binary distinct expressions at a row group granularity. Downstream projections would only evaluate the expression if all results were null (i.e. no optimization was able to be made).

This logic broke down when we started accumulating row group conversion results since binary expression evaluations could be partially optimized (i.e. we could fail to optimize some row groups in a record), therefore rows belonging to a row group where the expression was not pre-evaluated would result in the result for that row automatically evaluating to NULL.

This commit changes the logic to only skip evaluation when the expression result has been fully computed by the scan layer. In case of partial scan layer results, those are merged into the projection expression evaluation result. Note that pre-computed results in a partial scan evaluation have a NULL left hand column (since reading this column would be expensive and defeat the point of pre-evaluating the expression), so the projection expression evaluation is skipped for these values in any case.

This commit also fixes and adds a test for another bug where the ParquetConverter's builder is not correctly reset when there is more than one column with more than one distinct value.

### DIFF
--- a/distinct_test.go
+++ b/distinct_test.go
@@ -35,7 +35,8 @@ func TestDistinct(t *testing.T) {
 	samples := dynparquet.Samples{{
 		Labels: []dynparquet.Label{
 			{Name: "label1", Value: "value1"},
-			{Name: "label2", Value: "value2"},
+			{Name: "label2", Value: "value1"},
+			{Name: "label5", Value: "value1"},
 		},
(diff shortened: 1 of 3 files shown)


metalmatze (approved):
Amazing work! 💯

asubiotto:
TFTR!

PR #202, opened 2022-09-20
query/physicalplan: Use concurrency in the physicalplan 

metalmatze:
Mostly making the operators concurrent by using the hardcoded concurrency (for now) to spawn many callbacks. We added a Synchronizer to synchronize the callbacks to in the end call the final callback synchronously. 

```
name           old time/op  new time/op  delta
QueryMerge-10  22.3ms ± 0%  13.5ms ± 2%  -39.34%  (p=0.010 n=4+6)

name           old time/op  new time/op  delta
QueryRange-10  12.0ms ± 1%   4.6ms ± 1%  -61.81%  (p=0.036 n=3+5)

name           old time/op  new time/op  delta
QueryTypes-10  5.06ms ± 0%  5.18ms ± 6%   ~     (p=0.250 n=3+5)
```


distinct_test.go
```diff
@@ -173,7 +181,6 @@ func TestDistinct(t *testing.T) {
 				Distinct(test.columns...).
 				Execute(context.Background(), func(ctx context.Context, ar arrow.Record) error {
 					defer ar.Release()
-					require.Equal(t, len(test.values), int(ar.NumRows()))
```
asubiotto, line 176: Accidental deletion?

query/physicalplan/aggregate.go
```diff
@@ -130,6 +135,7 @@ func NewHashAggregate(
 	aggregationFunction AggregationFunction,
 	columnToAggregate logicalplan.Expr,
 	groupByColumnMatchers []logicalplan.Expr,
+	finalStage bool,
```
asubiotto, line 138: Instead of adding the final stage boolean to the physical plan execution, I think this logic should be part of the physical plan planning. i.e. when planning a multi-stage count, instead of planning a `count(final=true)` operator, you would just plan a `sum()` operator at the final stage.

query/physicalplan/physicalplan.go
```diff
@@ -233,25 +246,85 @@ func Build(ctx context.Context, pool memory.Allocator, tracer trace.Tracer, s *d
...
+			}
+
+			phyPlans = make([]PhysicalPlan, 0, concurrency)
+			for i := 0; i < concurrency; i++ {
+				phyPlans = append(phyPlans, Distinct(pool, tracer, plan.Distinct.Exprs))
+			}
+
+			// If the prev is 1 then we need a synchronizer to handle the concurrency
+			// writing into a single callback of the previous PhysicalPlan.
+			if len(prev) == 1 {
```
asubiotto, line 271: If we decide to make concurrency configurable at runtime (which we might want to do based on load), I think this will create an unnecessary synchronizer if we have concurrency == 1 for any reason. What do you think of creating a struct with more planning state to make more informed decisions of what to plan? For example, store some sort of notion of whether we *have* to plan a synchronizer. Maybe it's fine to also check `len(phyPlans)`, but it seems to me like this state complexity will start to balloon. This struct I suggest might also be useful for more planning information (e.g. ordering column guarantees)

query/physicalplan/physicalplan.go
```diff
@@ -260,17 +333,21 @@ func Build(ctx context.Context, pool memory.Allocator, tracer trace.Tracer, s *d
...
+
 		prev = phyPlans
 		return true
 	}))
 
 	if len(prev) == 1 {
 		outputPlan.SetNextCallback(prev[0].Callback)
 	} else {
-		// TODO: Update to use sychronizer's single callback if we have concurrency.
+		// This case should be handled in the concurrent operator above.
```
asubiotto, line 350: Maybe returning errors here would be good. Or maybe a panic since this would be programming error?

query/physicalplan/synchronize.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func Synchronize(concurrency int) *Synchronizer {
+	return &Synchronizer{running: atomic.NewUint64(uint64(concurrency))}
+}
+
+func (m *Synchronizer) Callback(ctx context.Context, r arrow.Record) error {
+	// multiple threads can emit the results to the next step, but they will do
+	// it synchronously
+	m.nextMtx.Lock()
```
asubiotto, line 28: Instead of using a mutex, I would use a buffered channel here (maybe with `concurrency` slots) with a synchronizer goroutine reading from the channel and pushing. The nice thing with a buffered channel is that upstream goroutines can push data to be read even if the downstream goroutine is not ready to read this data yet (e.g. so the upstream goroutine can perform more scan work concurrently), while still being able to apply backpressure if the downstream goroutine is super slow for some reason.

query/physicalplan/synchronize.go
```diff
@@ -0,0 +1,55 @@
...
+
+	err := m.next.Callback(ctx, r)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (m *Synchronizer) Finish(ctx context.Context) error {
+	if m.running.Dec() > 0 {
```
asubiotto, line 39: Not sure if it makes sense, but for safety I would also add a `< 0` check and return an error if so to protect against double close. This shouldn't happen, but it happens more often than you would think (e.g. at Cockroach), which needs to be debugged if Finish calls are not idempotent.

table.go
```diff
@@ -604,6 +605,18 @@ func (t *Table) View(ctx context.Context, fn func(ctx context.Context, tx uint64
 	return fn(ctx, tx)
 }
 
+// recordBuilderExceedsNumRows assumes all the RecordBuilder's fields have equal
+// length and returns whether at least one field is longer than or equal to
+// threshold.
+func recordBuilderExceedsNumRows(builder *array.RecordBuilder, threshold int) bool {
```
asubiotto, line 611: I think this can be deleted since we now use `ParquetConverter.NumRows()`

asubiotto (changes requested):
Fantastic work! What do the benchmarks say?

distinct_test.go
metalmatze: No, this is intentional, since we sometimes get 1 row and then 2 rows or the other way around, using the callback concurrently under the hood. We just need to make sure we've seen all rows in the end.

query/physicalplan/aggregate.go
```diff
@@ -410,19 +417,46 @@ func (a *HashAggregate) Finish(ctx context.Context) error {
...
+		}
+	default:
+		aggregateArray, err = a.aggregationFunction.Aggregate(a.pool, arrs)
+	}
 	if err != nil {
 		return fmt.Errorf("aggregate batched arrays: %w", err)
 	}
 
-	aggregateField := arrow.Field{Name: a.resultColumnName, Type: aggregateArray.DataType()}
+	// Only the last Aggregation should rename the underlying column
```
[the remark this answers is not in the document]
metalmatze: Yes, exactly right. This should actually be part of the projection. I can follow up on this and create an issue from this comment. WDYT?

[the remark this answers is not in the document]
asubiotto: SGTM

table.go
```diff
@@ -644,54 +657,93 @@ func (t *Table) Iterator(
...
-				r := converter.NewRecord()
-				prepareForFlush(r, rg.Schema())
-				if err := callback(ctx, r); err != nil {
-					return err
+	errg, ctx := errgroup.WithContext(ctx)
+
+	for _, callback := range callbacks {
+		callback := callback
+		errg.Go(func() error {
+			var builder *array.RecordBuilder
```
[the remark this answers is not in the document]
metalmatze: Oh, yes. I had to copy the `distinct.go` yesterday as well... I messed something up in the rebase 😢

query/physicalplan/synchronize.go
metalmatze: Yes, I had that in between implementing but reverted it since there were too many moving things at once. That 💯 a follow up for this PR.

PR #160, opened 2022-08-12
Replace atomic unsafe pointers with generic atomic pointers

Mihai22125:
Replace atomic unsafe pointers with generic atomic pointers.

First-time contributor, so any feedback is more than welcome!

This resolves: #157.

PR #203, opened 2022-09-22
Atomic fixes

thorfour:
This is all @Mihai22125 work from https://github.com/polarsignals/frostdb/pull/160

Just updated the ci files to use 1.19, and rebased on top of main.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,7 +27,6 @@ linters-settings:
     list-type: blacklist
     include-go-root: true
     packages-with-error-message:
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -27,7 +27,6 @@ linters-settings:
     list-type: blacklist
     include-go-root: true
     packages-with-error-message:
-      - sync/atomic: "Use go.uber.org/atomic instead of sync/atomic"
```
metalmatze, line 30: We could do it the other way around now?

PR #214, opened 2022-09-27
Fix writer partition

thorfour:
Introduces a table level option of `WithRowGroupSize` which informs the writer when to `Flush()` the buffer and write out a RowGroup. 

It uses this setting both during granule compaction, as well as block persistence.

### DIFF
--- a/granule.go
+++ b/granule.go
@@ -73,7 +73,7 @@ func (g *Granule) addPart(p *Part, r *dynparquet.DynamicRow) (uint64, error) {
 	}
 	node := g.parts.Prepend(p)
 
-	newSize := g.metadata.size.Add(uint64(p.Buf.ParquetFile().NumRows()))
+	newSize := g.metadata.size.Add(uint64(p.Buf.ParquetFile().Size()))
 
 	for {
 		least := g.metadata.least.Load()
--- a/table.go
+++ b/table.go
@@ -128,6 +128,7 @@ type TableBlock struct {
 type tableMetrics struct {
 	blockRotated              prometheus.Counter
 	granulesCreated           prometheus.Counter
+	compactions               prometheus.Counter
 	granulesSplits            prometheus.Counter
 	rowsInserted              prometheus.Counter
 	zeroRowsInserted          prometheus.Counter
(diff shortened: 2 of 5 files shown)


granule.go
```diff
@@ -73,7 +73,7 @@ func (g *Granule) addPart(p *Part, r *dynparquet.DynamicRow) (uint64, error) {
 	}
 	node := g.parts.Prepend(p)
 
-	newSize := g.metadata.size.Add(uint64(p.Buf.ParquetFile().NumRows()))
+	newSize := g.metadata.size.Add(uint64(p.Buf.ParquetFile().Size()))
```
asubiotto, line 76: Oof, nice catch

table.go
```diff
@@ -54,11 +54,22 @@ func (e ErrCreateSchemaWriter) Error() string {
...
 
 type TableConfig struct {
-	schema *dynparquet.Schema
+	schema       *dynparquet.Schema
+	rowGroupSize int
+}
+
+type TableOption func(*TableConfig) error
+
+func WithRowGroupSize(numRows int) TableOption {
```
asubiotto, line 63: Should this be specified in bytes? It's harder to reason about where exactly row groups get split, but other observations (e.g. expected size over the wire) are easier to do. Additionally, it seems like bytes would be more in line with the other knobs (active block size).

nit: I would also add a quick comment to describe what behavior is triggered when a row group reaches this size

table.go
```diff
@@ -1384,44 +1375,11 @@ func (t *TableBlock) Serialize(writer io.Writer) error {
...
-	if err != nil {
-		return err
-	}
-
 	rows := merged.Rows()
 	defer rows.Close()
+	total := 0
 	for {
-		rowsBuf := make([]parquet.Row, count)
+		rowsBuf := make([]parquet.Row, 1)
```
asubiotto, line 1382: I think we should increase the size of this read slice buffer so we can read many rows at a time instead of just 1. Row-by-row seems like it'll be unnecessarily slow.

brancz: Agreed. We actually already know how many we want so we could make this buffer exactly `rowGroupSize` large.

brancz (approved):

table.go
thorfour: addressing in: https://github.com/polarsignals/frostdb/pull/215

PR #215, opened 2022-09-28
buffer the number of rows for a row group if given

thorfour:
Addressing comment from https://github.com/polarsignals/frostdb/pull/214

### DIFF
--- a/table.go
+++ b/table.go
@@ -1375,11 +1375,16 @@ func (t *TableBlock) writeRows(writer io.Writer, rowGroups []dynparquet.DynamicR
 	defer t.table.config.schema.PutWriter(w)
 	defer w.Close()
 
+	buffSize := 1
+	if t.table.config.rowGroupSize > 0 {
+		buffSize = t.table.config.rowGroupSize
+	}
+
 	rows := merged.Rows()
 	defer rows.Close()
 	total := 0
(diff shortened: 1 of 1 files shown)


table.go
```diff
@@ -1375,11 +1375,16 @@ func (t *TableBlock) writeRows(writer io.Writer, rowGroups []dynparquet.DynamicR
 	defer t.table.config.schema.PutWriter(w)
 	defer w.Close()
 
+	buffSize := 1
```
asubiotto, line 1378: What do you think of defaulting to something like `256`?

asubiotto (approved):

table.go
thorfour: yea I wasn't sure what a 'sane' default was which is why I left it at 1. But I'm happy to bump it to 256 if we want?

table.go
```diff
@@ -1391,7 +1396,7 @@ func (t *TableBlock) writeRows(writer io.Writer, rowGroups []dynparquet.DynamicR
 		if _, err = w.WriteRows(rowsBuf); err != nil {
```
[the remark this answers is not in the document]
thorfour: I'll add it for safety, but the way the writer code works it doesn't matter

PR #218, opened 2022-10-04
pqarrow/builder: optimize Parquet to Arrow conversion

asubiotto:
This PR adds the concept of optimized builders to FrostDB. These are array.Builders that can build arrow arrays more efficiently than the default builders by only copying when necessary. This is especially important when handling data like variable-sized strings, where we'd go through the following process:
- Read parquet strings stored in flat format, create a two-dimensional slice of values representing these.
- Append values to an arrow builder, which reads each string and converts it back to flat format.

With the new builder, we can immediately just copy the flat data + offsets and cast them to an arrow array when necessary. The improvements are noticeable especially in QueryRange:
```
QueryTypes-8    3.84ms ± 3%    3.79ms ± 3%     ~     (p=0.151 n=5+5)
QueryMerge-8    12.9ms ± 2%    12.8ms ± 6%     ~     (p=0.548 n=5+5)
QueryRange-8    5.23ms ± 6%    4.49ms ± 2%  -14.26%  (p=0.016 n=4+5)

QueryTypes-8    5.92MB ± 0%    5.69MB ± 0%   -3.90%  (p=0.008 n=5+5)
QueryMerge-8    90.9MB ± 0%    81.9MB ± 0%   -9.90%  (p=0.008 n=5+5)
QueryRange-8    28.8MB ± 0%    23.7MB ± 0%  -17.81%  (p=0.008 n=5+5)

QueryTypes-8     97.6k ± 0%     93.4k ± 0%   -4.31%  (p=0.008 n=5+5)
QueryMerge-8      303k ± 0%      301k ± 0%   -0.68%  (p=0.008 n=5+5)
QueryRange-8     59.8k ± 0%     58.0k ± 0%   -3.08%  (p=0.008 n=5+5)
```

The nice thing about these builders is that they give us freedom to manipulate memory as we wish (see ergonomic improvements to appending nulls and repeating last values during distinct optimizations). This is dangerous, but nice for performance reasons. I envision these builders to be used in operators in the future (which is what I think my next step is).

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -602,7 +602,6 @@ func (c *ParquetConverter) writeColumnToArray(
 	dictionaryOnly bool,
 	w writer.ValueWriter,
 ) error {
-	optional := n.Optional()
 	repeated := n.Repeated()
 	if !repeated && dictionaryOnly {
 		// Check all the page indexes of the column chunk. If they are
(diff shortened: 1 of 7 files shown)


brancz (approved):
Incredible work! 💯

thorfour (approved):

asubiotto:
TFTRs!

PR #225, opened 2022-10-07
logictest: move distinct tests to logictest format

asubiotto:


asubiotto:
We're seeing the PR fail because of ordering issues where tests have different row groups. This is because we don't enforce ordering on the row groups scanned when concurrency is enabled. Will need to fix the testing harness to not care about ordering.

asubiotto:
OK, I just added another commit to ignore output ordering. If CI ends up green, this is RFAL.

metalmatze (approved):
Amazing stuff! :rocket:

asubiotto:
TFTR!

PR #229, opened 2022-10-12
Range over a list instead of map

thorfour:
Since maps return a random ordering, this screws up the record schema ordering, which results in bad aggregations


brancz (approved):
Nice find!

PR #231, opened 2022-10-13
frostdb: add compactorPool to limit compaction goroutines

### LINKED ISSUE #119: Recycle goroutine doing compaction
While working on https://github.com/polarsignals/frostdb/pull/118 I realised that we create a go-routine per compaction for each granule.

I think this should be bounded but also we should probably recycle go-routine since those can grow pretty large in term of memory

image

I suggest we create 8 go-routines that will take care of compactions, then send granule to compact via  a buffered channel of 32, meaning it can block on the write path.

asubiotto:
This commit adds a worker pool of goroutines (`runtime.NumCPU` by default) per database that will iterate over all the tables in that database and attempt to perform a compaction every defined sweep interval (`100ms` by default).

The compaction logic has not been modified.

This PR also adds a `compacting` atomic boolean to the granule so that no two goroutines attempt to compact the same granule. Previously, I think this was the case because the only atomic that was checked was `pruned`, which is only set once the granule has been pruned from the index.

Closes #119

### DIFF
--- a/granule.go
+++ b/granule.go
@@ -28,13 +28,15 @@ type Granule struct {
 // GranuleMetadata is the metadata for a granule.
 type GranuleMetadata struct {
 	// least is the row that exists within the Granule that is the least.
-	// This is used for quick insertion into the btree, without requiring an iterator
+	// This is used for quick insertion into the btree, without requiring an
+	// iterator.
 	least atomic.Pointer[dynparquet.DynamicRow]
 
-	// size is the raw commited, and uncommited size of the granule. It is used as a suggestion for potential compaction
+	// size is the raw committed, and uncommitted size of the granule. It is
+	// used as a suggestion for potential compaction.
 	size *atomic.Uint64
 
-	// pruned indicates if this Granule is longer found in the index
+	// pruned indicates if this Granule is longer found in the index.
 	pruned *atomic.Uint64
 }
 
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -82,9 +82,6 @@ func TestAggregate(t *testing.T) {
 	_, err = table.InsertBuffer(context.Background(), buf)
 	require.NoError(t, err)
 
-	// Ensure all transactions are completed
-	table.Sync()
-
 	engine := query.NewEngine(
 		memory.NewGoAllocator(),
 		db.TableProvider(),
--- a/bench_test.go
+++ b/bench_test.go
@@ -53,7 +53,9 @@ func newDBForBenchmarks(ctx context.Context, b testing.TB) (*ColumnStore, *DB, e
 	if err != nil {
 		return nil, nil, err
 	}
-	table.Sync()
+	if err := table.EnsureCompaction(); err != nil {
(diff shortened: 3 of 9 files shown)


granule.go
```diff
@@ -28,14 +28,22 @@ type Granule struct {
...
 	size *atomic.Uint64
 
-	// pruned indicates if this Granule is longer found in the index
+	// pruned indicates if this Granule is longer found in the index.
 	pruned *atomic.Uint64
+
+	// compacting indicates if this Granule is currently being compacted.
+	// TODO(asubiotto): Is this necessary? We append a sentinel node to the
+	// parts list when compacting, but I'm unsure how to "read" that without
+	// appending a part. What's the intention?
```
brancz, line 45: @thorfour knows this best

asubiotto:
I think that's a great idea @maxbrunet, thanks for the suggestion.

granule.go
thorfour: This isn't necessary, granules have the (now poorly named) `pruned` variable that prevents multiple go routines from compacting the same granule

thorfour (approved):
I think overall I'm fine with this change, my only question is if we considered instead of doing a sweep, we do a notify? Where after adding to a granule and determining it requires compaction we send that granule pointer over on a channel to the compaction pool? Notify might be more responsive, obviously does have the drawback of potentially blocking on the channel writes.

granule.go
asubiotto: Ah, gotcha! Yes I guess I was confused by the naming and somehow my brain read the CAS operation at the top of compaction to just be a load operation (with the assumption that `pruned` was set after pruning). Might rename the variable then.

asubiotto:
> I think overall I'm fine with this change, my only question is if we considered instead of doing a sweep, we do a notify? Where after adding to a granule and determining it requires compaction we send that granule pointer over on a channel to the compaction pool? Notify might be more responsive, obviously does have the drawback of potentially blocking on the channel writes.

Yep! #119 has some discussion on this and the main takeaway is that we want to keep the insert path non-blocking. Another option is to use condition variables, but those still acquire locks to signal and it'd be to trigger a sweep rather than sending the pointer anywhere. I think an interval sweep style is probably a good approach for now since it's tunable.

granule.go
thorfour: Yea it used to be... We should rename it

thorfour:
> > I think overall I'm fine with this change, my only question is if we considered instead of doing a sweep, we do a notify? Where after adding to a granule and determining it requires compaction we send that granule pointer over on a channel to the compaction pool? Notify might be more responsive, obviously does have the drawback of potentially blocking on the channel writes.
> 
> Yep! #119 has some discussion on this and the main takeaway is that we want to keep the insert path non-blocking. Another option is to use condition variables, but those still acquire locks to signal and it'd be to trigger a sweep rather than sending the pointer anywhere. I think an interval sweep style is probably a good approach for now since it's tunable.

Sounds great to me

Issue #223, opened 2022-10-07
Reduce compaction memory usage by marking Parts as compacted


asubiotto:
There are two goals during `Granule` compaction:
1) Reduce the `Granule` size (i.e. the sum of the serialized size of all the parquet files contained in this granule).
2) Since users can insert in whatever size batches they want, we would like to "reshape" the row groups to the desired size. i.e. if we do n 1 row inserts, merge all of those into a single n-row rowgroup to take advantage of parquet compression benefits.

The current compaction strategy is to trigger a compaction on a granule during Insertions if the data added to a `Granule` makes this `Granule`'s size exceed a threshold. If this condition is met, a goroutine is spawned to compact this granule. All that granule's `Part`s (essentially a wrapper over a `parquet.File`) will be deserialized and merge-sorted according to the sorting columns (using `parquet.MergeRowGroups`) and rewritten in serialized format to an in-memory buffer that will become a single `Part` of a new granule. If the new granule is still above the target size (note that merging multiple `Part`s into one takes advantage of parquet compression, so this size should theoretically be lower on the first compaction of a `Granule`), the new `Granule` will be split into two new `Granule`s. This requires another cycle of deserialization and reserialization because it is the only way to split a single `Part`.

This compaction strategy achieves our stated goals, but is inefficient. We should mark a `Part` when it is the result of a compaction. This implies that the row groups contained within this `Part` are already the desired row group size. We could then enhance our existing compaction strategy to not deserialize these parts, avoiding the memory usage this entails. For any new compactions of an already compacted `Granule`, only the uncompacted `Part`s will be deserialized and reserialized into a single `Part`. When splitting granules, we would only reassign compacted parts (ideally maintaining sorted order), which would also avoid the memory usage of deserializing the `Part` to split during a split.

asubiotto:
Taking this on

asubiotto:
Implemented in #234

PR #234, opened 2022-10-19
frostdb: add leveled compaction

asubiotto:
This PR implements simple leveled compaction of parts into level 1 parts as described in #223, which are guaranteed to never be overlapping. This will hopefully improve our memory usage during compactions and avoid the spikes we've seen in our clusters. Please refer to the individual commits for details. I've tried to split it up into 1) The addition of leveled compaction, 2) Handling out-of-order inserts which should only happen very rarely but it is an edge case we need to handle.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -10,48 +10,88 @@ import (
 	"sync"
 	"time"
 
+	"github.com/dustin/go-humanize"
+	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/google/btree"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/thanos-io/objstore/errutil"
 
 	"github.com/polarsignals/frostdb/dynparquet"
 )
 
+type CompactionOption func(*CompactionConfig) error
+
+type CompactionConfig struct {
+	concurrency          int
+	interval             time.Duration
+	l1ToGranuleSizeRatio float64
+}
+
+// NewCompactionConfig creates a new compaction config with the given options.
+// If none are specified, a default compaction config is created.
+func NewCompactionConfig(options ...CompactionOption) *CompactionConfig {
+	c := &CompactionConfig{
+		concurrency: runtime.GOMAXPROCS(0),
+		interval:    100 * time.Millisecond,
+		// 0.5 was chosen so that a level1 part takes up around half the space
+		// in a full granule, allowing for level0 parts to grow up to the
+		// remaining space before compaction. The higher this ratio, the less
+		// new data is compacted and the more compactions occur. However, a
+		// lower ratio implies leaving some parquet compression size savings on
+		// the table.
+		l1ToGranuleSizeRatio: 0.5,
+	}
+	return c
+}
+
+// WithConcurrency specifies the number of concurrent goroutines compacting data
+// for each database.
+func WithConcurrency(concurrency int) CompactionOption {
+	return func(c *CompactionConfig) error {
+		c.concurrency = concurrency
+		return nil
+	}
+}
+
+// WithInterval specifies the compaction sweep interval.
+func WithInterval(i time.Duration) CompactionOption {
+	return func(c *CompactionConfig) error {
+		c.interval = i
+		return nil
+	}
+}
+
+// WithL1ToGranuleSizeRatio sets the target level1 part size relative to the
+// granule size. The closer this value is to 1, the more compacted data becomes
+// with an expected rise in memory and CPU usage.
+func WithL1ToGranuleSizeRatio(r float64) CompactionOption {
+	return func(c *CompactionConfig) error {
+		c.l1ToGranuleSizeRatio = r
+		return nil
+	}
+}
+
 type compactorPool struct {
 	db *DB
 
-	concurrency int
-	interval    time.Duration
-	wg          sync.WaitGroup
-	cancel      context.CancelFunc
+	cfg    *CompactionConfig
+	wg     sync.WaitGroup
+	cancel context.CancelFunc
 }
 
-var (
-	defaultConcurrency   = runtime.GOMAXPROCS(0)
-	defaultSweepInterval = 100 * time.Millisecond
-)
-
-func newCompactorPool(db *DB) *compactorPool {
-	concurrency := defaultConcurrency
-	if override := db.columnStore.compactionOptions.concurrency; override != 0 {
-		concurrency = override
-	}
-
-	sweepInterval := defaultSweepInterval
-	if override := db.columnStore.compactionOptions.sweepInterval; override != 0 {
-		sweepInterval = override
-	}
+func newCompactorPool(db *DB, cfg *CompactionConfig) *compactorPool {
 	return &compactorPool{
-		db:          db,
-		concurrency: concurrency,
-		interval:    sweepInterval,
+		db:  db,
+		cfg: cfg,
 	}
 }
 
 func (c *compactorPool) start() {
 	ctx, cancelFn := context.WithCancel(context.Background())
 	c.cancel = cancelFn
-	for i := 0; i < c.concurrency; i++ {
+	for i := 0; i < c.cfg.concurrency; i++ {
 		c.wg.Add(1)
 		go func() {
 			defer c.wg.Done()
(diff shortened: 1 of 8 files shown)


asubiotto:
Converting to draft until I investigate why this has made some things so much worse.

asubiotto:
By "much worse" I was referring to benchmark performance. However I think this compaction strategy already makes things better. Based on my investigation, we were never actually running a compaction on main during these benchmarks. This is because we would never unset `pruned` on a granule in cases where only parts with uncompleted txns were found (this is the case of compacting during WAL replays). Since `pruned` was already set, the granule would never be compacted.

It turns out the inserts are actually very well batched. Here is a comparison of main with and without our current compaction strategy (simply adding a `t.abort(granule)` in the case of `len(bufs) == 0` in `compactGranule`):
```
name           old time/op    new time/op     delta
Query/Types-8    3.77ms ± 1%   641.41ms ±93%  +16929.55%  (p=0.000 n=10+10)
Query/Merge-8    12.4ms ± 1%     27.9ms ± 2%    +124.03%  (p=0.000 n=9+9)
Query/Range-8    4.04ms ± 0%    14.10ms ± 1%    +249.33%  (p=0.000 n=8+10)

name           old alloc/op   new alloc/op    delta
Query/Types-8    5.69MB ± 0%  5353.70MB ±91%  +93949.13%  (p=0.000 n=10+10)
Query/Merge-8    74.9MB ± 0%    162.8MB ± 0%    +117.36%  (p=0.000 n=9+10)
Query/Range-8    16.3MB ± 0%     58.5MB ± 0%    +258.73%  (p=0.000 n=10+10)

name           old allocs/op  new allocs/op   delta
Query/Types-8     93.4k ± 0%   2495.5k ±359%        ~     (p=0.720 n=10+9)
Query/Merge-8      301k ± 0%       204k ± 0%     -32.32%  (p=0.000 n=9+10)
Query/Range-8     57.7k ± 0%       7.8k ± 0%     -86.49%  (p=0.000 n=10+10)
```

Now compare the leveled compaction strategy with the our current compaction strategy:
```
Query/Types-8     641ms ±93%      42ms ± 6%  -93.46%  (p=0.000 n=10+10)
Query/Merge-8    27.9ms ± 2%    22.3ms ± 1%  -19.81%  (p=0.000 n=9+10)
Query/Range-8    14.1ms ± 1%    10.3ms ± 0%  -27.25%  (p=0.000 n=10+10)

Query/Types-8    5.35GB ±91%    0.39GB ± 0%  -92.63%  (p=0.000 n=10+10)
Query/Merge-8     163MB ± 0%     119MB ± 0%  -26.90%  (p=0.000 n=10+10)
Query/Range-8    58.5MB ± 0%    36.7MB ± 0%  -37.27%  (p=0.000 n=10+10)

Query/Types-8    2.50M ±359%     0.02M ± 0%  -99.33%  (p=0.004 n=9+9)
Query/Merge-8      204k ± 0%      203k ± 0%   -0.13%  (p=0.000 n=10+10)
Query/Range-8     7.80k ± 0%     7.54k ± 0%   -3.32%  (p=0.000 n=10+10)
```

The leveled compaction strategy still performs worse than no compactions but is definitely better than the current compaction strategy. For reference here is no compactions vs leveled compaction:
```
Query/Types-8    3.77ms ± 1%   41.94ms ± 6%  +1013.38%  (p=0.000 n=10+10)
Query/Merge-8    12.4ms ± 1%    22.3ms ± 1%    +79.65%  (p=0.000 n=9+10)
Query/Range-8    4.04ms ± 0%   10.26ms ± 0%   +154.13%  (p=0.000 n=8+10)

Query/Types-8    5.69MB ± 0%  394.45MB ± 0%  +6829.39%  (p=0.000 n=10+10)
Query/Merge-8    74.9MB ± 0%   119.0MB ± 0%    +58.89%  (p=0.000 n=9+10)
Query/Range-8    16.3MB ± 0%    36.7MB ± 0%   +125.03%  (p=0.000 n=10+10)

Query/Types-8     93.4k ± 0%     16.7k ± 0%    -82.18%  (p=0.000 n=10+9)
Query/Merge-8      301k ± 0%      203k ± 0%    -32.41%  (p=0.000 n=9+10)
Query/Range-8     57.7k ± 0%      7.5k ± 0%    -86.94%  (p=0.000 n=10+10)
```

My explanation for this is that `QueryTypes` worsened a lot because distinct performance improvements cannot be taken advantage of after reshaping the data during compaction (still to be confirmed). Merge and Range I am not quite sure of, but it's probably due to a similar explanation.

I think that I will add some improvements to this PR that I thought about when replaying WALs (writing l1s iteratively from a single merged row group of all l0s) and advocate to merge this given its improvement over our current compaction strategy.

It's clear that we have to do more work when it comes to "shaping" parts so that the scan level can reach the non-compacted latencies, but that is left for future work. The important part is that this compaction is an improvement over our current compaction strategy.

compaction.go
```diff
@@ -100,10 +106,433 @@ func (t *TableBlock) compact() error {
...
+	)
+	if err != nil {
+		return err
+	}
+
+	if len(level0Parts) == 0 && len(level1Parts) == 0 {
+		// Edge case in which only parts with uncompleted txns are found and
+		// they are already bigger than the max size. A future GC cycle will
+		// take care of this, so abort the compaction and return.
+		t.abortCompaction(granule)
```
thorfour, line 194: `nit` prefer aborting in a single place than the potential to abort in multiple places. Can we instead return an error here and have the caller abort like we do for the rest of the error cases?

asubiotto:
This is now ready for a look! Add a commit with the new "cursor" strategy which simplifies things a bit. The performance hit from no compactions has also been mitigated:
```
Query/Types-8    3.77ms ± 1%   11.07ms ± 3%   +193.88%  (p=0.000 n=10+10)
Query/Merge-8    12.4ms ± 1%    20.4ms ± 0%    +63.69%  (p=0.000 n=9+7)
Query/Range-8    4.04ms ± 0%    8.32ms ± 1%   +106.15%  (p=0.000 n=8+9)

Query/Types-8    5.69MB ± 0%   92.23MB ± 0%  +1520.17%  (p=0.000 n=10+8)
Query/Merge-8    74.9MB ± 0%    93.8MB ± 1%    +25.26%  (p=0.000 n=9+10)
Query/Range-8    16.3MB ± 0%    27.0MB ± 0%    +65.72%  (p=0.000 n=10+10)

Query/Types-8     93.4k ± 0%     24.2k ± 0%    -74.13%  (p=0.000 n=10+9)
Query/Merge-8      301k ± 0%      204k ± 0%    -32.12%  (p=0.000 n=9+10)
Query/Range-8     57.7k ± 0%      8.4k ± 0%    -85.44%  (p=0.000 n=10+10)
```

asubiotto:
Before merging, I'd like to add some logging/metrics.

thorfour (approved):
This looks great! Outstanding work!


I agree with adding some metrics around the compaction, but otherwise this looks great.

compaction.go
asubiotto: It's a good point. I did think about returning the error here but it would be logged as a warning by the higher level compactor pool and I wasn't keen on putting error code inspection in there. I ended up considering this as a successful "edge case" and aborting here. I'm not 100% sure this is the right path so I'd be keen on getting your perspective.

asubiotto:
I made some modifications (level 1 target size and size to split granule) and ran these new compaction changes on a local parca instance ingesting profiles from demo and it stabilized performance significantly. In the following graphs, I ran the old compactor until a block rotation and then ran the new compactor. The new compactor starts at 12:05 on the graph, so basically where all curves drop back down midway through:

Here is the p50 of Query (Merge/Single):
image

p50 of QueryRange:
image

p90 latencies are also relatively more stable:
image
image

The downside is that the active block is rotated earlier than with the old compactor since the old compactor recompacted everything.
image

In general, we also distribute the parts evenly to avoid excessive compactions:
image

This probably means that we could probably be doing more compactions while not affecting query latency/stability. I think I'm fine merging this as a first version and then iterating with the parameters to find what works for us.

compaction.go
asubiotto: I ended up returning a boolean to accompany the error.

asubiotto:
I think I'd appreciate a last review/sanity check for the metrics etc... and can merge on approval.

thorfour (approved):
It might be nice to have a way to configure the level of compaction? i.e tune it to full compaction, or off entirely.

PR #236, opened 2022-10-20
frostdb: move all BenchmarkQuery* benchmarks into one benchmark

asubiotto:
This commit moves BenchmarkQuery{Types,Merge,Range} into a single top-level BenchmarkQuery that then runs each scenario in a sub-benchmark. This allows setting up the db only once and then sharing it across benchmarks. Doing this seems to also reduce the number of times the setup code is run to one time. Previously, it seems like golang's benchmark runtime would run setup multiple times.
```
name           time/op
Query/Types-8  3.77ms ± 1%
Query/Merge-8  12.4ms ± 1%
Query/Range-8  4.04ms ± 0%

name           alloc/op
Query/Types-8  5.69MB ± 0%
Query/Merge-8  74.9MB ± 0%
Query/Range-8  16.3MB ± 0%

name           allocs/op
Query/Types-8   93.4k ± 0%
Query/Merge-8    301k ± 0%
Query/Range-8   57.7k ± 0%
```

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -30,8 +30,6 @@ const (
 func newDBForBenchmarks(ctx context.Context, b testing.TB) (*ColumnStore, *DB, error) {
 	b.Helper()
 
-	b.Logf("initializing %s", b.Name())
-
 	col, err := New(
 		WithWAL(),
 		WithStoragePath(storagePath),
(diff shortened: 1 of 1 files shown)


asubiotto:
And this is what the new log lines show:
```
BenchmarkQuery
    bench_test.go:119: replaying WAL
    bench_test.go:119: replayed WAL in 15.212137875s
    bench_test.go:119: ensuring compaction
    bench_test.go:119: ensured compaction in 792ns
    bench_test.go:119: db initialized and WAL replayed, starting benchmark BenchmarkQuery
```

brancz:
Very nice!

PR #238, opened 2022-10-25
deps: update parquet to 63215c8

asubiotto:


brancz:
Is there something specific we’re pulling in with this?

asubiotto:
Yep, sorry I thought it was in the commit message. https://github.com/segmentio/parquet-go/pull/371 should reduce the huge buffer allocations we've seen during compaction a bit.

PR #243, opened 2022-11-03
Stop writes on close

thorfour:
This change
- prevents further writes from happening after a table has been closed.
- adds a histogram to record how long it takes a database to shutdown.
- shuts down databases concurrently

### DIFF
--- a/db.go
+++ b/db.go
@@ -226,7 +226,8 @@ func (s *ColumnStore) ReplayWALs(ctx context.Context) error {
 }
 
 type dbMetrics struct {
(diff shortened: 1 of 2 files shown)


brancz (approved):
lgtm!

Since closing happens most often on shutdown, it probably makes sense to add two more counters to those metrics 1 for shutdowns started and 1 for shutdowns finished.

thorfour:
> lgtm!
> 
> Since closing happens most often on shutdown, it probably makes sense to add two more counters to those metrics 1 for shutdowns started and 1 for shutdowns finished.

Good call, I'll add that

PR #249, opened 2022-11-17
fixes a panic caused by a nil active block

thorfour:


table.go
```diff
@@ -242,13 +242,19 @@ func newTable(
 		Name: "index_size",
 		Help: "Number of granules in the table index currently.",
 	}, func() float64 {
+		if t.ActiveBlock() == nil {
```
brancz, line 245: Wouldn't we want to save the result of `t.ActiveBlock` to a variable and test it for nil? Otherwise by the below call to `t.ActiveBlock` might be a different value again, no?

thorfour: it's only nil during initialization
so it can't go backwards to nil
but it will hedge against that in the future so wil change

PR #251, opened 2022-11-21
Add tracing to TableBlock/RowGroupIterator

metalmatze:
Especially then row group event may be useful here.

### DIFF
--- a/table.go
+++ b/table.go
@@ -134,6 +134,7 @@ type WAL interface {
 type TableBlock struct {
 	table  *Table
 	logger log.Logger
+	tracer trace.Tracer
 
 	ulid   ulid.ULID
 	minTx  uint64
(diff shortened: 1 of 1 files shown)


table.go
```diff
@@ -955,6 +957,10 @@ func (t *TableBlock) RowGroupIterator(
 	filter TrueNegativeFilter,
 	rowGroups chan<- dynparquet.DynamicRowGroup,
 ) error {
+	ctx, span := t.tracer.Start(ctx, "TableBlock/RowGroupIterator")
+	span.SetAttributes(attribute.Int64("tx", int64(tx))) // Attributes don't support uint64...
```
brancz, line 961: I'd be proud if we reach the max int64 :P

brancz (approved):

PR #252, opened 2022-11-22
Aggregation projection

thorfour:
This updates the query engine to support two things:
- BinaryExpr Projections of Aggregation columns
- Alias of BinaryExpr projections

ex:
```
	Aggregate(
			logicalplan.Sum(logicalplan.Col("value")),
			logicalplan.DynCol("labels"),
			logicalplan.Col("timestamp"),
		).
		Project(
			logicalplan.Col(logicalplan.Sum(logicalplan.Col("value")).Name()),
			logicalplan.DynCol("labels"),
			logicalplan.Col("timestamp").Gt(logicalplan.Literal(1)).Alias("timestamp"),
		).
```

Where we project the timestamp column from the aggregation to a `timestamp > 1` column and then rename that column to `timestamp`

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -351,3 +351,101 @@ func TestAggregateInconsistentSchema(t *testing.T) {
 		})
 	}
 }
+
+func Test_Aggregation_Projection(t *testing.T) {
+	config := NewTableConfig(
+		dynparquet.NewSampleSchema(),
+	)
+
+	logger := newTestLogger(t)
+
+	c, err := New(
+		WithLogger(logger),
+	)
+	require.NoError(t, err)
+	defer c.Close()
+	db, err := c.DB(context.Background(), "test")
+	require.NoError(t, err)
+	table, err := db.Table("test", config)
+	require.NoError(t, err)
+
+	samples := dynparquet.Samples{{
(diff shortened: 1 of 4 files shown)


aggregate_test.go
```diff
@@ -351,3 +351,101 @@ func TestAggregateInconsistentSchema(t *testing.T) {
 		})
 	}
 }
+
+func Test_Aggregation_Projection(t *testing.T) {
```
asubiotto, line 355: Can you write a logic test instead of a go test for this?

query/logicalplan/expr.go
```diff
@@ -99,8 +99,8 @@ func (e *BinaryExpr) Computed() bool {
 	return true
 }
 
-func (e *BinaryExpr) Alias(alias string) AliasExpr {
-	return AliasExpr{Expr: e, Alias: alias}
+func (e *BinaryExpr) Alias(alias string) *AliasExpr {
```
asubiotto, line 102: Does this need to be a pointer?

thorfour: Yes, all the `AliasExpr` functions are pointer receivers so I believe it was a typo to not have it be one already.

asubiotto (approved):

Issue #239, opened 2022-10-25
Compaction memory usage spike


asubiotto:
We see some unexpected memory spikes during compaction (this is a local demo instance scraping itself):
![](https://cdn.discordapp.com/attachments/963045173535834182/1034366857869799454/unknown.png)
In the graph above, there is a ~1GB memory spike that is attributed to compaction. Matching this with the compaction that occurred during this time in the logs (debug level), we get:
```
level=debug db=parca name=parca ts=2022-10-25T06:57:11.229938Z caller=compaction.go:187 msg="compaction complete" duration=3.117003833s l0Before="[sz=21 MiB,cnt=483]" l1Before="[sz=4.3 MiB,cnt=1]" l1After="[sz=5.9 MiB,cnt=1]" overlaps=1 splits=0
```
So this ~1GB of allocations was a result of merging 483 parts that sum up to 21 MiB and 1 4.3MiB part. Given parquet decompression, we do expect memory usage to be higher than the l0+l1 size, but 1GB seems like too much. One theory of what is going on is that buffer pooling is not working as expected in parquet (https://github.com/segmentio/parquet-go/issues/372).

gernest:
Has this been resolved yet? I would lie to address it, whether here or on `parquet-go` side, any pointers to lead me in the right direction will be highly appreciated.

asubiotto:
Yes, this has been resolved!

PR #253, opened 2022-11-22
recursive merge the dynamic buffers

thorfour:
Much of the compaction memory spikes noted in #239 are due to opening all the pages in all the row groups to sort them. If we instead perform a recursive merge sort we only need to open 2 pages at a time keeping the memory usage from spiking.

image

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -548,7 +548,7 @@ func compactLevel0IntoLevel1(
 	}
 
 	cursor := 0
-	merged, err := t.table.config.schema.MergeDynamicRowGroups(bufs)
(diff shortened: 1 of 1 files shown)


asubiotto:
I need to check again, but I'm wondering if we need this recursive merging with the cursor change you discovered (https://github.com/polarsignals/frostdb/pull/254). Given that this now means that memory usage is a function of the number of parts in a granule and we aim to reduce this using leveled compaction (i.e. we'll trend towards fewer parts that need merging since there'll be less space for new l0 parts), I wonder what the memory usage looks like now.

PR #254, opened 2022-11-23
frostdb: use parquet.MultiRowGroup to reduce compaction memory usage

asubiotto:
Inspired by @thorfour's recent changes to our cloud product to use a MultiRowGroup to wrap multiple sorted row groups in a single row group, this change does the same for compaction in FrostDB.

The idea is to reduce the number of decompressed pages held in memory at any one time, which should drastically reduce memory usage. Given that we know that all row groups within a part are sorted, exposing a part as a single row group lets us have one cursor per part instead of one cursor per row group per part.


brancz:
Looking forward to seeing the effect of this!

PR #256, opened 2022-11-24
Support boolean types

thorfour:
This adds supports for boolean type columns


table_test.go
```diff
@@ -973,3 +973,73 @@ func Test_Table_EmptyRowGroup(t *testing.T) {
 	})
 	require.NoError(t, err)
 }
+
+func Test_BoolColumn(t *testing.T) {
```
brancz, line 977: Should this be a logical test?

pqarrow/writer/writer.go
```diff
@@ -239,3 +239,62 @@ func (w *float64ValueWriter) WritePage(p parquet.Page) error {
...
 }
+
+type booleanValueWriter struct {
+	b   *array.BooleanBuilder
+	buf []bool
+}
+
+func NewBooleanValueWriter(b builder.ColumnBuilder, numValues int) ValueWriter {
+	res := &booleanValueWriter{
+		b: b.(*array.BooleanBuilder),
```
asubiotto, line 250: I think it would be nice to write an `OptBoolBuilder` and use that for efficient parquet to arrow conversion. Can be added in the future but I think it's worth exploring now since I don't think it should be too much code. See `optbuilders.go` and especially `OptInt64Builder`

metalmatze (commented):
Seems good. Nothing to add on top of what has been said.

table_test.go
thorfour: Done.

pqarrow/writer/writer.go
thorfour: Done

brancz (approved):
lgtm

Issue #226, opened 2022-10-07
Move operator tests to logic tests


asubiotto:
A datadriven logic testing framework was recently added in https://github.com/polarsignals/frostdb/pull/211. Slowly but surely, it would be nice to move our operator tests to this logic testing framework to promote readability and conciseness. #255 does this for the distinct operator. This issue can be closed once the following tests have been moved over:
- [ ] aggregate_test.go
- [x] filter_test.go

bitethecode:
@asubiotto Can you assign this to me?

asubiotto:
Sure!

bitethecode:
@asubiotto Thanks!!!

palash25:
hi @bitethecode are you still working on this? if not then i would like to pick this up if that's ok with you

asubiotto:
Hi @palash25, are you still interested in working on this? I can reassign to you given that we haven't heard back.

asubiotto:
No problem! Thanks for giving it a shot. @palash25 assigned to you. Feel free to use https://github.com/polarsignals/frostdb/pull/225 as a guide.

palash25:
thanks for assigning i will look into it this weekend

palash25:
as i have already submitted 2 PRs, i would like to be unassigned so that someone else can try to convert the remaining go tests

asubiotto:
I think there are still some tests to be moved (notably the projection tests in `aggregate_test.go`). I believe #302 will likely address this, but reopening to track.

PR #257, opened 2022-11-27
logictest: convert operator unit tests to logic tests

palash25:
Issue:  https://github.com/polarsignals/frostdb/issues/226

Not ready for review yet, just opened a draft for visibility.


palash25:
@asubiotto i wasn't able to convert any more unit tests as i run into some issues while trying to convert them:
- `TestAggregationProjection` and `filter_test.go` all uses Projections which is currently not supported by the sqlparser, i see there is already an issue for that so I am assuming that these unit tests were already out of scope for this #226 till they get unblocked
- `TestAggregateInconsistentSchema` sorts the result before checking it but since the sqlparser does not support `*ast.OrderByClause` as of now I don't think i will be able to convert this test case correct me if I am wrong. I didn't find an equivalent of `Order By` in frostdb either so I am guessing that this can't be done even if sqlparser supported OrderBy?
- the sqlparser does not support comparison operators with more than one characters for e.g. `<=` or `>=` when i was trying to convert a filter test which used `GtEq(logicalplan.Literal(2))` the logictest panicked with this error i think `panic: unsupported binary boolean expression [recovered]`  Since frostdb supports these operators i think there should be an issue to add support for these in the sqlparser too


Another unrelated issue i ran across while playing around with logictest was this. I wrote the following logictest
```
createtable schema=default
----

insert cols=(labels.label1, labels.label2, labels.label3, labels.label4, stacktrace, timestamp, value)
value1  value2  null    null    stack1  1   1
value2  value2  value3  null    stack1  2   2
value3  value2  null    value4  stack1  3   3
----

exec
select labels, stacktrace, timestamp, value where labels.label4 = value4
----
value3  value2  null    value4  stack1  3   3
```

and it panicked with this 

```
panic: something terrible has happened, this should have errored previously during validation

goroutine 1033 [running]:
github.com/polarsignals/frostdb/query/physicalplan.BinaryScalarOperation({0x19ce800, 0xc00064e3c0}, {0x0?, 0x0}, 0x1)
        /home/palash/dev/frostdb/query/physicalplan/binaryscalarexpr.go:93 +0xa11
github.com/polarsignals/frostdb/query/physicalplan.BinaryScalarExpr.Eval({0xc000664540?, 0x4905b0?, {0x0?, 0x0?}}, {0x19cf498, 0xc000b0a570})
        /home/palash/dev/frostdb/query/physicalplan/binaryscalarexpr.go:54 +0xfd
github.com/polarsignals/frostdb/query/physicalplan.filter({0x19bef40, 0x27931c8}, {0x19bcd00, 0xc0009ba160}, {0x19cf498, 0xc000b0a570?})
        /home/palash/dev/frostdb/query/physicalplan/filter.go:252 +0x97
github.com/polarsignals/frostdb/query/physicalplan.(*PredicateFilter).Callback(0xc000324c40, {0x19c0e78, 0xc000324e00}, {0x19cf498, 0xc000b0a570})
        /home/palash/dev/frostdb/query/physicalplan/filter.go:235 +0xd0
github.com/polarsignals/frostdb.(*Table).Iterator.func1()
        /home/palash/dev/frostdb/table.go:725 +0x6b5
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /home/palash/go/pkg/mod/golang.org/x/sync@v0.0.0-20220601150217-0de741cfad7f/errgroup/errgroup.go:75 +0x87
created by golang.org/x/sync/errgroup.(*Group).Go
        /home/palash/go/pkg/mod/golang.org/x/sync@v0.0.0-20220601150217-0de741cfad7f/errgroup/errgroup.go:72 +0x12f
FAIL    github.com/polarsignals/frostdb/logictest       0.111s
FAIL
```

i am not sure if I did something wrong here but since the panic message said "it should have errored previously" i thought to let you folks know in case this is some missed edge case.

Let me know if this PR can be merged as is or would you like some more changes.

I would like to come back and convert the rest of the testcases once Projections are supported. 😄

logictest/testdata/aggregate/aggregate_nils
```diff
@@ -0,0 +1,26 @@
...
+insert cols=(labels.label1, labels.label2, stacktrace, timestamp, value)
+value1  null    stack1  1   1
+null    value2  stack1  2   2
+null    value2  stack1  3   3
+----
+
+exec
+select sum(value) as value_sum group by labels.label2
+----
+        1
```
asubiotto, line 13: Interesting that these blank values make the test pass. I assume they're considered nulls? Do you know why this is passing? In any case, I think it would be good to be explicit about the null value here.

logictest/testdata/aggregate/aggregate_nils
```diff
@@ -0,0 +1,26 @@
+createtable schema=default
```
asubiotto, line 1: re file name, I would call it `aggregate_nulls` instead of `nils` (even though the previous test mentioned nils) for consistency.

asubiotto (commented):
Thank you very much for your contribution! Fantastic to see this work.

Just have a couple nits/questions.

As you can see, our logictest framework is still very recent and there are a bunch of things that still need to be supported. Could you create an issue for the validation panic?

Once this is merged, feel free to work on adding more parsing support (e.g. the missing operators, or even projections). Intuitively it shouldn't be too much work and I'm happy to point you in the right direction if you run into any trouble.

thorfour:
So as not to step on toes, I should have a PR out today for adding Projections

logictest/testdata/aggregate/aggregate_nils
palash25: i am not sure, initially i tried to use `null` in the result but it ended up failing with this error 

```
logic_test.go:78: 
        testdata/aggregate/aggregate_nulls:10:
         select sum(value) as value_sum group by labels.label2
        expected:
        null    1
        value2  5
        
        found:
                1
        value2  5
```

so i switched to whitespace

asubiotto: Hmm, I guess it's probably an artifact of the aggregator that doesn't correctly set nulls in the output (and instead sets an empty string). This is probably something to open an issue about as well although it's not too high priority. We can probably merge this as is with at least a comment pointing to the issue.

asubiotto (approved):
Thanks for your contribution!

asubiotto:
> TestAggregateInconsistentSchema sorts the result before checking it but since the sqlparser does not support *ast.OrderByClause as of now I don't think i will be able to convert this test case correct me if I am wrong. I didn't find an equivalent of Order By in frostdb either so I am guessing that this can't be done even if sqlparser supported OrderBy?

btw just read this in more detail. The `exec` directive has an `unordered` argument which doesn't look at output ordering. Not sure if using something like that would be equivalent.

palash25:
thanks for mentioning i didn't know that, but then won't it turn out to identical to the aggregate_nulls logic test (except for the use of the directive ofcourse)

even both the unit tests `AggregateNils` and `AggregateInconsistent` schema were identical except for the sorting

palash25:
i just tried this 

```

exec unordered
select sum(value) as value_sum group by labels.label2
----
value2  5
             1
```

used unordered and flipped the order of the result but it fails

```
    logic_test.go:78: 
        testdata/aggregate/aggregate_nulls:30:
         select sum(value) as value_sum group by labels.label2
        expected:
        value2  5
                1
        
        found:
                1
        value2  5
```

it expects the same result as `aggregate_nulls`

asubiotto:
Hmm not quite sure why that would be failing. Can you run the test once with the `-rewrite` flag and see if it still fails? We'd need to dig a bit deeper

PR #258, opened 2022-11-29
Logictest projections

thorfour:
This adds Projection support to Logictests as well as adding support for additional schemas.


logictest/schemas/default.json
```diff
@@ -0,0 +1,60 @@
+{
+    "name": "default",
```
asubiotto, line 2: Love this idea. Is there a reason to add an additional json parsing layer? In my mind we'd do something like:
```
var schemas = map[string]*dynparquet.Schema{
    "default": dynparquet.NewSampleSchema(),
    "bools": dynparquet.SchemaFromDefinition(
        < your bools def here >
    )
}
```

I'm open to this json idea, just curious.

sqlparse/visitor.go
```diff
@@ -62,6 +62,8 @@ func (v *astVisitor) leaveImpl(n ast.Node) error {
 				v.exprStack[0],
 				v.exprStack[1:]...,
 			)
+		default:
```
asubiotto, line 65: Is it correct to have this as the `default` case? I guess previously we just returned all columns, so we do want to do a Project in all cases, is that right?

logictest/runner.go
```diff
@@ -105,10 +105,10 @@ func (r *Runner) handleCreateTable(ctx context.Context, c *datadriven.TestData)
 	var schema *dynparquet.Schema
 	for _, arg := range c.CmdArgs {
 		if arg.Key == "schema" {
-			if len(arg.Vals) != 1 && arg.Vals[0] != "default" {
+			if len(arg.Vals) != 1 {
 				return "", fmt.Errorf("createtable: unexpected schema values %v", arg.Vals)
 			}
-			schema = r.defaultSchema
+			schema = r.schemas[arg.Vals[0]]
```
asubiotto, line 111: nit: return an error if schema is not found

asubiotto (commented):
nice!

logictest/schemas/default.json
thorfour: I typically don't put configuration settings into code, but I don't have strong feelings against doing it that way

logictest/runner.go
thorfour: We do below where we check if schema is `nil`

sqlparse/visitor.go
thorfour: I think we'd want to extend this to support something like `select *` in the future, but I think for what we have now this is okay?

logictest/logic_test.go
```diff
@@ -2,6 +2,10 @@ package logictest
 
 import (
 	"context"
+	"encoding/json"
+	"io/fs"
+	"io/ioutil"
```
metalmatze, line 7: Deprecated. Use the newer APIs.

thorfour: thanks; old habits die hard.

brancz (approved):

PR #265, opened 2022-12-02
dynparquet: Use parquet-go's copy rows

brancz:
I think previously the only thing that was wrong was our usage ... we never called `.Close()`, which caused the goroutine opened by `.Rows()` to leak.

Roughly ~30% of the ingest path of Parca/Polar Signals Cloud is spent in this code, so hopefully this improves that.

### DIFF
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -652,12 +652,6 @@ func (s *Schema) NewBuffer(dynamicColumns map[string][]string) (*Buffer, error) 
 	}, nil
 }
 
-var rowBufPool = &sync.Pool{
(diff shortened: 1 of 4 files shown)


brancz:
Nevermind, looks like it's still leaking (leak tests detected it).

brancz:
🤦‍♂️ Managed to create exactly the same leak again by passing `.Rows()` in directly. Now not leaking anymore according to tests.

brancz:
Waiting for https://github.com/segmentio/parquet-go/pull/420 to merge first.

asubiotto (approved):

Issue #177, opened 2022-08-28
Support more aggregation functions


yeya24:
Right now frostdb only has limited support for aggregation functions. For observability and analytical purpose, more aggregation functions should be supported to cover more use cases. For me, I have the requirement to support aggregations like `count`, `average` and `quantile`. 

I opened this as an umbrella issue to track the support to other aggregation functions. For aggregation functions to support, feel free to add your use case. For a good start, probably we can take a look at https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators and https://duckdb.org/docs/sql/aggregates.

- [x] Support `count` aggregate function. PR: https://github.com/polarsignals/frostdb/pull/176
- [ ] Support `quantile` aggregate function
- [ ] Support `average` aggregate function

palash25:
hi @yeya24 are you working on either average or quantile functions?

yeya24:
@palash25 Please go ahead.

PR #266, opened 2022-12-03
aggregation: Add average function

palash25:
Ref: #177

Signed-off-by: Palash Nigam <npalash25@gmail.com>


palash25:
Not ready for review yet just opened this for better visibility. Average function is working fine but the logictests fail because of an arrow an error. Will try to figure it out later

PR #268, opened 2022-12-05
Added tracing to row group filtering of persisted blocks

thorfour:


asubiotto (approved):

store.go
```diff
@@ -127,10 +127,17 @@ func (t *Table) ProcessFile(ctx context.Context, blockDir string, lastBlockTimes
 		return err
 	}
 
+	return t.filterRowGroups(ctx, buf, filter, rowGroups)
+}
+
+func (t *Table) filterRowGroups(ctx context.Context, buf *dynparquet.SerializedBuffer, filter TrueNegativeFilter, rowGroups chan<- dynparquet.DynamicRowGroup) error {
+	_, span := t.tracer.Start(ctx, "Table/IterateBucketBlocks/Iter/ProcessFile/filterRowGroups")
```
metalmatze, line 134: Don't have hard feelings about this, but I think I'd go with `Table/filterRowGroups`  the call stack is after all visible on the tracing UI too.

thorfour: Oh good! I didn't know what the standard was for naming traces, I do like shorter myself

PR #271, opened 2022-12-06
physicalplan: add OrderedAggregate

asubiotto:
This commit adds an aggregation operator that performs aggregates when the
grouping column is at least partially ordered. This allows us to assume that
all values for a given group are found contiguously, removing the need to hash
the grouping column and accumulate values until all values have been read.

Forcing the usage of this OrderedAggregate for QueryMerge gives us a ~15%
performance improvement:
```
Query/Merge-8    14.6ms ± 1%    12.5ms ± 1%  -14.43%  (p=0.000 n=9+9)

Query/Merge-8    56.3MB ± 0%    51.1MB ± 0%   -9.11%  (p=0.000 n=10+10)

Query/Merge-8     88.0k ± 0%     33.4k ± 0%  -61.98%  (p=0.000 n=10+10)
```

This commit introduces the OrderedAggregate as is and only supports a limited
use case (single-group final aggregations). There are still a couple
of steps to implement to use this in general execution flows:
- Support multiple groups.
- Support ordered merging of internal results (i.e. on partially ordered
streams).
- Support ordered merging of concurrent ordered aggregators.
- Support planning.

> **Note**
> My idea is to PR this ordered aggregate and do the follow up work in future PRs as I think it will be easier to review this way. Please let me know if you prefer I simply work on all the necessary things to plan the ordered aggregate automatically. Tests are also elided on purpose since we will be able to use logic tests once this is all hooked up.

### DIFF
--- a/query/physicalplan/ordered_aggregate.go
+++ b/query/physicalplan/ordered_aggregate.go
@@ -1,0 +1,394 @@
+package physicalplan
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/apache/arrow/go/v8/arrow"
+	"github.com/apache/arrow/go/v8/arrow/array"
+	"github.com/apache/arrow/go/v8/arrow/memory"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/polarsignals/frostdb/pqarrow/builder"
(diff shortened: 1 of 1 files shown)


brancz (approved):
Very very nice! lgtm 💯

thorfour (approved):

PR #272, opened 2022-12-07
deps: update parquet-go to db0f5ce

asubiotto:
This bump includes some changes (mentioned in commits) that improve our performance:
```
Query/Types-8    7.08ms ± 1%    4.55ms ± 2%  -35.70%  (p=0.000 n=9+9)
Query/Merge-8    14.4ms ± 0%    11.8ms ± 1%  -18.32%  (p=0.000 n=9+9)
Query/Range-8    5.80ms ± 1%    4.82ms ± 4%  -16.93%  (p=0.000 n=10+9)

Query/Types-8    52.5MB ± 0%    35.3MB ± 0%  -32.73%  (p=0.000 n=9+9)
Query/Merge-8    91.7MB ± 0%    53.6MB ± 1%  -41.59%  (p=0.000 n=10+10)
Query/Range-8    24.5MB ± 0%    13.4MB ± 0%  -45.48%  (p=0.000 n=9+10)

Query/Types-8     45.4k ± 0%     25.6k ± 0%  -43.65%  (p=0.000 n=10+8)
Query/Merge-8      169k ± 0%      134k ± 0%  -20.64%  (p=0.000 n=10+10)
Query/Range-8     11.9k ± 0%      7.6k ± 0%  -36.40%  (p=0.000 n=9+10)
```

> **Warning**
> The bump introduced some breaking changes which required a significant change to our comparison function as well as object storage reads. Please review these two code changes carefully: [comparison link](https://github.com/polarsignals/frostdb/compare/alfonso-parquet?expand=1#diff-71873ced73141be0a44105e7f4ef261b8d8e1b30f2967499174d3548a403d372R50), [object storage read link](https://github.com/polarsignals/frostdb/compare/alfonso-parquet?expand=1#diff-ed648181f98484bd12541509e0ae7b5ad1d1de7674ab1721814b47ddd5c95de4R87)


brancz (commented):
Nice! lgtm!

asubiotto:
Fixed the comparison test failures (minus a data race I'm fixing now). The issue was that if we had two rows: `["cpu", "node"="test3"]` `["cpu", "namespace"="default"]` we were not passing in the parquet rows to the comparator with the merged dynamic column values: (i.e. `["cpu", NULL, "node"="test3"]`, `["cpu", "namespace"="default", NULL]`.
The one issue I've spent some time thinking about is that the parquet rows that are constructed as a result have column indexes that might cause problems (i.e. [-1, -2, -2] for the first and second row). I'm specifically wary of this comparison code: https://github.com/segmentio/parquet-go/blob/db0f5ce13cf195dc76826140671b1edf10800d9c/compare.go#L284

The code only iterates until the desired column index is found. So far I haven't been able to come up with a scenario where this will be an issue since if there are repeated column indexes (e.g. -2 above), it necessarily means that it is a dynamic column that is not equal among rows, thus terminating the comparison. I added a comment about this.

asubiotto:
Actually scratch that, we can just use `Level` to set the column index correctly and not worry about this in the future.

PR #273, opened 2022-12-07
Schema v2 nested

thorfour:
This is the first step in supported nested schemas in FrostDB. I wanted to make the changes incrementally so that they'll be easier to review.

This adds a new schema definition in v1alphav2 that allows for nested schema definitions. It updates the `dynparquet/schema.go` to support parsing the definition into a valid Schema. The unit test is able to create a schema, write a row into a buffer and store it into the table. 

What's next to be implemented is being able to query the data back out, right now our arrow converter does not support group types and therefor cannot convert the parquet structs into arrow.


proto/frostdb/schema/v1alpha2/schema.proto
```diff
@@ -0,0 +1,177 @@
...
+    bool nullable = 4;
+
+    // Indicates whether the parquet column is repeated.
+    bool repeated = 5;
+}
+
+// SortingColumn definition.
+message SortingColumn {
+    // Name of the column to sort by.
+    string name = 1;
```
brancz, line 160: this can't just be a name anymore, it needs to be a "path" to the leaf, this also has consequences for our Ascending/Descending functions (eg. https://github.com/segmentio/parquet-go/blob/95f6ec7f32ee32a5e9016dcbcb0845e192e6aa41/row_group.go#L95)

brancz (commented):
Just a small semantic issue with sorting columns, the rest looks good to me!

proto/frostdb/schema/v1alpha2/schema.proto
thorfour: Yea, I think you're right

PR #276, opened 2022-12-08
Allow aggregations to be grouped by durations

metalmatze:
We want to be able to group aggregations into interval/duration buckets.

A simple example in English would be: Sum up all messages in a chat by the minutes in an hour.
More technically correct is the logical test in `logictest/testdata/aggregate/window`. 

For this, to work we exploit the fact that we can group timestamps into intervals by dividing these through the duration.

| timestamp 	| duration 	| math       	| hashed bucket 	|
|-----------	|----------	|------------	|---------------	|
| 0         	| 2        	| floor(0/2) 	| 0             	|
| 1         	| 2        	| floor(1/2) 	| 0             	|
| 2         	| 2        	| floor(2/2) 	| 1             	|
| 3         	| 2        	| floor(3/2) 	| 1             	|
| 4         	| 2        	| floor(4/2) 	| 2             	|
| 5         	| 2        	| floor(5/2) 	| 2             	|
| 6         	| 2        	| floor(6/2) 	| 3             	|

_and so on._

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -126,7 +126,7 @@ type HashAggregate struct {
 
 	// Buffers that are reused across callback calls.
 	groupByFields      []arrow.Field
-	groupByFieldHashes []uint64
+	groupByFieldHashes []hashCombiner
 	groupByArrays      []arrow.Array
 }
 
@@ -155,7 +155,7 @@ func NewHashAggregate(
 		finalStage:            finalStage,
 
 		groupByFields:      make([]arrow.Field, 0, 10),
-		groupByFieldHashes: make([]uint64, 0, 10),
+		groupByFieldHashes: make([]hashCombiner, 0, 10),
 		groupByArrays:      make([]arrow.Array, 0, 10),
 	}
 }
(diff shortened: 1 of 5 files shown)


query/physicalplan/aggregate.go
```diff
@@ -185,6 +185,37 @@ func hashCombine(lhs, rhs uint64) uint64 {
 	return lhs ^ (rhs + 0x9e3779b9 + (lhs << 6) + (lhs >> 2))
 }
 
+// hashCombiner combines a given hash with another hash that is passed.
+type hashCombiner interface {
+	hashCombine(rhs uint64) uint64
```
brancz, line 190: hash combining is something that we do _a lot_ in the hot path, could we run a quick benchmark what this added virtualization costs us? (we can probably just add an explicit devirtualization if this has a perf impact)

thorfour (approved):
Nice!

metalmatze:
Doesn't seem to bad. :relaxed: 
```
Aggregation/sum-24      1.83ms ± 1%    1.85ms ± 1%  +0.85%  (p=0.011 n=9+9)
Aggregation/count-24    1.82ms ± 2%    1.84ms ± 1%  +1.50%  (p=0.002 n=10+10)
Aggregation/max-24      1.82ms ± 1%    1.85ms ± 1%  +1.94%  (p=0.000 n=10+10)

Aggregation/sum-24      1.63MB ± 0%    1.63MB ± 0%    ~     (p=0.274 n=10+8)
Aggregation/count-24    1.63MB ± 0%    1.63MB ± 0%    ~     (p=0.905 n=9+10)
Aggregation/max-24      1.63MB ± 0%    1.63MB ± 0%    ~     (p=0.190 n=10+10)

Aggregation/sum-24       1.43k ± 0%     1.44k ± 0%  +0.14%  (p=0.000 n=8+8)
Aggregation/count-24     1.43k ± 0%     1.44k ± 0%  +0.14%  (p=0.000 n=9+9)
Aggregation/max-24       1.43k ± 0%     1.44k ± 0%  +0.14%  (p=0.000 n=8+10)
```

brancz:
yep, looks good, happy with that

PR #278, opened 2022-12-12
dynparquet: fix out of bounds comparison panic

asubiotto:
This was caused by using the general parquet schema of all columns and calling its comparison function with rows of just the sorting columns, causing an index OOB panic in some cases. This commit creates a parquet schema of just the sorting columns for the comparison.

### DIFF
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -611,6 +611,35 @@ func (s Schema) parquetSchema(
 	}
 }
 
+// parquetSortingSchema returns the parquet schema of just the sorting columns
+// with the concrete dynamic column names given in the argument.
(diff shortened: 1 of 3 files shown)


dynparquet/schema.go
```diff
@@ -611,6 +611,35 @@ func (s Schema) parquetSchema(
 	}
 }
 
+// parquetSortingSchema returns the parquet schema of just the sorting columns
+// with the concrete dynamic column names given in the argument.
+func (s Schema) parquetSortingSchema(
```
asubiotto, line 616: cc @thorfour since this schema is just used for sorting I assume there's nothing to do extra for handling nested schemas?

brancz (approved):

PR #282, opened 2022-12-12
Accept multiple aggregations within a single query

metalmatze:
FrostDB should be able to accept queries like:
```
select sum(value), count(value), max(value) group by labels.label2
select sum(value) as value_sum, count(value) as value_count group by second(3)
```

For this, to work FrostDB needs to accept multiple aggregations within a single query. 
FrostDBs Aggregate interface has to be updated:
```diff
-Aggregate(aggExpr logicalplan.Expr, groupExprs ...logicalplan.Expr) Builder
+Aggregate(aggExpr, groupExprs []logicalplan.Expr) Builder
```

To make our lives easier to figure out all the things that belong to a single column that's being aggregated, I've introduced `AggregateColumn`. 

Benchmarks show no significant impact:
```
Aggregation/sum-24      1.84ms ± 3%    1.87ms ± 1%    ~     (p=0.065 n=10+9)
Aggregation/count-24    1.85ms ± 1%    1.85ms ± 1%    ~     (p=0.489 n=9+9)
Aggregation/max-24      1.83ms ± 2%    1.86ms ± 1%  +1.25%  (p=0.004 n=10+9)

Aggregation/sum-24      1.45MB ± 0%    1.46MB ± 0%  +0.16%  (p=0.000 n=10+10)
Aggregation/count-24    1.45MB ± 0%    1.46MB ± 0%  +0.17%  (p=0.000 n=10+10)
Aggregation/max-24      1.45MB ± 0%    1.46MB ± 0%  +0.17%  (p=0.000 n=10+10)

Aggregation/sum-24       1.43k ± 0%     1.50k ± 0%  +5.23%  (p=0.000 n=10+9)
Aggregation/count-24     1.43k ± 0%     1.50k ± 0%  +5.25%  (p=0.000 n=10+10)
Aggregation/max-24       1.43k ± 0%     1.50k ± 0%  +5.25%  (p=0.000 n=9+8)
```

### DIFF
--- a/query/logicalplan/validate.go
+++ b/query/logicalplan/validate.go
@@ -146,7 +146,7 @@ func ValidateSingleFieldSet(plan *LogicalPlan) *PlanValidationError {
 // ValidateAggregation validates the logical plan's aggregation step.
 func ValidateAggregation(plan *LogicalPlan) *PlanValidationError {
 	// check that the expression is not nil
-	if plan.Aggregation.AggExpr == nil {
+	if plan.Aggregation.AggExprs == nil {
 		return &PlanValidationError{
 			plan:    plan,
 			message: "invalid aggregation: expression cannot be nil",
@@ -167,17 +167,23 @@ func ValidateAggregation(plan *LogicalPlan) *PlanValidationError {
 }
 
 func ValidateAggregationExpr(plan *LogicalPlan) *ExprValidationError {
+	// TODO: Check that aggregation columns don't have the same result name
+
 	// check that the aggregation expression has the required structure
 	colFinder := newTypeFinder((*Column)(nil))
-	plan.Aggregation.AggExpr.Accept(&colFinder)
+	for _, expr := range plan.Aggregation.AggExprs {
+		expr.Accept(&colFinder)
+	}
 
 	aggFuncFinder := newTypeFinder((*AggregationFunction)(nil))
-	plan.Aggregation.AggExpr.Accept(&aggFuncFinder)
+	for _, expr := range plan.Aggregation.AggExprs {
+		expr.Accept(&aggFuncFinder)
+	}
 
 	if colFinder.result == nil || aggFuncFinder.result == nil {
 		return &ExprValidationError{
 			message: "aggregation expression is invalid. must contain AggregationFunction and Column",
-			expr:    plan.Aggregation.AggExpr,
+			expr:    plan.Aggregation.AggExprs[0], // TODO
 		}
 	}
 
@@ -192,7 +198,7 @@ func ValidateAggregationExpr(plan *LogicalPlan) *ExprValidationError {
 	if !found {
 		return &ExprValidationError{
 			message: fmt.Sprintf("column not found: %s", colExpr.ColumnName),
-			expr:    plan.Aggregation.AggExpr,
+			expr:    plan.Aggregation.AggExprs[0], // TODO
 		}
 	}
 
@@ -204,12 +210,12 @@ func ValidateAggregationExpr(plan *LogicalPlan) *ExprValidationError {
 		case AggFuncSum:
 			return &ExprValidationError{
 				message: "cannot sum text column",
-				expr:    plan.Aggregation.AggExpr,
+				expr:    plan.Aggregation.AggExprs[0], // TODO
 			}
 		case AggFuncMax:
 			return &ExprValidationError{
 				message: "cannot max text column",
-				expr:    plan.Aggregation.AggExpr,
+				expr:    plan.Aggregation.AggExprs[0], // TODO
 			}
 		}
 	}
--- a/query/logicalplan/validate.go
+++ b/query/logicalplan/validate.go
@@ -146,7 +146,7 @@ func ValidateSingleFieldSet(plan *LogicalPlan) *PlanValidationError {
 // ValidateAggregation validates the logical plan's aggregation step.
 func ValidateAggregation(plan *LogicalPlan) *PlanValidationError {
 	// check that the expression is not nil
-	if plan.Aggregation.AggExprs == nil {
+	if plan.Aggregation.AggExprs == nil || len(plan.Aggregation.AggExprs) == 0 {
 		return &PlanValidationError{
 			plan:    plan,
 			message: "invalid aggregation: expression cannot be nil",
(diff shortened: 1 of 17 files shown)


metalmatze:
There are some TODOs left in the validation we should fix. I think before merging this PR, otherwise, it won't happen :grimacing:

query/physicalplan/aggregate.go
```diff
@@ -27,51 +26,56 @@ func Aggregate(
...
+	aggregateColumns := make([]AggregateColumn, 0, len(agg.AggExprs))
+
+	for _, expr := range agg.AggExprs {
+		var (
+			aggColumn      AggregateColumn
+			aggFunc        logicalplan.AggFunc
+			aggFuncFound   bool
+			aggColumnFound bool
+		)
+		expr.Accept(PreExprVisitorFunc(func(expr logicalplan.Expr) bool {
```
asubiotto, line 38: Maybe it's time to extract this aggColumn finding/creation logic to a separate helper function while you're here? up to you

query/physicalplan/aggregate.go
```diff
@@ -103,21 +107,26 @@ func chooseAggregationFunction(
 	}
 }
 
+// AggregateColumn groups together some lower level primitives to aggregate.
+type AggregateColumn struct {
+	expr       logicalplan.Expr
+	resultName string
+	function   AggregationFunction
+	arrays     []builder.ColumnBuilder // TODO: These can actually live outside this struct and be shared. Only at the very end will they be read by each column and then aggregated separately.
```
asubiotto, line 115: Since `builder.ColumnBuilder` is an interface (i.e. pointer), I think these can actually stay here but each `AggregateColumn.arrays[i]` can point to the same underlying builder.

query/physicalplan/aggregate.go
```diff
@@ -103,21 +107,26 @@ func chooseAggregationFunction(
...
+
 type AggregationFunction interface {
 	Aggregate(pool memory.Allocator, arrs []arrow.Array) (arrow.Array, error)
 }
 
 type HashAggregate struct {
 	pool                  memory.Allocator
 	tracer                trace.Tracer
-	resultColumnName      string
+	columns               []AggregateColumn
```
asubiotto, line 125: nit: I think it'd be nice to make the naming a bit less generic. Are these the values we're aggregating? Maybe `columnsToAggregate`

query/physicalplan/aggregate.go
```diff
@@ -175,7 +179,8 @@ func (a *HashAggregate) Draw() *Diagram {
 		groupings = append(groupings, grouping.Name())
 	}
 
-	details := fmt.Sprintf("HashAggregate (%s by %s)", a.columnToAggregate.Name(), strings.Join(groupings, ","))
+	// details := fmt.Sprintf("HashAggregate (%s by %s)", a.columnsToAggregate[0].Name(), strings.Join(groupings, ","))
```
asubiotto, line 182: leftover?

query/physicalplan/aggregate.go
```diff
@@ -315,14 +320,25 @@ func (a *HashAggregate) Callback(ctx context.Context, r arrow.Record) error {
...
+				if col.resultName == field.Name {
+					columnToAggregate[j] = r.Column(i)
+					aggregateFieldFound[j] = true
+				}
+			} else {
+				if col.expr.MatchColumn(field.Name) {
+					columnToAggregate[j] = r.Column(i)
+					aggregateFieldFound[j] = true
+				}
+			}
```
asubiotto, line 334: What was the need for the addition of these `if` statements? If they're necessary I would add a comment and here's a rewrite suggestion:
```suggestion
				if col.expr.MatchColumn(field.Name) || col.resultName == field.Name {
					columnToAggregate[j] = r.Column(i)
					aggregateFieldFound[j] = true
				}
```

query/physicalplan/aggregate.go
```diff
@@ -315,14 +320,25 @@ func (a *HashAggregate) Callback(ctx context.Context, r arrow.Record) error {
...
 		}
 	}
 
-	if !aggregateFieldFound {
-		return errors.New("aggregate field not found, aggregations are not possible without it")
+	for _, found := range aggregateFieldFound {
```
asubiotto, line 338: nit: another option is to use a counter instead of `n` booleans. Avoids allocations as well

query/physicalplan/aggregate.go
```diff
@@ -390,7 +410,7 @@ func (a *HashAggregate) Finish(ctx context.Context) error {
 	defer span.End()
 
 	numCols := len(a.groupByCols) + 1
-	numRows := len(a.arraysToAggregate)
+	numRows := len(a.columns[0].arrays)
```
asubiotto, line 413: Will the length of `arrays` be equal across all `columns[i]`? This code is starting to get pretty complex/hard to understand but maybe it's just me.

sqlparse/visitor.go
```diff
@@ -59,11 +59,20 @@ func (v *astVisitor) leaveImpl(n ast.Node) error {
...
-			)
+			var agg []logicalplan.Expr
+			var groups []logicalplan.Expr
+
+			for _, expr := range v.exprStack {
+				switch expr.(type) {
+				case *logicalplan.AliasExpr:
+					agg = append(agg, expr)
+				case *logicalplan.AggregationFunction:
+					agg = append(agg, expr)
```
asubiotto, line 70: nit: 
```suggestion
				case *logicalplan.AliasExpr, *logicalplan.AggregationFunction:
					agg = append(agg, expr)
```

asubiotto (commented):
Nice work!

query/logicalplan/validate.go
```diff
@@ -167,17 +167,23 @@ func ValidateAggregation(plan *LogicalPlan) *PlanValidationError {
...
-	plan.Aggregation.AggExpr.Accept(&aggFuncFinder)
+	for _, expr := range plan.Aggregation.AggExprs {
+		expr.Accept(&aggFuncFinder)
+	}
 
 	if colFinder.result == nil || aggFuncFinder.result == nil {
 		return &ExprValidationError{
 			message: "aggregation expression is invalid. must contain AggregationFunction and Column",
-			expr:    plan.Aggregation.AggExpr,
+			expr:    plan.Aggregation.AggExprs[0], // TODO
```
thorfour, line 186: Do we intend to leave these TODO in here, if so can we add a comment as to why?

query/physicalplan/aggregate.go
metalmatze: Yeah, ask me about implementing these things. I guess we should figure out a way to refactor things into smaller pieces. An attempt was the introduction of `AggregateColumn`.

query/physicalplan/aggregate.go
```diff
@@ -412,44 +432,46 @@ func (a *HashAggregate) Finish(ctx context.Context) error {
 		groupByArrays = append(groupByArrays, arr)
 	}
 
-	arrs := make([]arrow.Array, 0, numRows)
-	for _, arr := range a.arraysToAggregate {
-		arrs = append(arrs, arr.NewArray())
-	}
+	aggregateColumns := append(groupByArrays)
```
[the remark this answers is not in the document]
metalmatze: Good point, I guess we can simply append the other fields to the existing `groupByArrays`.

query/logicalplan/validate.go
metalmatze: This is one left over from my side. I want to fix. 
Additionally, I'll address the comments made here too.

query/physicalplan/aggregate.go
metalmatze: In what way? I think this is actually fine the way it is, but I'm up for proposals. It's definitely not the biggest offender in `aggregate.go` :grimacing:

query/physicalplan/aggregate.go
metalmatze: I'm not 100% sure what you're referring to with "can actually stay here". Do you mean they can stay outside of the AggregateColumn in the `HashAggregate` struct?

query/physicalplan/aggregate.go
metalmatze: Yes, agreed. The naming isn't perfect. It's not even about columns really. 
Given a query like `select sum(value), count(value), max(value) group by labels.label2` it's all the same `value` column, but different aggregations. So I'll probably call it `[]Aggregation`.

query/physicalplan/aggregate.go
metalmatze: > Will the length of `arrays` be equal across all `columns[i]`?

I double-checked this. This way of getting the num of Rows is fine.
For each different hash, we group by we append a new array for each aggregation to this array. Each of these arrays is going to be aggregated into a single value based on its function.
They `len(arrays)` will be consistent across aggregations.

Here is the code where this is taken care of:
https://github.com/polarsignals/frostdb/blob/4d6fe5bb56bf0d4527b25691c509ee977202601a/query/physicalplan/aggregate.go#L364-L367

metalmatze: To describe this with a relationship. Each aggregation will have the exact same amount of arrays, because they each correspond to a hash in the `hashToAggregate` map.

metalmatze: I'll add this as a comment though, so it's less confusing in the future! :+1:

metalmatze: Actually, this should be way more descriptive.
```
// Each hash that's aggregated by will become one row in the final result.
numRows := len(a.hashToAggregate)
```

query/physicalplan/aggregate.go
asubiotto: No, that `arrays` can stay as an `Aggregation` field, just that you can always set it up so that two aggregation struct array field indices point to the same builder.

asubiotto (approved):

PR #286, opened 2022-12-14
physicalplan: support muliple group by columns in the OrderedAggregate

asubiotto:
This commit supports grouping by multiple columns in ordered aggregations. It implements this using a min-heap of group ranges so that the first group by value to differ in any column is considered the start of a new group.

This commit also adds a TestOrderedAggregate unit test that creates and pushes arrow records to the ordered aggregate.


metalmatze (approved):
Nice one! :rocket:

PR #288, opened 2022-12-15
physicalplan: skip TestOrderedAggregate due to flake

asubiotto:
Informs #287

PR #289, opened 2022-12-15
wal: fix wal.proto backwards incompatibility

### LINKED ISSUE #281: Wal replay not correctly decoding schema
When trying to run some benchmarks on a WAL, the recent schema v2 changes (https://github.com/polarsignals/frostdb/pull/273) cause this error to occur:
```
go test -benchmem -run=^$ -bench BenchmarkQuery github.com/polarsignals/frostdb -test.timeout 24h -count 10 -v
goos: darwin
goarch: arm64
BenchmarkQuery
    bench_test.go:119: replaying WAL
    bench_test.go:120:
        	Error Trace:	bench_test.go:120
        	Error:      	Received unexpected error:
        	            	second WAL replay: call replay handler: insert buffer into block: failed to get writer: expected 0 dynamic column names, got 3
        	Test:       	BenchmarkQuery
```

I tracked this down to the fact that neither of the cases in this switch case is being hit: https://github.com/polarsignals/frostdb/blob/fb9896e4077d42a2a36d35acd4d292f5aa261b26/db.go#L432-L443
and there is no `default` catchall that errors out. This seems like in an empty schema being created, thus no dynamic columns.

cc @thorfour

asubiotto:
The new schema changes overrode the proto field, which caused the old schema message to be incorrectly unmarshalled. This commit adds the field back in (as deprecated) so that we can handle WALs generated before the change.

Closes #281


asubiotto:
Test failure is a known flake. Skipped in: https://github.com/polarsignals/frostdb/pull/288

brancz (approved):

metalmatze (approved):

asubiotto:
Actually, using `oneof` instead of `types.Any` works. Added another commit so you can compare. What do we think? I'll wait for Thor to take a look as well and we can make a decision.

asubiotto:
Awesome, TFTRs! Will clean up and merge

PR #292, opened 2022-12-15
query/physicalplan: Fix aggregations for columns that are grouped by

metalmatze:
In case of a query like `select count(timestamp) as timestamp_count group by second(3)` we have to add the `timestamp` column arrays once to the `columnToAggregate` with the way it was rewritten the column was added twice. 
In the end `columnToAggregate` had the timestamp column twice and return `aggregate field not found, aggregations are not possible without it` as error.

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -326,9 +326,19 @@ func (a *HashAggregate) Callback(ctx context.Context, r arrow.Record) error {
 		}
 
 		for j, col := range a.aggregations {
-			if col.expr.MatchColumn(field.Name) || col.resultName == field.Name {
-				columnToAggregate[j] = r.Column(i)
-				aggregateFieldsFound++
+			// If we're aggregating at the final stage we have previously
+			// renamed the pre-aggregated columns to their result names.
+			if a.finalStage {
+				if col.resultName == field.Name {
+					columnToAggregate[j] = r.Column(i)
+					aggregateFieldsFound++
+				}
+			} else {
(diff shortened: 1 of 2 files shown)


query/physicalplan/aggregate.go
```diff
@@ -326,9 +326,19 @@ func (a *HashAggregate) Callback(ctx context.Context, r arrow.Record) error {
...
+				if col.resultName == field.Name {
+					columnToAggregate[j] = r.Column(i)
+					aggregateFieldsFound++
+				}
+			} else {
+				// If we're aggregating the raw data we need to find the columns by their actual names for now.
+				if col.expr.MatchColumn(field.Name) {
+					columnToAggregate[j] = r.Column(i)
+					aggregateFieldsFound++
+				}
```
asubiotto, line 341: Suggestion to not repeat code:
```suggestion
			// If we're aggregating at the final stage we have previously
			// renamed the pre-aggregated columns to their result names.
			// Otherwise, we need to find the columns by their actual names for now.
			if (a.finalStage && col.resultName == field.Name) || col.expr.MatchColumn(field.Name) {
					columnToAggregate[j] = r.Column(i)
					aggregateFieldsFound++
				}
			}
```

asubiotto (approved):

query/physicalplan/aggregate.go
metalmatze: Tried this again. It doesn't work. The `col.expr.MatchColumn(field.Name)`  must only be called if `!a.finalStage`. With the short method it still evaluates `true` for the underlying column and then matches. Causing columns to possibly be selected twice.

PR #293, opened 2022-12-19
arrowutils: introduce arrow utility package and MergeRecords

asubiotto:
This PR introduces a way to merge two sorted arrow records according to given sorting columns.

I also have a change on the back burner to drop duplicates/call a callback on equal values, but it's not working given I assumed some incorrect things about `container/heap`. I think the best thing will be to merge this without special duplicate behavior and then see how we use it in the ordered aggregator.


brancz (approved):
lgtm

PR #297, opened 2022-12-22
Add support for average aggregations

### LINKED ISSUE #177: Support more aggregation functions

metalmatze:
Thanks to @asubiotto we had the idea to make the average aggregation a projection. 

The way it works now is that the logical plan creates an average aggregation but then optimizes those into the two `sum` and `count` aggregations plus an `avg` projection. The logical plan makes sure to wrap both aggregations with the average projection, such that anything reading the column afterward will only see the final result. 

TODO:
- [x] Improve `AverageAggregationPushDown.Optimize()` to support with and without alias columns

Closes #177 
Superseeds #266 & #190

### DIFF
--- a/logictest/testdata/aggregate/aggregate
+++ b/logictest/testdata/aggregate/aggregate
@@ -29,6 +29,17 @@ select count(value) as value_count group by labels.label2
 value2  6
 
 exec
+select avg(value) as value_avg group by labels.label2
+----
+value2  3
+
+exec
+select avg(value) as value_avg group by labels.label4
+----
+        3
+value4  4
+
+exec
 select sum(value), count(value) group by stacktrace
 ----
 stack1  21      6
--- a/query/logicalplan/optimize.go
+++ b/query/logicalplan/optimize.go
@@ -19,6 +19,10 @@ var DefaultOptimizers = []Optimizer{
 type AverageAggregationPushDown struct{}
 
 func (p *AverageAggregationPushDown) Optimize(plan *LogicalPlan) *LogicalPlan {
+	if plan.Aggregation == nil {
+		return plan
+	}
+
 	for i, aggExpr := range plan.Aggregation.AggExprs {
 		if aliasExpr, ok := aggExpr.(*AliasExpr); ok {
 			if aggFunc, ok := aliasExpr.Expr.(*AggregationFunction); ok {
--- a/query/logicalplan/optimize.go
+++ b/query/logicalplan/optimize.go
@@ -24,26 +24,52 @@ func (p *AverageAggregationPushDown) Optimize(plan *LogicalPlan) *LogicalPlan {
 	}
 
 	for i, aggExpr := range plan.Aggregation.AggExprs {
-		if aliasExpr, ok := aggExpr.(*AliasExpr); ok {
-			if aggFunc, ok := aliasExpr.Expr.(*AggregationFunction); ok {
-				if aggFunc.Func == AggFuncAvg {
-					// Delete this average aggregation from the logicalplan.
-					plan.Aggregation.AggExprs = slices.Delete(plan.Aggregation.AggExprs, i, i+1)
-					// Add sum and count aggregation for the column to the logicalplan.
-					plan.Aggregation.AggExprs = append(plan.Aggregation.AggExprs,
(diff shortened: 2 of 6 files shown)


brancz:
Did you still want to take care of that TODO before merging?

metalmatze:
Yep, currently on it.

brancz (approved):
Nice!

logictest/testdata/aggregate/aggregate
```diff
@@ -28,6 +28,17 @@ select count(value) as value_count group by labels.label2
 ----
 value2  6
 
+exec
+select avg(value) as value_avg group by labels.label2
+----
+value2  3
+
+exec
+select avg(value) as value_avg group by labels.label4
```
asubiotto, line 37: I think we should also run this function in tests with multiple row groups (i.e. insert statements), I think we already had tests like this for the other functions too?

query/physicalplan/project.go
```diff
@@ -184,6 +184,8 @@ func projectionFromExpr(expr logicalplan.Expr) (columnProjection, error) {
 		return binaryExprProjection{
 			boolExpr: boolExpr,
 		}, nil
+	case *logicalplan.AverageExpr:
+		return &averageProjection{expr: e}, nil
```
asubiotto, line 188: I'm not sure this is the right way to define this projection. In my mind, this projection is just a general division projection which intuitively fits into the switch case above: i.e. this is just a binary expression where the operator is the division operator. Not sure if our code is structured for this right now, but long term I think that's what we should do. This feels too specific to aggregations.

query/logicalplan/optimize.go
```diff
@@ -1,16 +1,81 @@
...
+	&AverageAggregationPushDown{},
 	&PhysicalProjectionPushDown{},
 	&FilterPushDown{},
 	&DistinctPushDown{},
 	&ProjectionPushDown{},
 }
 
+type AverageAggregationPushDown struct{}
+
+func (p *AverageAggregationPushDown) Optimize(plan *LogicalPlan) *LogicalPlan {
```
asubiotto, line 21: This is also something I'm not sure belongs in our optimizer code. Intuitively, optimizations are applied conditionally based on some heuristics, while this I think we always want to do so it seems like it belongs in our planning code.

asubiotto (commented):
Sorry I'm reviewing this so late and once merged. Just wanted to add a couple of comments. Feel free to ignore.

logictest/testdata/aggregate/aggregate
metalmatze: Above we have two inserts, so multiple row groups, that should be what you are looking for, isn't it?

asubiotto: I see, in that case how is this passing without an `exec unordered` argument? The output ordering shouldn't be deterministic I think.

query/physicalplan/project.go
metalmatze: Here's a changed implementation for making it less specific to the average case: https://github.com/polarsignals/frostdb/pull/302

query/logicalplan/optimize.go
metalmatze: Agreed, this is where I also started back in December iirc. However, the way the structs need to be constructed would have resulted in the logical and physical plan packages having cyclic dependencies. In case we want to put it elsewhere we should rethink the packages in some cases (e.g. create another package both can rely on).

asubiotto: Could you elaborate? I'm not sure I understand the cyclic dependency.

PR #298, opened 2022-12-22
Arrow record ingestion support

thorfour:
This adds support to ingest Arrow records directly into FrostDB. It **does not** deprecate the previous method of ingesting parquet buffers directly. That may be a future PR to deprecate that write path. 

It stores arrow records inside the `Part` object, where a part may now hold either a parquet buffer or an arrow record. 

Things that are **not** implemented in this PR that should be implemented in follow-on PRs:
- WAL support (writing the wal with arrow records is not supported at this time)
- Nested schema support 
- Dictionary arrays. (We want this to reduce the memory overhead of storing arrow records)

### DIFF
--- a/bufutils/dyncols.go
+++ b/bufutils/dyncols.go
@@ -2,6 +2,7 @@ package bufutils
 
 import "sort"
 
+// Dedupe deduplicates the slices of values for each key in the map.
 func Dedupe(s map[string][]string) map[string][]string {
 	final := map[string][]string{}
 	set := map[string]map[string]struct{}{}
--- a/bufutils/dyncols.go
+++ b/bufutils/dyncols.go
@@ -1,0 +1,24 @@
+package bufutils
+
+import "sort"
+
+func Dedupe(s map[string][]string) map[string][]string {
+	final := map[string][]string{}
+	set := map[string]map[string]struct{}{}
+	for k, v := range s {
+		if set[k] == nil {
+			set[k] = map[string]struct{}{}
+		}
+		for _, i := range v {
+			if _, ok := set[k][i]; !ok {
+				set[k][i] = struct{}{}
+				final[k] = append(final[k], i)
+			}
+		}
+	}
+
+	for _, s := range final {
+		sort.Strings(s)
+	}
+	return final
+}
--- a/compaction.go
+++ b/compaction.go
@@ -134,7 +134,8 @@ func (c *compactorPool) compactLoop(ctx context.Context) {
 
 func (t *TableBlock) compact(cfg *CompactionConfig) error {
 	var compactionErrors errutil.MultiError
-	t.Index().Ascend(func(i btree.Item) bool {
+	index := t.Index()
+	index.Ascend(func(i btree.Item) bool {
 		granuleToCompact := i.(*Granule)
 		if granuleToCompact.metadata.size.Load() < uint64(t.table.db.columnStore.granuleSizeBytes) {
 			// Skip granule since its size is under the target size.
@@ -456,24 +457,24 @@ func (t *TableBlock) compactGranule(granule *Granule, cfg *CompactionConfig) (bo
 	}
 
 	for {
-		curIndex := t.Index()
+		index := t.Index()
 		t.mtx.Lock()
-		index := curIndex.Clone() // TODO(THOR): we can't clone concurrently
+		newIdx := index.Clone() // TODO(THOR): we can't clone concurrently
 		t.mtx.Unlock()
 
-		if index.Delete(granule) == nil {
+		if newIdx.Delete(granule) == nil {
 			level.Error(t.logger).Log("msg", "failed to delete granule during split")
-			continue
+			return false, fmt.Errorf("failed to delete granule")
 		}
 
 		for _, g := range newGranules {
-			if dupe := index.ReplaceOrInsert(g); dupe != nil {
+			if dupe := newIdx.ReplaceOrInsert(g); dupe != nil {
 				level.Error(t.logger).Log("duplicate insert performed")
 			}
 		}
 
 		// Point to the new index.
-		if t.index.CompareAndSwap(curIndex, index) {
+		if t.index.CompareAndSwap(index, newIdx) {
 			sizeDiff := int64(stats.level1SizeAfter) - (stats.level0SizeBefore + stats.level1SizeBefore)
 			t.size.Add(sizeDiff)
 
--- a/compaction.go
+++ b/compaction.go
@@ -551,7 +551,7 @@ func compactLevel0IntoLevel1(
 		// All the row groups in a part are wrapped in a single row group given
 		// that all rows are sorted within a part. This reduces the number of
 		// cursors open when merging the row groups.
-		bufs = append(bufs, p.Buf().MultiDynamicRowGroup())
+		bufs = append(bufs, buf.MultiDynamicRowGroup())
 	}
 
 	cursor := 0
--- a/compaction.go
+++ b/compaction.go
@@ -16,10 +16,12 @@ import (
 	"github.com/google/btree"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/segmentio/parquet-go"
 	"github.com/thanos-io/objstore/errutil"
 
 	"github.com/polarsignals/frostdb/dynparquet"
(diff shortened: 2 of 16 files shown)


compaction.go
```diff
@@ -541,10 +548,35 @@ func compactLevel0IntoLevel1(
 
 	bufs := make([]dynparquet.DynamicRowGroup, 0, len(level0Parts))
 	for _, p := range partsToCompact {
+		buf := p.Buf()
+
+		// If this is a Arrow record part, convert the record into a serialized buffer
```
asubiotto, line 553: I wonder if a better interface would be to make this internal to the `parts` package and maybe also make the encoding explicit. i.e.:
```
// Returns whether the part is an Arrow record part.
(p *Part) IsArrow() bool

etc...
// AsSerializedBuffer returns the data as a parquet serialized buffer, if this
// is an arrow record part, the record will be converted to a serialized buffer.
(p *Part) AsSerializedBuffer()
```

compaction_test.go
```diff
@@ -249,22 +285,65 @@ func TestCompaction(t *testing.T) {
...
 					},
 				},
 			},
 		},
+		{
+			name: "ArrrowSimpleLevel0ToLevel1",
+			// Insert three rows.
+			inserts: []int64{1, 2, 3, compactCommand},
+			// Insert as an arrow record
+			record: true,
```
asubiotto, line 300: We could also rerun all the test cases above but with arrow records as input instead of parquet, what do you think?

pqarrow/parquet.go
```diff
@@ -27,3 +33,161 @@ func ArrowScalarToParquetValue(sc scalar.Scalar) (parquet.Value, error) {
...
+
+func appendToRow(row []parquet.Value, c arrow.Array, index, rep, def, col int) ([]parquet.Value, error) {
+	switch c.DataType().ID() {
+	case arrow.INT64:
+		array, ok := c.(*array.Int64)
+		if !ok {
+			return nil, fmt.Errorf("column not of expected type")
+		}
+
+		row = append(row, parquet.ValueOf(array.Value(index)).Level(rep, def, col))
```
asubiotto, line 45:
```suggestion
	switch array := c.(type) {
	case *array.Int64:
		row = append(row, parquet.ValueOf(array.Value(index)).Level(rep, def, col))
<etc...>
```

table.go
```diff
@@ -744,21 +763,30 @@ func (t *Table) Iterator(
...
-					// This RowGroup had no relevant data. Ignore it.
-					if len(converter.Fields()) == 0 {
-						continue
-					}
-					if converter.NumRows() >= bufferSize {
-						r := converter.NewRecord()
-						prepareForFlush(r, rgSchema)
-						if err := callback(ctx, r); err != nil {
+
+					switch t := rg.(type) {
```
asubiotto, line 767: nit: add error case

table.go
```diff
@@ -816,25 +844,41 @@ func (t *Table) SchemaIterator(
...
-					}
+
 					b := array.NewRecordBuilder(pool, schema)
 
-					parquetFields := rg.Schema().Fields()
-					fieldNames := make([]string, 0, len(parquetFields))
-					for _, f := range parquetFields {
-						fieldNames = append(fieldNames, f.Name())
-					}
+					switch t := rg.(type) {
```
asubiotto, line 850: nit: add error case

asubiotto (commented):
Nice work

compaction.go
asubiotto: I think I also saw another case below that would benefit from this `AsSerializedBuffer`

bufutils/dyncols.go
```diff
@@ -0,0 +1,24 @@
+package bufutils
+
+import "sort"
+
+func Dedupe(s map[string][]string) map[string][]string {
```
metalmatze, line 5:
```suggestion
// Dedupe the slices of values for each key in the map. 
func Dedupe(s map[string][]string) map[string][]string {
```

compaction.go
thorfour: Great ideas! Will do

compaction_test.go
thorfour: That's a great idea

asubiotto (approved):

dynparquet/example.go
```diff
@@ -50,6 +53,36 @@ func (s Samples) ToBuffer(schema *Schema) (*Buffer, error) {
 	return pb, nil
 }
 
+func (s Samples) ToRecord(schema *arrow.Schema) (arrow.Record, error) {
+	bld := array.NewRecordBuilder(memory.NewGoAllocator(), schema)
+	defer bld.Release()
+
+	numLabels := len(schema.Fields()) - 4
```
metalmatze, line 60: It would be nice to get a comment on why the `- 4` is needed. Is it because of the stack trace, timestamp, and value addition at the end?

thorfour:
Notable updates that have been made since commit [a5ea80e](https://github.com/polarsignals/frostdb/pull/298/commits/a5ea80e5f7957dd11c70ff714d152e333fb0ba81):

- Granules `least` sorting keys are now immutable. When Granules are created they should be created with the parts already to set the least sorting key. This was to prevent the sorting key from changing during runtime.
- There were some corner cases during insert that would allow a part that is smaller than least to be added to a Granule. This is fixed in the Record insertion path where now we create a new granule and insert that granule in this corner case, but remains to be fixed in the normal insertion path.

Issues I'm currently seeing with this branch on Parca is huge CPU usage 
image

Which appears to be caused by GC & compaction

There's also still an accounting error where we try to use a Record that has already been released:
```
panic: runtime error: index out of range [0] with length 0

goroutine 132 [running]:
github.com/apache/arrow/go/v8/arrow/array.(*simpleRecord).Column(0x142d76544e0, 0x0)
	/Users/thor/go/pkg/mod/github.com/apache/arrow/go/v8@v8.0.1/arrow/array/record.go:202 +0x70
github.com/polarsignals/frostdb/pqarrow.getRecordRow(0x14000360380, 0x14115faaa80, {0x107ebbf18, 0x142d76544e0}, 0x0, {0x14002ff0300, 0xc, 0xc}, {0x1400836d200, 0xc, ...})
	/Users/thor/go/src/github.com/polarsignals/frostdb/pqarrow/parquet.go:72 +0x3e0
github.com/polarsignals/frostdb/pqarrow.RecordToRow(0x14000360380, 0x14115faaa80, {0x107ebbf18, 0x142d76544e0}, 0x0)
	/Users/thor/go/src/github.com/polarsignals/frostdb/pqarrow/parquet.go:58 +0xb8
github.com/polarsignals/frostdb/pqarrow.RecordToDynamicRow(0x14000360380, {0x107ebbf18, 0x142d76544e0}, 0x0)
	/Users/thor/go/src/github.com/polarsignals/frostdb/pqarrow/parquet.go:140 +0x180
github.com/polarsignals/frostdb/parts.(*Part).most(0x1425fb0e818)
	/Users/thor/go/src/github.com/polarsignals/frostdb/parts/part.go:190 +0xf4
github.com/polarsignals/frostdb/parts.Part.OverlapsWith({0x0, {0x107ebbf18, 0x142d76544e0}, 0x14000360380, 0xf5, 0x0, 0x140da46e080, 0x0}, 0x14000360380, 0x140a45b0f80)
	/Users/thor/go/src/github.com/polarsignals/frostdb/parts/part.go:224 +0xc0
github.com/polarsignals/frostdb.compactLevel0IntoLevel1(0x14000712310, 0xf5, {0x14013fc0000, 0x21e, 0x350}, {0x141228f7980, 0x4, 0x4}, 0x3fe0000000000000, 0x1425fb0eca0)
	/Users/thor/go/src/github.com/polarsignals/frostdb/compaction.go:529 +0x610
github.com/polarsignals/frostdb.(*TableBlock).compactGranule(0x14000712310, 0x142f9a0f590, 0x14000213188)
	/Users/thor/go/src/github.com/polarsignals/frostdb/compaction.go:361 +0x2ac
	```

thorfour:
I think the CPU usage is a issue of Arrow records being larger in size than parquet files, causing constant compaction/splitting since I was using a much lower setting during testing ` --storage-granule-size=524288`. So that may be a red herring

thorfour:
Not that the above was obtained using a lower granule size setting `--storage-granule-size=524288`

thorfour:
image

thorfour:
image

PR #299, opened 2022-12-29
physicalplan: add partial ordering support do the OrderedAggregate 

### LINKED ISSUE #287: Test flake
```
=== RUN   TestOrderedAggregate/MultiGroupCol
    ordered_aggregate_test.go:130: 
        	Error Trace:	ordered_aggregate_test.go:130
        	            				physicalplan.go:68
        	            				ordered_aggregate.go:272
        	            				ordered_aggregate.go:226
        	            				ordered_aggregate_test.go:188
        	Error:      	Not equal: 
        	            	expected: "a"
        	            	actual  : "b"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-a
        	            	+b
        	Test:       	TestOrderedAggregate/MultiGroupCol
```

@asubiotto

asubiotto:
This PR adds the missing partial ordering support to the OrderedAggregate.
Instead of returning a record on each call to Callback as was previously done,
the OrderedAggregate now buffers the aggregation results and groups for each
ordered set and then merges the results on Finish.

A bunch of incidental bugfixes and test flake fixes have also been included. The first couple of commits also introduce some helper code used by the last (main) commit.

Closes #287 

The ordered aggregate is still not ready for production usage since it doesn't handle columns appearing/disappearing from the input. However, I want to take a different approach (starting down the static schema route I mentioned) to solve this, so would rather merge a complete working version (at least from the unit test perspective) of the OrderedAggregate before working on these edge cases.

In terms of performance difference, as expected, the OrderedAggregate doesn't perform much better than the HashAggregate for QueryMerge (it provides a small 3% perf improvement) given the large number of groups and the new requirement that all data must be buffered in order to merge partially ordered sets. This is expected, and there are some performance angles that can be worked on to improve this (e.g. plan time settings, reducing unnecessary allocations).

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -451,29 +451,12 @@ func (a *HashAggregate) Finish(ctx context.Context) error {
 	aggregateFields := groupByFields
 
 	for _, aggregation := range a.aggregations {
-		var (
-			aggregateArray arrow.Array
-			err            error
-		)
-
 		arr := make([]arrow.Array, 0, numRows)
 		for _, a := range aggregation.arrays {
 			arr = append(arr, a.NewArray())
 		}
 
-		switch aggregation.function.(type) {
-		case *CountAggregation:
-			if a.finalStage {
-				// The final stage of aggregation needs to sum up all the counts of the previous steps,
-				// instead of counting the previous counts.
-				aggregateArray, err = (&Int64SumAggregation{}).Aggregate(a.pool, arr)
-			} else {
-				// If this isn't the final stage we simply run the count aggregation.
-				aggregateArray, err = aggregation.function.Aggregate(a.pool, arr)
-			}
-		default:
-			aggregateArray, err = aggregation.function.Aggregate(a.pool, arr)
-		}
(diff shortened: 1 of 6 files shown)


query/physicalplan/aggregate.go
```diff
@@ -634,3 +617,20 @@ func (a *CountAggregation) Aggregate(pool memory.Allocator, arrs []arrow.Array)
 	}
 	return res.NewArray(), nil
 }
+
+// aggregationHelper is a helper to run the given aggregation function given
+// the set of values. It is aware of the final stage and chooses the aggregation
+// function appropriately.
+func aggregationHelper(
```
brancz, line 624: I think we can probably find a better name for this function. `runAggregation`?

metalmatze (approved):

query/physicalplan/aggregate.go
asubiotto: Done.

asubiotto:
Enhanced the `OrderedAggregate` comment with a more in-depth example as discussed on discord. Let me know if it's understandable.

brancz (approved):
Yes, makes sense!

PR #300, opened 2023-01-02
physicalplan: change from a pre to a post visit plan building strategy

asubiotto:
This is easier to reason about since by using a post visit strategy the plan is built from the scan layer up, rather than the result layer down as was previously done. This also follows our data flow in the execution engine.


brancz (approved):
Agreed that this is more easily understandable!

PR #301, opened 2023-01-04
physicalplan: add OrderedSynchronizer

asubiotto:
This commit adds an OrderedSynchronizer component that merges ordered records
pushed from many concurrent inputs and pushes the merged record to a single
output.

The strategy this commit implements is to block the inputs until all inputs
have pushed a record. This is the most straightforward strategy since the other
option is to store the pushed records, but that requires copying all data which
might be more expensive.

### DIFF
--- a/query/physicalplan/main_test.go
+++ b/query/physicalplan/main_test.go
@@ -1,0 +1,12 @@
+package physicalplan
+
+import (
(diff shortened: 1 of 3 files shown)


brancz (approved):
Nice work, and so little code!

PR #302, opened 2023-01-04
Add Arithmetic Expressions to FrostDB

metalmatze:
Right now the `BinaryExpr` get interpreted as `BooleanExpression` exclusively. 
This needs to be changed so that a BinaryExpr can also be of other types, like AverageExpression, going forward.

The problem with `BooleanExpression` is the fact only returns a bitmap of what rows to keep but doesn't allow to change (in the average case merge columns) in the record. 

https://github.com/polarsignals/frostdb/blob/cbce8509f8bd0b3880774627ff4f8cfe5335c029/query/physicalplan/filter.go#L42

### DIFF
--- a/query/physicalplan/filter.go
+++ b/query/physicalplan/filter.go
@@ -142,7 +142,7 @@ func binaryBooleanExpr(expr *logicalplan.BinaryExpr) (BooleanExpression, error) 
 			Left:  left,
 			Right: right,
 		}, nil
-	case logicalplan.OpMul, logicalplan.OpAdd:
+	case logicalplan.OpAdd, logicalplan.OpSub, logicalplan.OpMul, logicalplan.OpDiv:
 		var leftColumnRef *ArrayRef
 		var leftScalar scalar.Scalar
 		expr.Left.Accept(PreExprVisitorFunc(func(expr logicalplan.Expr) bool {
@@ -272,11 +272,23 @@ func (m *ArithmeticExpr) Project(mem memory.Allocator, r arrow.Record) ([]arrow.
 			for i := int64(0); i < r.NumRows(); i++ {
 				res.Append(left + right)
 			}
+		case "-":
+			// Subtract the two scalars and return a single value as often as the number of rows.
+			for i := int64(0); i < r.NumRows(); i++ {
+				res.Append(left - right)
+			}
 		case "*":
 			// Multiply the two scalars and return a single value as often as the number of rows.
 			for i := int64(0); i < r.NumRows(); i++ {
 				res.Append(left * right)
 			}
+		case "/":
+			// Divide the two scalars and return a single value as often as the number of rows.
+			for i := int64(0); i < r.NumRows(); i++ {
+				res.Append(left / right)
+			}
+		default:
+			return nil, nil, fmt.Errorf("unsupported operation %s", m.operation)
 		}
 
 		fields = append(fields, arrow.Field{Name: m.String(), Type: arrow.PrimitiveTypes.Int64})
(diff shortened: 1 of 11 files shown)


asubiotto (commented):
This is more the direction I was suggesting in the other PR, although as I've mentioned I think we're trying to shoehorn numerical expressions into our existing boolean expression code too much. I would suggest that you think how to generalize our expression code a bit more for us to fit numerical expressions into the execution engine.

One other thing is that there's a focus on making an "average" expression. Rather than thinking about this in terms of aggregations I would say the first step is implementing a generic division operator to be able to run queries like (in logic tests):
```
select a / b
```
This will then be the component you can reuse when building average aggregations (including column naming/renaming which doesn't belong in expression code).

metalmatze:
I've started implementing addition and multiplication as projections. 
Next, I'll tackle subtraction and division before adding the average back.

They'll probably move out of the existing `Binaryexpr` and into an `ArithmeticExpr` struct that doesn't have to implement the `Eval` method and only uses the `Project`.

brancz:
Arithmetic operations should be able to done on any arbitrary source, not just on physical columns. I think keeping that is important.

metalmatze:
Next to columns and scalars, what arbitrary sources do you have in mind?

query/physicalplan/filter.go
```diff
@@ -142,6 +142,52 @@ func binaryBooleanExpr(expr *logicalplan.BinaryExpr) (BooleanExpression, error)
 			Left:  left,
 			Right: right,
 		}, nil
+	case logicalplan.OpAdd, logicalplan.OpSub, logicalplan.OpMul, logicalplan.OpDiv:
```
asubiotto, line 145: It still feels like we're creating this `ArithmeticExpr` at the wrong level of abstraction. The fact that the boolean expressions return a bitmap is an implementation detail abstracted out by the `columnProjection` interface. The boolean expressions are abstracted out by the `binaryExprProjection` implementation. I think what we should do is rename `binaryExprProjection` to `boolExprProjection` and add a new implementation of `columnProjection` that implements arithmetic expression evaluation (and is constructed with this arithmetic expr). Or something along those lines.

query/physicalplan/filter.go
```diff
@@ -142,6 +142,52 @@ func binaryBooleanExpr(expr *logicalplan.BinaryExpr) (BooleanExpression, error)
...
+		if rightColumnRef == nil && rightScalar == nil {
+			return nil, errors.New("right side of arithmetic expression must be a column or a scalar")
+		}
+
+		return &ArithmeticExpr{
+			left:        leftColumnRef,
+			leftScalar:  leftScalar,
+			right:       rightColumnRef,
+			rightScalar: rightScalar,
+			operation:   expr.Op.String(),
```
asubiotto, line 189: Why use the string representation instead of the `Op` itself since it's an enum?

query/physicalplan/filter.go
```diff
@@ -197,6 +243,237 @@ func (a *OrExpr) String() string {
...
+	leftScalar  scalar.Scalar
+	right       *ArrayRef
+	rightScalar scalar.Scalar
+}
+
+func (m *ArithmeticExpr) Eval(r arrow.Record) (*Bitmap, error) {
+	return nil, fmt.Errorf("not used")
+}
+
+func (m *ArithmeticExpr) Project(mem memory.Allocator, r arrow.Record) ([]arrow.Field, []arrow.Array, error) {
```
asubiotto, line 258: Just a reminder that you probably need to handle nulls here otherwise we'll get weird results and/or panics.

query/physicalplan/project.go
```diff
@@ -88,6 +88,11 @@ func (b binaryExprProjection) Project(mem memory.Allocator, ar arrow.Record) ([]
 		}
 	}
 
+	// TODO: It probably makes sense to create arithmetics outside binaryExprProjection?
```
asubiotto, line 91: Yes, exactly. This is what I'm talking about in my other comment.

query/physicalplan/filter.go
metalmatze: Yes, that was my thought too. Ok, good to know this is the preferred way going forward. :relieved:

PR #304, opened 2023-01-05
physicalplan: make OrderedAggregate work in the presence of dynamic columns

asubiotto:
This required adding some more fields to the OrderedAggregate to keep track of which columns we have/have not seen.

### DIFF
--- a/pqarrow/arrowutils/groupranges.go
+++ b/pqarrow/arrowutils/groupranges.go
@@ -149,6 +149,18 @@ func GetGroupsAndOrderedSetRanges(
 					return nil, nil, nil, err
 				}
 			}
+		case VirtualNullArray:
+			for j := 0; j < arr.Len(); j++ {
+				cmp, ok := nullComparison(curGroup[i] == nil, true)
+				if !ok {
+					return nil, nil, nil, fmt.Errorf(
+						"null comparison should always be valid but group was: %v", curGroup[i],
+					)
+				}
+				if err := handleCmpResult(cmp, i, t, j); err != nil {
+					return nil, nil, nil, err
+				}
+			}
 		default:
 			panic("unsupported type")
 		}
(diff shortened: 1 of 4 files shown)


pqarrow/arrowutils/groupranges.go
```diff
@@ -149,6 +149,16 @@ func GetGroupsAndOrderedSetRanges(
 					return nil, nil, nil, err
 				}
 			}
+		case VirtualNullArray:
+			for j := 0; j < arr.Len(); j++ {
+				cmp, ok := nullComparison(curGroup[i] == nil, true)
+				if !ok {
+					panic("null comparison should always be valid")
```
metalmatze, line 156: Cannot this be an error?

pqarrow/arrowutils/nullarray.go
```diff
@@ -0,0 +1,74 @@
...
+func (n VirtualNullArray) IsNull(_ int) bool {
+	return true
+}
+
+func (n VirtualNullArray) IsValid(_ int) bool {
+	return false
+}
+
+func (n VirtualNullArray) Data() arrow.ArrayData {
+	panic("VirtualNullArray: Data not implemented")
```
brancz, line 65: is this not needed by flight?

brancz (approved):
Nice work!

pqarrow/arrowutils/nullarray.go
asubiotto: This virtual null array is only used internally by operators (just the ordered agg) and materialized into a physical null array of a given type before being passed on downstream so would never be serialized by flight.

asubiotto:
Looks like CI found a flake, will investigate and fix.

pqarrow/arrowutils/nullarray.go
brancz: gotcha, sounds good

asubiotto:
The flake was due to expecting ordering due to map iteration. This is now fixed.

PR #305, opened 2023-01-11
logictest: add ability to explain plans

asubiotto:
This commit additionally adds two high-level directories to the testdata logic tests. One is exec, where all the execution-related tests can be found, and one is plan, where all the planning-related tests can be found. My experience at Cockroach has shown that it is better to separate these two test types.


brancz (approved):
lgtm

metalmatze:
This is so cool! :tada: :clap:

PR #307, opened 2023-01-16
physicalplan: plan ordered aggregations

asubiotto:
This PR is a bunch of commits with fixes and planning changes in order to plan ordered aggregations when the grouping columns are a prefix of ordering columns. This includes cases in which a filter "covers" a prefix of the ordering columns up to the group columns (e.g. when the order is "name, labels, timestamp" and we filter name="somename" and group by (labels,timestamp)). For more explicit information on what cases are and aren't covered, please refer to the `plan/aggregate` logic test which goes into more detail.

Ordered aggregations are disabled by default and can be enabled (as is done with logic tests) by calling `query.NewEngine` and adding the `WithOrderedAggregations` option given that there are most likely some latent bugs.

The performance improvement over hash aggregates is reasonable as expected and I believe that there is some more room for improvement. I want to write up a couple of issues with some ideas for perf improvements on the current state:
```
Query/Range-8    4.82ms ± 1%    4.51ms ± 0%   -6.47%  (p=0.000 n=10+8)

Query/Range-8    22.3MB ± 0%    21.4MB ± 0%   -4.32%  (p=0.000 n=10+10)

Query/Range-8     9.55k ± 0%     7.58k ± 0%  -20.61%  (p=0.000 n=9+10)
```

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -26,7 +26,8 @@ func Aggregate(
 	s *parquet.Schema,
 	agg *logicalplan.Aggregation,
 	final bool,
-) (*HashAggregate, error) {
+	ordered bool,
+) (PhysicalPlan, error) {
 	aggregations := make([]Aggregation, 0, len(agg.AggExprs))
 
 	for _, expr := range agg.AggExprs {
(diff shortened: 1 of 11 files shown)


query/physicalplan/aggregate.go
```diff
@@ -26,7 +26,8 @@ func Aggregate(
 	s *parquet.Schema,
 	agg *logicalplan.Aggregation,
 	final bool,
-) (*HashAggregate, error) {
+	ordered bool,
```
metalmatze, line 29: Eventually, it might make sense to pass something similar to the execOpts or an execution context here.

metalmatze (commented):
Very nice!
The planning logic is relatively easy to understand and follow! :+1:

brancz (approved):
Really nice, I love how when I started reading the PR I asked myself "how do we decide to plan an ordered aggregation" and I just had to scroll down to see the answer!

lgtm 🎉 👍

PR #310, opened 2023-01-23
Insert fixes

thorfour:
This refactors the `parts` object into their own separate package. 
It refactors the Granule object to make least a field of Granules that does not get updated when new parts are appended. This is important because least is the sorting key of a Granule for the index so it shouldn't be getting updated at runtime.

It also adds a fix for insertion where in some cases the Granule was having new least parts inserted into it instead of inserting a new Granule into the index.


granule.go
```diff
@@ -38,52 +36,49 @@ type GranuleMetadata struct {
...
-				break
-			}
-		} else {
-			break
-		}
+	_ = g.parts.Prepend(p)
+	g.metadata.size.Add(uint64(p.Size()))
+
+	if g.metadata.least == nil || g.tableConfig.schema.RowLessThan(r, g.metadata.least) {
+		g.metadata.least = r
```
asubiotto, line 66: I'm not sure I understand, I thought this commit made least immutable but here you're overwriting with a new part's least row?

thorfour: So this _only_ happens at `NewGranule` creation, which you can initialize a granule with many parts (we do this in compaction). During that initialization we find and set the least row to use as a sorting key. After creation that value is no longer able to be updated.

brancz (approved):

Issue #311, opened 2023-01-24
wal: improve durability


asubiotto:
This issue tracks a couple of improvements to our handling of the write ahead log uncovered while running frostdb in production.

- [ ] Initialize in-memory state to on-disk state.
The WAL package that we use uses an index for each entry written to the WAL. The writer provides this index when writing to the WAL (https://github.com/tidwall/wal/blob/master/wal.go#L310). The WAL records the "last" index it has seen and the next write's index *must* be `lastIndex + 1`, otherwise the write is rejected. In frostdb, we conveniently use the txn id as the WAL index for each WAL entry https://github.com/polarsignals/frostdb/blob/e386dd63002bbb0eaf74999ff5f6ce8baabbc3fb/wal/wal.go#L193
When restarting an in-memory frostdb instance, the transaction ID is reset to 1 (since we have no on-disk state). This presents an issue when writing to the WAL because all writes to the WAL will be rejected as out-of-order (something we're observing) since the transaction id will be less than the `lastIndex` in the WAL (it is read from the lastest entry in WAL files). The proposal is to initialize the transaction high watermark to the on-disk WAL index. This can be obtained by first opening the wal and calling `LastIndex` https://github.com/tidwall/wal/blob/0fa32977a196086591e6eace7bdf13f67bc305aa/wal.go#L504.

- [ ] Truncate torn writes to avoid WAL corruption.
We have also observed `log corrupt` errors when reopening a WAL. After further analysis, this is caused by the fact that only half an entry has been persisted to disk (even when calling fsync). Concretely, WAL entries are written as `<size of entry><entry>` and we encounter an entry with a size larger than the following bytes. I don't think there is a mechanism in our WAL library that allows us to truncate corrupt entries and in any case we can't open a corrupt WAL to perform operations on it. Here, we probably have to write some custom WAL file reading to read the last file in the WAL (since corrupt entries should always be the last entries) and truncate after the penultimate entry (by reading the file and interpreting the size of each entry).

brancz:
Regarding the first point, since we use the transaction, why does this happen even though after replaying we set the "correct" next transaction ID? https://github.com/polarsignals/frostdb/blob/e386dd63002bbb0eaf74999ff5f6ce8baabbc3fb/wal/wal.go#L301

brancz:
Regarding the second point, I think we could even start with deleting/skipping the last file and do what you suggested as a follow up.

metalmatze:
[...]

PR #313, opened 2023-01-24
wal: try repair the wal when encountering log corrupt error

asubiotto:
These errors can be encountered due to torn writes not fully writing tailing WAL entries. The approach this commit takes is to delete the last file in the directory, which should be where the corruption is, before reattempting an Open.

cc @brancz 

Addresses #311


metalmatze (approved):
Very nice!

PR #314, opened 2023-01-25
physicalplan: allow specifying input to Build

asubiotto:
This allows Build to work on arbitrary input plans.


asubiotto:
I'm going to merge this on green since it's a small change that we need on cloud.

PR #316, opened 2023-01-26
support arrow dictionaries

thorfour:
**NOTE** This is a breaking change as is. 

This adds support for arrow dictionaries both in the new arrow ingestion path, but also in the query path. It determines if a dictionary should be used based on the Parquet schema (if the Parquet schema's encoding is a dictionary type). 

If we don't want to to implement this as a breaking change we can add a feature flag that can turn on the dictionary encoding detection, but since the adaption for Parca is very straightforward I recommend we leave it as a breaking change.


brancz:
Just making sure I'm understanding this right, this is a breaking change because of the type that needs to be type-asserted for building the buffers, is that right?

If so I'm also ok with the breaking change and lgtm

thorfour:
> Just making sure I'm understanding this right, this is a breaking change because of the type that needs to be type-asserted for building the buffers, is that right?
> 
> If so I'm also ok with the breaking change and lgtm

That's correct

Issue #317, opened 2023-01-26
WAL: Improve replay memory pressure


metalmatze:
### Problem
When starting a process the WAL might get replayed. If there is a lot of data in the WAL all of the data gets inserted into FrostDB without limitations which is way faster compared to normal operation.
What we're seeing now, is that FrostDB has a hard time keeping up compacting the interserted records fast enough resulting in using a lot more memory than necessary and potentially getting OOM killed.

### Proposal

Inform the WAL replay on how fast to insert based on the compaction back pressure.

As a start, we can check after every insert from the WAL if compaction is running. 
In case compaction is running the WAL will hold until compaction is done and only then continue. 
While this might not be the fastest way to solve the problem it seems like a pragmatic solution for now.

In the future, it might make sense to replay the WAL based on the overall active memory and only slow down if the memory limit is sooner after reached. 

cc @brancz

brancz:
Slight detail, we can't tell whether compactions are running due to the lock-free nature, but we can scan the index to check whether compactions _should be running_. And only continue inserting when no further compactions should be running.

asubiotto:
I think something to consider is that I vaguely recall that compaction doesn't consider the records inserted during Wal replay because the txn on the parts is further ahead than what the txn high watermark of the database is (this isn't updated until the end of the replay). Perhaps as a first step we can see how allowing compactions to occur during Wal replay affects memory usage. This should at least avoid one huge compaction after replay.

metalmatze:
Good catch @asubiotto! I'll look into that first.

asubiotto:
Can this be closed now?

brancz:
It appears replaying no longer uses more memory than regular operation, so I think we can close this.

PR #318, opened 2023-01-26
Unblock compaction by setting tx during WAL replay

metalmatze:
This is an idea for #317 

Sadly the bytes on the heap read from the runtime don't seem to significantly improve with this approach. :thinking: 

Screenshot from 2023-01-26 16-25-10

_This chart shows the memory heap in bytes every 128 inserts_


metalmatze:
Setting the `tx` and `highWatermark` has now been implemented (the yellow line).
WAL Replay   Compaction

brancz (approved):

PR #320, opened 2023-01-28
Arrow wal

thorfour:
Enables basic wal support for arrow records using the arrow ipc readers/writers to serialized records to byte slices.


metalmatze (approved):
Very nice!

PR #321, opened 2023-01-30
Handle NULLs from dictionaries in Distinct

metalmatze:
With the usage of dictionaries, we get `colHashes` that have value `0`. 
If the entire hash is `0` the row had only NULL values. This is most noticeable for Distinct on a specific column, like `labels.compiler` for example.

This fixes at least an issue where the final seen had `0, 123, 3345` as values where `123, 3345` should be the only valid distinct values in the end.
Sadly, I cannot really explain this shows now with the move to Dictionaries. Maybe it has been there all along.


brancz:
Can we write a test for this?

metalmatze:
I'm still not sure if completely getting rid of the `NULL` here is semantically correct.
Maybe it is for now, but overall I get the impression that we need to support NULLs better and this wouldn't be more of a hack.

brancz:
I see, no I think that's not quite right. At this layer, the distinct should return a single `NULL`. The layer above eg. PromQL or Parca's query language then decides what to do, but here the correct thing is to include `NULL` in the result set.

metalmatze:
Updated and only added a single test to not regress in the future.

PR #323, opened 2023-02-07
wal: miscellaneous fixes

asubiotto:
This PR proposes a couple of changes to the WAL to protect from getting into unwanted state.

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -210,9 +210,9 @@ func (w *FileWAL) Run(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			w.mtx.Lock()
-			len := w.queue.Len()
+			n := w.queue.Len()
 			w.mtx.Unlock()
-			if len > 0 {
+			if n > 0 {
 				// Need to drain the queue before we can shutdown.
 				continue
 			}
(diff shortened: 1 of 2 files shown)


wal/wal.go
```diff
@@ -202,6 +209,55 @@ func tryRepairWAL(path string) error {
...
+}
+
+func (w *FileWAL) stateTransition(newState walState) error {
+	var curState walState
+	for {
+		curState = walState(w.state.Load())
+		if curState == newState {
+			return nil
+		}
+		if ok := w.state.CompareAndSwap(uint32(curState), uint32(newState)); ok {
```
thorfour, line 250: does this need to be atomic? Could we instead have the wal move it's state to ready/writing once replay has finished? So that way it only has a single call

asubiotto:
Moved the compaction commit to a separate PR: https://github.com/polarsignals/frostdb/pull/324

wal/wal.go
asubiotto: We could, but we don't necessarily replay a WAL when starting up so we also need a mechanism to avoid replays if we have written at least once. I agree that this is not perfect but not sure what else we could do to support both cases.

thorfour: What if we changed that? Instead of `ReplayWALs` being a public function that users need to call that it's apart of creating a new columnstore. That way it can always move the WAL through the state machine

asubiotto: I think I suggested this a couple of weeks ago (what you suggest would be my preferred approach) but there was a bit of push back from @brancz and @metalmatze due to wanting to be able to indicate that the grpc server is healthy before replaying, but not ready until then. Should we revisit?

asubiotto: Removed this state commit and tracking the change to the API in #328 since it should probably be a separate PR.

asubiotto:
RFAL

wal/wal.go
metalmatze: Just for completeness as this has been proposed on Discord:
If we can get the approach of registering the ColumnStore server at a later time to an existing gRPC server to work we should be fine with replaying during creating the column store.

metalmatze (approved):

PR #325, opened 2023-02-07
sanity check in granule append

thorfour:


granule.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/apache/arrow/go/v10/arrow"
 	"github.com/google/btree"
+	"github.com/pingcap/errors"
```
asubiotto, line 9: I think we should just use the golang errors package. Where else do we use this package?

granule.go
```diff
@@ -76,6 +77,15 @@ func (g *Granule) Append(p *parts.Part) (uint64, error) {
 	node := g.parts.Prepend(p)
 	newSize := g.metadata.size.Add(uint64(p.Size()))
 
+	least, err := p.Least()
+	if err != nil {
+		return 0, err
+	}
+
+	if g.tableConfig.schema.RowLessThan(least, g.metadata.least) {
+		return 0, errors.New("added part to Granule that is less than least")
```
asubiotto, line 86: Is there a way to pretty print the rows so we can have more information? Otherwise I think it would be interesting to also print the result of `RowLessThan(g.metadata.least, least)` to see if they are equal

asubiotto:
I like the idea of sanity checks, maybe we should add some in compaction too?

thorfour:
> 

I was actually leaning towards closing this instead, it felt like extra work that doesn't gain us much. Really only useful for debug

asubiotto:
long-term we could use compile-time directives to enable sanity checks

PR #326, opened 2023-02-07
use option pattern for iteropts

thorfour:
Refactored the `IterOptions` in the table interface to consume the option pattern to prevent having to pass empty structs for lots of unit testing.


brancz (approved):
lgtm

PR #329, opened 2023-02-08
*: fix all lint issues

asubiotto:
This PR fixes all the lint issues reported by golangci-lint that have creeped into the codebase over time as well as ignoring lint issues that would take too long to fix.

My idea is to make passing golangci-lint a required step before merging, otherwise there doesn't seem to be much use in having a linter.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,9 +18,8 @@ linters:
 
 issues:
   exclude-rules:
-    - path: _test.go
-      linters:
-        - errcheck
+    - path: bench_test.go
+      text: 'is unused'
 
 linters-settings:
   depguard:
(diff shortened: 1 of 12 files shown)


metalmatze:
Oh wow. Yes we should totally make it a required step. 
Do you want to do it in this PR or follow up?

metalmatze (approved):

asubiotto:
I don't think I have the necessary permissions to do this (it can be found under the repo's settings tab->code and automation)

metalmatze:
> I don't think I have the necessary permissions to do this (it can be found under the repo's settings tab->code and automation)

Added the `lint` and `build` steps to be required.

asubiotto:
Thanks!

Issue #322, opened 2023-02-06
Panic in binaryscalarexpr.go with array.Dictionary


metalmatze:
There seems to be a panic when querying with the dictionary types:
https://github.com/polarsignals/frostdb/blob/395932c312ca14c228bc94292fba8930efaede3c/query/physicalplan/binaryscalarexpr.go#L144

It seems like we should support more of the logicalplan operators. Sadly it's not obvious from the stacktrace at which one we should the closet. 

```
panic: unsupported value comparison: Type(?) [recovered]
	panic: unsupported value comparison: Type(?) [recovered]
	panic: unsupported value comparison: Type(?)


To see all goroutines, visit https://github.com/maruel/panicparse#gotraceback

1: running [Created by errgroup.(*Group).Go @ golang.org/x/sync@v0.1.0/errgroup/errgroup.go:72]
    trace    go.opentelemetry.io/otel/sdk@v1.11.2/trace/span.go:383                                     (*recordingSpan).End.func1()
    trace    go.opentelemetry.io/otel/sdk@v1.11.2/trace/span.go:421                                     (*recordingSpan).End(0xc0524d8f00, {0, 0, 0x1d6cbdd500a4d2d8})
             runtime/panic.go:884                                                                       panic({#1, #7})
    trace    go.opentelemetry.io/otel/sdk@v1.11.2/trace/span.go:383                                     (*recordingSpan).End.func1()
    trace    go.opentelemetry.io/otel/sdk@v1.11.2/trace/span.go:421                                     (*recordingSpan).End(0xc056039500, {0, 0, 0})
             runtime/panic.go:884                                                                       panic({#1, #7})
    frostdb  github.com/polarsignals/frostdb@v0.0.0-20230127114537-7817186cf2b4/binaryscalarexpr.go:144 compare({0, 0, 0, 0, 0, 0}, {0xc0974ed930, 0x10, 0xf9, 0, ...})
    frostdb  github.com/polarsignals/frostdb@v0.0.0-20230127114537-7817186cf2b4/binaryscalarexpr.go:120 Max({#4, #6})
    frostdb  github.com/polarsignals/frostdb@v0.0.0-20230127114537-7817186cf2b4/binaryscalarexpr.go:75  BinaryScalarOperation({#4, #6}, {#8, 0x13, 0xf9, 0, 0, 0}, 0x19e8728)
[...]

PR #330, opened 2023-02-08
Create TestBinaryScalarOperation

metalmatze:
In an attempt to recreate #322 I've created `TestBinaryScalarOperation` that will have many different variants of comparing BinaryScalars.
Sadly, for now, I couldn't reproduce the panic in #322 and will continue with that tomorrow. 
As so often, it might be related to `null`.

Ideas welcome.

### DIFF
--- a/binaryscalarexpr_test.go
+++ b/binaryscalarexpr_test.go
@@ -1,0 +1,52 @@
+package frostdb
+
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/segmentio/parquet-go"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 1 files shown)


brancz:
How come this isn't a logic test?

metalmatze:
I started out with this test thinking it was too low level for logic tests. I think you have a point where we can explore how we can cover this via logic tests.
One area I'm not sure that logic tests will be enough though is testing different types that we would compare. This is ultimately related to the bug from #322  I'm trying to hunt down.

metalmatze:
Just checked and most usual cases in the `BinaryScalarOperation` are already covered in the filter logic tests :+1:

PR #333, opened 2023-02-09
dynparquet: create comparison rows according to sorting schema

asubiotto:
When comparing two rows, the comparison code needs to merge the dynamic columns found in each row and create a copy of each row with the union of both dynamic column sets. These new row values were previously created in the same order as the sorting column names. However parquet assigns column indexes to each column name and this index does not necessarily correspond to the index into the slice of sorting columns. Therefore, the parquet comparison code could compare a value at an incorrect column index.

This commit iterates over the merged schema so that parquet comparison code lookups work successfully.


thorfour (approved):
Great find!

PR #339, opened 2023-02-10
*: remove manual creation of directories in tests

asubiotto:
A lot of tests would create temporary directories and then clean them up, or do so and forget to (e.g. filesystem.NewBucket). This commit switches all these instances to use t.TempDir() which takes care of all the directory creation/cleanup for tests and returns the path.


metalmatze (approved):
sweet

PR #340, opened 2023-02-13
frostdb: skip replaying WAL writes for tables that have already been persisted

asubiotto:
When replaying writes, we would only check if the table did not exist to skip replaying a write, implying that the new table block entry had been truncated and therefore persisted. However, sometimes the database can only truncate entries if ALL tables have been persisted and only then up to the minimum transaction persisted across all tables. This means that even if a block belonging to a larger table is persisted, writes belonging to this block might not be truncated from the WAL yet. This commit skips these writes if it sees that the block the writes belong to were already persisted, reducing unnecessary memory pressure on the databse.

### DIFF
--- a/db.go
+++ b/db.go
@@ -390,17 +390,13 @@ func (db *DB) walDir() string {
 }
 
 func (db *DB) replayWAL(ctx context.Context) error {
-	persistedBlocks := map[ulid.ULID]struct{}{}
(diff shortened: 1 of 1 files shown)


metalmatze (approved):
Very nice! Makes a ton of sense.

PR #343, opened 2023-02-14
dynparquet: don't recompute dynamic column full name

asubiotto:
This was a hot path during WAL replay. Computing the full name only once
provides an improvement when replaying production WALs:

```
Replay-8     32.3s ± 2%     28.1s ± 9%  -13.02%  (p=0.000 n=9+10)

Replay-8    81.5GB ± 1%    75.1GB ± 2%   -7.79%  (p=0.000 n=9+10)

Replay-8      704M ± 2%      532M ± 4%  -24.51%  (p=0.000 n=9+10)
```

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -347,3 +347,21 @@ func BenchmarkQuery(b *testing.B) {
 		}
 	})
 }
+
+func BenchmarkReplay(b *testing.B) {
+	b.Skip(skipReason)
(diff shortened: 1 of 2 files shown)


metalmatze (approved):
Oh wow. Wouldn't have expected this to have such an impact.
Benchmarks :heart:

PR #346, opened 2023-02-14
*: check rows are sorted on insert

asubiotto:
This commit checks that an invariant we expect, that rows are sorted on insert, is upheld. If not, this commit increments a metric and sorts the rows.

### DIFF
--- a/dynparquet/row.go
+++ b/dynparquet/row.go
@@ -47,6 +47,15 @@ func (r *DynamicRows) GetCopy(i int) *DynamicRow {
 	}
 }
 
+func (r *DynamicRows) IsSorted(schema *Schema) bool {
+	for i := 1; i < len(r.Rows); i++ {
+		if schema.RowLessThan(r.Get(i), r.Get(i-1)) {
+			return false
+		}
+	}
+	return true
+}
+
 func NewDynamicRow(row parquet.Row, schema *parquet.Schema, dyncols map[string][]string, fields []parquet.Field) *DynamicRow {
 	return &DynamicRow{
 		Row:            row,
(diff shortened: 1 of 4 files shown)


metalmatze (approved):
Nice. Makes sense. 
We also have a metric for the total amount of inserted rows so we can nicely calculate the percentage of sorted rows :+1:

asubiotto:
I'm going to hold off on merging until I investigate things a little more, looks like the check is failing on some wals and I want to make sure things are ironed out first.

asubiotto:
In any case, it seems like this `IsSorted` check is very expensive (doubles replay times), so either we make it cheaper (which the whole database will benefit from) or we put this behind a debug tag.

asubiotto:
This PR is ready for a review again! I added `log(n)` inserts that without the `IsSorted` check really improve replay times:
```
Replay-8     60.6s ± 2%     40.5s ± 1%  -33.14%  (p=0.008 n=5+5)

Replay-8    87.3GB ± 1%    66.0GB ± 0%  -24.35%  (p=0.008 n=5+5)

Replay-8     1.10G ± 0%     0.73G ± 0%  -33.69%  (p=0.008 n=5+5)
```

However, with the addition of `IsSorted`, these two commits make it so that basically the check is free. I would advocate for merging this and removing the check once we are certain we have no more unsorted inserts (I think views are the last place where we insert unsorted rows):
```
Replay-8     60.6s ± 2%     60.1s ± 2%   ~     (p=0.310 n=5+5)

Replay-8    87.3GB ± 1%    87.3GB ± 1%   ~     (p=0.690 n=5+5)

Replay-8     1.10G ± 0%     1.10G ± 0%   ~     (p=0.841 n=5+5)
```

asubiotto:
I think there are even more improvements we can make in the insert path, but this is fine for now.

brancz (approved):
Makes so much sense 🤯

PR #349, opened 2023-02-16
*: fix ReadRows usage

asubiotto:
Rows read using ReadRows are not safe to reuse after a subsequent call to ReadRows. This commit changes calls to either use the rows right after the ReadRows call or calls Clone on the rows. Calling Clone has a slight perf hit on simple insert benchmarks:

```
name            old time/op    new time/op    delta
InsertSimple-8    15.7ms ± 1%    16.3ms ± 2%  +3.84%  (p=0.000 n=9+10)

name            old alloc/op   new alloc/op   delta
InsertSimple-8    13.5MB ± 1%    14.2MB ± 1%  +5.25%  (p=0.000 n=10+10)

name            old allocs/op  new allocs/op  delta
InsertSimple-8      150k ± 0%      164k ± 0%  +9.35%  (p=0.000 n=10+10)
```

But improves things when replaying a production WAL.

```
Replay-8     65.6s ± 3%     60.6s ± 2%  -7.68%  (p=0.008 n=5+5)

Replay-8    90.6GB ± 0%    87.3GB ± 1%  -3.70%  (p=0.008 n=5+5)

Replay-8     1.18G ± 0%     1.10G ± 0%  -6.83%  (p=0.008 n=5+5)
```

This is because rows will actually be in sorted order, making it more efficient to insert into a b-tree when the insert batch is large.

### DIFF
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -2,6 +2,7 @@ package dynparquet
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"sort"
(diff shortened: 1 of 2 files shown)


asubiotto:
Introducing bugs that I then fix is what I call job security.

PR #351, opened 2023-02-16
Expect `io.EOF` at the end of files

metalmatze:
We expect an io.EOF at the end of files and in that case return total bytes read.

This is not an issue for the header of ranges within the file, however, especially when reading the parquet footer we read until the end of the file and this would previously return io.EOF even though it's expected.

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -7,6 +7,7 @@ package storage
 
 import (
 	"context"
+	"errors"
 	"io"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


storage/bucket.go
```diff
@@ -59,7 +59,8 @@ func (b *FileReaderAt) ReadAt(p []byte, off int64) (n int, err error) {
 	for total < len(p) { // Read does not guarantee the buffer will be full, but ReadAt does
 		n, err = rc.Read(p[total:])
 		total += n
-		if err != nil {
+		// If io.EOF is returned it means we read the end of the file and simply return the total
+		if err != nil && err != io.EOF {
```
asubiotto, line 63: nit: I prefer using `errors.Is` (it's more complete). I also have a minor suggestion to make this code clearer
```suggestion
		if err != nil {
		    if errors.Is(err, io.EOF) {
		        // If io.EOF is returned it means we read the end of the file and simply return the total.
		        break
		    }
```

metalmatze: I was wondering if we should use the newer APIs. Thanks!

metalmatze:
Hold off merging this until we know what semantics parquet-go expects from `io.ReadAt` interfaces.

PR #352, opened 2023-02-16
Add support for filtering for IS NULL or IS NOT NULL

### LINKED ISSUE #350: Filtering for existing labels being `NULL`
With system-wide profiling in Parca and Parca Agent it might be interesting to once in a while filter for all series that don't have a specific label. Like, "show me all series that don't have a `namespace` label", basically excluding everything that's Kubernetes related. 

Here are some examples we should support in FrostDB and add to the [filter](https://github.com/polarsignals/frostdb/blob/main/logictest/testdata/exec/filter/filter) logic test.

```
exec
select labels, stacktrace, timestamp, value where labels.label5 = ''
----
value1  value2  null    null    stack1  1       1
value2  value2  value3  null    stack1  2       2
value3  value2  null    value4  stack1  3       3

exec
select labels, stacktrace, timestamp, value where labels.label4 = null
----
value1  value2  null    null    stack1  1       1
value2  value2  value3  null    stack1  2       2
```

We can debate whether the first one makes sense. However, the concept of `null` isn't really a concept in PromQL and most other languages afaik.

metalmatze:
Closes #350

### DIFF
--- a/binaryscalarexpr.go
+++ b/binaryscalarexpr.go
@@ -69,6 +69,13 @@ var ErrUnsupportedBinaryOperation = errors.New("unsupported binary operation")
 func BinaryScalarOperation(left parquet.ColumnChunk, right parquet.Value, operator logicalplan.Op) (bool, error) {
 	switch operator {
 	case logicalplan.OpEq:
+		if right == parquet.NullValue() {
+			// Assume all ColumnChunk have NULLs for now.
+			// They will be read and added to a bitmap later on.
+			// TODO: Maybe there's a nice way of reading the NumNulls from the Pages, for me they always return 0
+			return true, nil
+		}
+
 		bloomFilter := left.BloomFilter()
 		if bloomFilter == nil {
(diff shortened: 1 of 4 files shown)


kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

brancz (commented):
I think we need an extra case (or at least a test to prove it works), for when the column doesn’t exist at all in the underlying row groups. Those row groups should be returned by that filter.

metalmatze:
> I think we need an extra case (or at least a test to prove it works), for when the column doesn’t exist at all in the underlying row groups. Those row groups should be returned by that filter.

Yes, agreed. That's a follow-up. I left it out of this because this is mostly a correctness thing for the DB and not specifically a use case for Parca.

metalmatze:
> I think we need an extra case (or at least a test to prove it works), for when the column doesn’t exist at all in the underlying row groups. Those row groups should be returned by that filter.

Opened #353

PR #358, opened 2023-02-27
query: Return error if schema is nil after building physical plan

metalmatze:
This will mostly likely occur in a distributed storage fashion and is less likely for Parca itself.

### DIFF
--- a/query/engine.go
+++ b/query/engine.go
@@ -2,6 +2,7 @@ package query
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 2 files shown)


brancz (approved):

metalmatze:
Turns out we have some invariants for startup where the schema can be nil. Those queries don't touch the aggregation, so it's fine. Therefore the check now lives right before the actual aggregation planning happens where we need the schema not to be nil.

PR #362, opened 2023-03-01
Fix panic on nil cols

thorfour:
The Min/Max functions could panic with 
```
panic: unsupported value comparison: Type(?) [recovered]
	panic: unsupported value comparison: Type(?)

goroutine 5 [running]:
testing.tRunner.func1.2({0x102a34080, 0x1400033b990})
	/usr/local/go/src/testing/testing.go:1396 +0x1c8
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1399 +0x378
panic({0x102a34080, 0x1400033b990})
	/usr/local/go/src/runtime/panic.go:884 +0x204
github.com/polarsignals/frostdb.compare({0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, {0x0, 0x0, 0x0, 0x0, ...})
	/Users/thor/go/src/github.com/polarsignals/frostdb/binaryscalarexpr.go:151 +0x320
```

When a column chunk had all null values. This adds a check to those functions to properly handle a empty column chunk.

### DIFF
--- a/binaryscalarexpr.go
+++ b/binaryscalarexpr.go
@@ -111,7 +111,13 @@ func Min(chunk parquet.ColumnChunk) parquet.Value {
 	columnIndex := chunk.ColumnIndex()
 	min := columnIndex.MinValue(0)
 	for i := 1; i < columnIndex.NumPages(); i++ {
-		if v := columnIndex.MinValue(i); compare(min, v) == 1 {
+		v := columnIndex.MinValue(i)
+		if min.IsNull() {
+			min = v
+			continue
+		}
+
+		if compare(min, v) == 1 {
 			min = v
 		}
 	}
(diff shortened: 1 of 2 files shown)


brancz (approved):
Oh very interesting! Nice find! lgtm

binaryscalarexpr.go
```diff
@@ -111,7 +111,13 @@ func Min(chunk parquet.ColumnChunk) parquet.Value {
 	columnIndex := chunk.ColumnIndex()
 	min := columnIndex.MinValue(0)
 	for i := 1; i < columnIndex.NumPages(); i++ {
-		if v := columnIndex.MinValue(i); compare(min, v) == 1 {
+		v := columnIndex.MinValue(i)
+		if min.IsNull() {
+			min = v
+			continue
```
asubiotto, line 117: nothing is less than `NULL` I think, so we could just return here as long as the ordering is ascending.

binaryscalarexpr_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+func (f *FakeColumnIndex) NullCount(int) int64        { return 0 }
+func (f *FakeColumnIndex) NullPage(int) bool          { return false }
+func (f *FakeColumnIndex) MinValue(int) parquet.Value { return parquet.ValueOf(nil) }
+func (f *FakeColumnIndex) MaxValue(int) parquet.Value { return parquet.ValueOf(nil) }
+func (f *FakeColumnIndex) IsAscending() bool          { return false }
+func (f *FakeColumnIndex) IsDescending() bool         { return false }
+
+/*
+	This is a regression test that ensures the Min/Max functions return a null value (instead of panicing)
```
asubiotto, line 40: nit: For consistency, I think we should avoid block comments and only use multi-line `//` comments.

asubiotto (commented):
Nice catch. Just added some observations.

Issue #361, opened 2023-02-28
Record schema changes in WAL


thorfour:
https://github.com/polarsignals/frostdb/blob/e98d6bfeb8bf59dbb593d2ac663058bb4d1fb8ab/db.go#L638

The Table function accepts a table config with a new schema, but never applies the schema to the table.  This can cause a panic during wal replays where the wal will replay the writes for the new schema under the old schema because the schema change was never recorded.

We need to 
- check if the passed in schema was different than what the table has, if it is, update the schema to the new schema.
- Record in the WAL a schema change event.
- During replay if schema change event detected, update the table schema with saved schema

asubiotto:
Fixed by #383

PR #363, opened 2023-03-02
unit test: read only table queries

thorfour:
Performs an aggregation read on a read only table. Aggregations require a schema, and this causes a panic as read only tables do not have a schema set.

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -647,6 +647,10 @@ func (a *CountAggregation) Aggregate(pool memory.Allocator, arrs []arrow.Array) 
 // the set of values. It is aware of the final stage and chooses the aggregation
 // function appropriately.
 func runAggregation(finalStage bool, fn logicalplan.AggFunc, pool memory.Allocator, arrs []arrow.Array) (arrow.Array, error) {
+	if len(arrs) == 0 {
+		return array.NewInt64Builder(pool).NewArray(), nil
+	}
+
 	aggFunc, err := chooseAggregationFunction(fn, arrs[0].DataType())
 	if err != nil {
 		return nil, err
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -13,7 +13,6 @@ import (
 	"github.com/apache/arrow/go/v10/arrow/memory"
 	"github.com/apache/arrow/go/v10/arrow/scalar"
 	"github.com/dgryski/go-metro"
-	"github.com/segmentio/parquet-go"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/polarsignals/frostdb/pqarrow/builder"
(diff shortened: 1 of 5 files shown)


thorfour:
I believe this test would be fixed if we address #361 because then the WAL replay will set the schema for the table.

However this is likely still a problem the db if running without a WAL. May be worth considering dropping the schema.json in the tables dir?

thorfour:
Deferred the decision to decide on aggregation function into the `runAggregation` function where it will select based on the `arrow.Array` data type that it's passed. This removes the need for a table schema to perform aggregation functions.

query/physicalplan/aggregate.go
```diff
@@ -658,16 +646,20 @@ func (a *CountAggregation) Aggregate(pool memory.Allocator, arrs []arrow.Array)
...
-	pool memory.Allocator,
-	arrs []arrow.Array,
-) (arrow.Array, error) {
-	if _, ok := fn.(*CountAggregation); ok && finalStage {
+func runAggregation(finalStage bool, fn logicalplan.AggFunc, pool memory.Allocator, arrs []arrow.Array) (arrow.Array, error) {
+	if len(arrs) == 0 {
+		return array.NewInt64Builder(pool).NewArray(), nil
+	}
+
+	aggFunc, err := chooseAggregationFunction(fn, arrs[0].DataType())
```
asubiotto, line 654: Could you run the aggregation benchmarks? I'm wondering if doing this logical to physical agg func conversion each time we run an aggregation is detrimental to performance.

asubiotto:
Rebased to pick up lint version bump

query/physicalplan/aggregate.go
thorfour:
```
thor@thors-MacBook-Pro frostdb % benchstat before.txt after.txt
Aggregation/sum-10      1.45ms ± 1%    1.43ms ± 1%  -0.73%  (p=0.003 n=9+9)
Aggregation/count-10    1.45ms ± 1%    1.43ms ± 1%  -1.36%  (p=0.000 n=10+9)
Aggregation/max-10      1.45ms ± 1%    1.44ms ± 1%  -0.99%  (p=0.003 n=10+10)

Aggregation/sum-10      1.40MB ± 0%    1.39MB ± 0%  -0.82%  (p=0.000 n=7+10)
Aggregation/count-10    1.40MB ± 0%    1.39MB ± 0%  -0.82%  (p=0.000 n=10+10)
Aggregation/max-10      1.40MB ± 0%    1.39MB ± 0%  -0.82%  (p=0.000 n=10+10)

Aggregation/sum-10       21.0k ± 0%     20.9k ± 0%  -0.69%  (p=0.000 n=10+10)
Aggregation/count-10     21.0k ± 0%     20.9k ± 0%  -0.69%  (p=0.000 n=10+10)
Aggregation/max-10       21.0k ± 0%     20.9k ± 0%  -0.69%  (p=0.000 n=10+10)
```

thorfour: Looks like it's negligible

metalmatze (approved):
Good you added the benchmark in the comments.

PR #364, opened 2023-03-06
Serialize less memory

thorfour:
This updates the `Serialize` function 

Previously we were collecting all row groups in the entire table block, and using the parquet merge function to merge all the row groups into a single row group. This causes a large memory spike because we would load at least a page for every row group (or maybe 2 if we're async reading the pages) into memory to perform the heap sort of all the row groups.

Instead of doing that; we can leverage the fact that Granules are already sorted in relation to each other and instead merge each Granules row groups into a single one, and write that directly to the final file. The benchmark shows some considerable gains in memory, and speed.

```
thor@thors-MacBook-Pro frostdb % benchstat before.txt after.txt
_Serialize-10    3.38ms ± 3%    1.42ms ± 0%  -58.13%  (p=0.008 n=5+5)

_Serialize-10    3.94MB ± 0%    0.13MB ± 0%  -96.59%  (p=0.008 n=5+5)

_Serialize-10     40.2k ± 0%      1.2k ± 0%  -97.04%  (p=0.008 n=5+5)
```

### DIFF
--- a/table.go
+++ b/table.go
@@ -1339,63 +1339,121 @@ func (t *TableBlock) abortCompaction(granule *Granule) {
 	}
 }
 
+func mergeDynCols(a, b map[string][]string) map[string][]string {
+	for key, vals := range b {
+		a[key] = append(a[key], vals...)
+	}
+
+	return bufutils.Dedupe(a)
+}
+
+/*
+	Serialize the table block into a single Parquet file.
+	The Serialize function will walk all Granules in the block, compact each Granule into a sorted set of Row groups, then write those
+	row groups to the final Parquet file, repeating for each Granule. This leverages the fact that the index has alreayd sorted the Granules
+	so no need to sort them further than the intra-granule parts.
+*/
 func (t *TableBlock) Serialize(writer io.Writer) error {
-	ctx := context.Background()
+	dynCols := map[string][]string{} // NOTE: it would be nice if we didn't have to go back and find all the dynamic columns...
+	t.Index().Ascend(func(i btree.Item) bool {
+		g := i.(*Granule)
 
-	// Read all row groups
-	rowGroups := []dynparquet.DynamicRowGroup{}
-
-	rowGroupsChan := make(chan any)
-
-	errg := &errgroup.Group{}
-	errg.Go(func() error {
-		for rg := range rowGroupsChan {
-			switch p := rg.(type) {
-			case arrow.Record:
-				b := &bytes.Buffer{}
-				w, err := t.table.config.schema.GetWriter(b, pqarrow.RecordDynamicCols(p))
-				if err != nil {
-					return err
-				}
-
-				if err := pqarrow.RecordToFile(t.table.config.schema, w.ParquetWriter(), p); err != nil {
-					return err
-				}
-
-				f, err := parquet.OpenFile(bytes.NewReader(b.Bytes()), int64(b.Len()))
-				if err != nil {
-					return err
-				}
-
-				buf, err := dynparquet.NewSerializedBuffer(f)
-				if err != nil {
-					return err
-				}
-
-				rowGroups = append(rowGroups, buf.MultiDynamicRowGroup())
-
-			case dynparquet.DynamicRowGroup:
-				rowGroups = append(rowGroups, p)
-			default:
-				return fmt.Errorf("unknown part type: %T", p)
+		g.PartsForTx(math.MaxUint64, func(p *parts.Part) bool {
+			buf, err := p.AsSerializedBuffer(t.table.config.schema)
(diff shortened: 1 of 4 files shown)


table.go
```diff
@@ -1339,63 +1339,121 @@ func (t *TableBlock) abortCompaction(granule *Granule) {
...
+	}
 
-	// Read all row groups
-	rowGroups := []dynparquet.DynamicRowGroup{}
+	return bufutils.Dedupe(a)
+}
 
-	rowGroupsChan := make(chan any)
+/*
+	Serialize the table block into a single Parquet file.
```
asubiotto, line 1351: nit: use multi-line comments `//` instead of block comments

table.go
```diff
@@ -1339,63 +1339,121 @@ func (t *TableBlock) abortCompaction(granule *Granule) {
...
+	The Serialize function will walk all Granules in the block, compact each Granule into a sorted set of Row groups, then write those
+	row groups to the final Parquet file, repeating for each Granule. This leverages the fact that the index has alreayd sorted the Granules
+	so no need to sort them further than the intra-granule parts.
+*/
+func (t *TableBlock) Serialize(writer io.Writer) error {
+	dynCols := map[string][]string{} // NOTE: it would be nice if we didn't have to go back and find all the dynamic columns...
```
asubiotto, line 1357: Agreed! At some point, these dynamic columns should be a stateful part of the schema. Would make execution code simpler too. Created an issue: https://github.com/polarsignals/frostdb/issues/366

table.go
```diff
@@ -1339,63 +1339,121 @@ func (t *TableBlock) abortCompaction(granule *Granule) {
...
-			case arrow.Record:
-				b := &bytes.Buffer{}
-				w, err := t.table.config.schema.GetWriter(b, pqarrow.RecordDynamicCols(p))
-				if err != nil {
-					return err
-				}
+		g.PartsForTx(math.MaxUint64, func(p *parts.Part) bool {
+			buf, err := p.AsSerializedBuffer(t.table.config.schema)
+			if err != nil {
+				panic("fuck")
```
asubiotto, line 1364: Probably leftover but let's find a way to return this error from `Serialize` with maybe a different wording 😂

table.go
```diff
@@ -1339,63 +1339,121 @@ func (t *TableBlock) abortCompaction(granule *Granule) {
...
+			return false
+		}
+		rows := merged.Rows()
+		defer rows.Close()
 
-	close(rowGroupsChan)
-	if err := errg.Wait(); err != nil {
-		return err
+		// Write the merged row groups to the writer
+		for i, n := 0, 0; i < int(merged.NumRows()); i += n {
```
asubiotto, line 1425: Can you reuse the `writeRows` method here?

asubiotto (commented):
Nice, we could probably also check each part's compaction level and avoid merging those as well since level 1 parts are non-overlapping with each other. This makes me think, should we maybe force a compaction of the table block to have only level 1 parts and then serialize those or would that be wasting CPU?

table.go
thorfour: Awesome thank you!

table.go
thorfour: I think this will be a larger rewrite since that's meant to be a one-shot function. But I'll look into what it would take to reuse all that code

asubiotto: Definitely let me know if you run into any obstacles. Compaction uses `writeRows` and the only thing I can see that is different is that you don't want to close the writer after writing the merged row group. Maybe you can extract the non-writer logic in `writeRows` to force the caller provide a writer and have the normal path close the writer after one call, while this path uses the same writer.

thorfour: Yea, but it also needs to keep state around rows written for flushing etc. So it needs to be able to handle that state

thorfour: Refactored into a stateful row writer. Let me know what you think

thorfour:
> Nice, we could probably also check each part's compaction level and avoid merging those as well since level 1 parts are non-overlapping with each other. This makes me think, should we maybe force a compaction of the table block to have only level 1 parts and then serialize those or would that be wasting CPU?

They're non-overlapping but are they ordered? I think I'll open up an issue around this improvement.

asubiotto:
> > Nice, we could probably also check each part's compaction level and avoid merging those as well since level 1 parts are non-overlapping with each other. This makes me think, should we maybe force a compaction of the table block to have only level 1 parts and then serialize those or would that be wasting CPU?
> 
> They're non-overlapping but are they ordered? I think I'll open up an issue around this improvement.

No, but you can sort them with the `PartSorter`.

asubiotto (approved):

PR #369, opened 2023-03-06
Prefix metrics with `frostdb_`

metalmatze:
Maybe some day we could use composable registries to not have to explicitly prefix all metrics now. I've opened for this PR and the Parca one: https://github.com/prometheus/client_golang/issues/1229


metalmatze:
`WrapRegistererWithPrefix` actually works as intended. I just added the wrapping to the registry that would later be overwritten...

thorfour (approved):
Forst lol

metalmatze:
> Forst lol

I hereby announce the name change to ForstDB. :deciduous_tree: :evergreen_tree: 
_(Forst is forest in German)_

Thankfully it's just in the title.

metalmatze:
Giving this some more thought since yesterday, I think explicitly prefixing the metric names explicitly with `frostdb_` makes sense here.

brancz (approved):

brancz:
just tripped over not having these again in prod, so would love to continue this

metalmatze:
Merge main back into this branch and if it goes green we should just merge it.

PR #371, opened 2023-03-07
frostdb: enable snapshots

### LINKED ISSUE #345: frostdb/wal: create snapshots
Currently when restarting a database with an associated WAL, all records are read one by one and replayed to achieve a final database state. We have seen that this process can be inefficient, prolonging startup time which makes the database unusable until the final state.

This issue proposes to introduce the idea of snapshots to accelerate WAL replay times. These snapshots will be taken periodically and stored alongside the `wal` directory. Each snapshot should have an associated txn that it represents so that the replayer can load the latest snapshot and only replay wal records from the associated txn.

The meat of the work here will be to figure out how to serialize a snapshot given that we ideally want to differentiate between l0/l1 parts (to avoid re-compaction) and might want to serialize other data.

asubiotto:
This commit adds the option to enable snapshotting a database whenever a block exceeds a given size or is rotated out.
A snapshot is taken of the db state and stored in the snapshots/ folder. The use of the WAL is not modified yet, even though it could be truncated. This will be done in a future change once snapshots have been running for a while.

Note that old snapshots are not cleaned up yet, which gives us more txns to reload at in case of issues. In the future, a cleanup policy will be implemented.

Closes #345


asubiotto:
Note that we could get into a state where we perform two snapshots close to each other in case of two large tables, but I think this is acceptable for now even though we could improve this to have a snapshot goroutine that routinely checks database size.

table.go
```diff
@@ -663,22 +684,55 @@ func (t *Table) appender() (*TableBlock, func(), error) {
 			return nil, nil, err
 		}
 
-		if block.Size() < t.db.columnStore.activeMemorySize {
+		blockSize := block.Size()
+		if t.db.columnStore.snapshotTriggerSize != 0 &&
+			blockSize-block.lastSnapshotSize.Load() > t.db.columnStore.snapshotTriggerSize {
```
metalmatze, line 689: The only place where I'd like some more context in form of a comment.
Why `blockSize - block.lastSnapshotSize`?

metalmatze: Thanks for the comment. That definitely helps. I'm still wondering:
Why does the `block.lastSnapshotSize` need to be subtracted? Shouldn't `blockSize > t.db.columnStore.snapshotTriggerSize` be enough here? Why does the last snapshot size matter?

asubiotto: Because we want to get the number of "new" bytes since the last snapshot. Otherwise we'll be snapshotting every new write once the block size exceeds the trigger size (e.g. if the trigger size is `10`, we'll snapshot at `blockSize=10,11,12,13,14 etc...` when we only want to snapshot at `blockSize=10,20,30,40 etc...`)

db.go
```diff
@@ -573,7 +596,7 @@ func (db *DB) Close() error {
 	for _, table := range db.tables {
 		table.close()
 		if db.bucket != nil {
-			table.writeBlock(table.ActiveBlock())
+			table.writeBlock(context.TODO(), table.ActiveBlock())
```
thorfour, line 599: Should probably update to a real ctx that's threaded through

thorfour (approved):
I'm unclear why we snapshot the new/empty block after a rotation?

db.go
asubiotto: This is a heavy lift because you have to pass it in through `columnStore.Close` and `db.Close` which seemed like a heavy refactor/out of scope for this PR. This is why I used `context.TODO`, to indicate that at some point in the future, we probably need to pass it an actual ctx.

asubiotto:
My thinking was that the snapshots are going to be pretty cheap since there's a lot less data, and it serves as an additional checkpoint to replay WAL entries from, but I'm happy to remove.

PR #375, opened 2023-03-08
Replay wal on open

### LINKED ISSUE #328: wal: make replaying part of opening the WAL
Currently, callers need to explicitly call `ReplayWAL` to replay a write ahead log. This should be part of opening a WAL. Given the caller needs to provide a replay function, this can be passed in to `wal.Open` by the db. If no replay function is provided, I suggest we just nuke all the files in the WAL.

thorfour:
In efforts to make the WAL/replay more robust I've added two things here:

- On wal read panics instead of continuing the panic we instead truncate the WAL up to the last good tx. Using some corrupted wals from production this works really well.
- On DB open (and wal open) replay the WAL. This is mostly a refactor, but it does remove some weird footguns where you could open a DB with a WAL and not replay it, or write to it while you were replaying it.

Closes #328

### DIFF
--- a/db.go
+++ b/db.go
@@ -259,11 +259,11 @@ func (s *ColumnStore) ReplayWALs(ctx context.Context) error {
 	for _, f := range files {
 		databaseName := f.Name()
 		g.Go(func() error {
-			db, err := s.DB(ctx, databaseName)
+			_, err := s.DB(ctx, databaseName)
 			if err != nil {
 				return err
 			}
-			return db.recover(ctx)
+			return nil
 		})
 	}
 
@@ -350,14 +350,6 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
 		db.bucket = storage.NewPrefixedBucket(s.bucket, db.name)
 	}
 
-	if s.enableWAL {
-		var err error
-		db.wal, err = db.openWAL()
-		if err != nil {
-			return nil, err
-		}
-	}
-
 	if dbSetupErr := func() error {
 		db.txPool = NewTxPool(db.highWatermark)
 		db.compactorPool = newCompactorPool(db, s.compactionConfig)
@@ -372,6 +364,20 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
 				return nil
 			}); err != nil {
 				return fmt.Errorf("bucket iter on database open: %w", err)
+			}
+		}
+
+		if s.enableWAL {
+			snapshotTx, err := db.loadLatestSnapshot(ctx)
+			if err != nil {
+				level.Debug(s.logger).Log(
+					"msg", "failed to load latest snapshot", "db", db.name, "err", err,
+				)
+				snapshotTx = 0
+			}
+			db.wal, err = db.openWAL(ctx, snapshotTx)
+			if err != nil {
+				return err
 			}
 		}
 
@@ -404,12 +410,26 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
 	return db, nil
 }
 
-func (db *DB) openWAL() (WAL, error) {
-	return wal.Open(
+func (db *DB) openWAL(ctx context.Context, firstIndex uint64) (WAL, error) {
+	// persistedTables is a map from a table name to the last transaction
+	// persisted.
+	persistedTables := map[string]uint64{}
+	lastTx := uint64(0)
+	wal, err := wal.Open(
 		db.logger,
 		db.reg,
 		db.walDir(),
+		wal.WithFirstIndex(firstIndex),
+		wal.WithReplayFunc(2, db.recover(ctx, persistedTables, &lastTx)),
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	db.tx.Store(lastTx)
+	db.highWatermark.Store(lastTx)
+
+	return wal, nil
 }
 
 func (db *DB) walDir() string {
@@ -422,7 +442,7 @@ func (db *DB) snapshotsDir() string {
 
 // recover attempts to recover database state from a combination of snapshots
 // and the WAL.
-func (db *DB) recover(ctx context.Context) error {
+func (db *DB) recover(ctx context.Context, persistedTables map[string]uint64, lastTx *uint64) wal.ReplayHandlerFunc {
 	level.Info(db.logger).Log("msg", "recovering db")
 	snapshotLoadStart := time.Now()
 	snapshotTx, err := db.loadLatestSnapshot(ctx)
@@ -456,7 +476,9 @@ func (db *DB) recover(ctx context.Context) error {
 
 	firstIndex, err := db.wal.FirstIndex()
 	if err != nil {
-		return err
+		level.Info(db.logger).Log(
+			"msg", "failed to get WAL first index",
+			"err", err)
 	}
 
 	if snapshotTx > firstIndex {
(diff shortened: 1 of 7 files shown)


db.go
```diff
@@ -398,198 +397,196 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
...
-					return err
+	wal, err := wal.Open(
+		db.logger,
+		db.reg,
+		db.walDir(),
+		wal.WithFirstIndex(firstIndex),
+		// NOTE: 2 is passed here because we require 2 passes through the WAL.
+		// Once to establish all the storage blocks that have rotated out, and a second pass to replay all the writes.
+		// This prevents us from replaying writes that have already been persisted/dropped by a block rotation.
+		wal.WithReplayFunc(2, func(i int, tx uint64, record *walpb.Record) error {
```
asubiotto, line 412: I find this interface a bit awkward. I think it should be enough for our purposes to call `wal.Open` first and then `wal.Replay` below.

asubiotto:
I rebased this on the latest main (had to skip the commit making the replay a function closure)

db.go
```diff
@@ -477,190 +502,177 @@ func (db *DB) recover(ctx context.Context) error {
...
 		default:
-			return fmt.Errorf("unexpected WAL entry type: %t", e)
+			return fmt.Errorf("unexpected number of replay counts; expected 2, got %v", i)
 		}
-		return nil
-	}); err != nil {
-		return fmt.Errorf("second WAL replay: %w", err)
 	}
-
-	level.Info(db.logger).Log("msg", "replaying WAL completed", "duration", time.Since(start))
```
asubiotto, line 658: These log messages have been pretty useful (I think they're not in the latest version of this PR)

db.go
```diff
@@ -258,18 +258,11 @@ func (s *ColumnStore) ReplayWALs(ctx context.Context) error {
 	for _, f := range files {
 		databaseName := f.Name()
 		g.Go(func() error {
-			db, err := s.DB(ctx, databaseName)
+			_, err := s.DB(ctx, databaseName)
```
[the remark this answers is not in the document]
thorfour: I left it in because it allows one to replay all the wals before the columnstore is considered ready. Otherwise whenever a write occurs on a db that hadn't been opened yet it would block that write to perform the replay.

db.go
thorfour: Yes! Thank you, I meant to include that again but forgot I had deleted it

thorfour: added back in

db.go
```diff
@@ -404,12 +410,31 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
...
 
-func (db *DB) openWAL() (WAL, error) {
-	return wal.Open(
+func (db *DB) openWAL(ctx context.Context, firstIndex uint64) (WAL, error) {
+	// persistedTables is a map from a table name to the last transaction
+	// persisted.
+	start := time.Now()
+	persistedTables := map[string]uint64{}
+	lastTx := uint64(0)
+	wal, err := wal.Open(
```
asubiotto, line 419: I think this comment got lost: I find this interface pretty awkward. Thinking about it more. I think we should make `Run` a separate method so this method will look like:
```
wal, err := wal.Open
// recover, aka call Replay
db.recover()
// And finally actually start the WAL.
wal.RunAsync()
```

thorfour: hmm I dislike that API because it's brittle. You have a WAL that you can open that hasn't been replayed which seems error prone

thorfour: Also apologies for it getting lost, I thought it was resolved with your refactor into using the recover function. I had misunderstood

asubiotto: No problem.

I see your point, but I feel like replaying is defined by the caller. It should be valid to `Open`, and do some operations on the wal (e.g. `Truncate`) before a `Replay`. `Replay`ing is conceptually loading wal bytes that only the caller knows how to interpret.

Additionally, this is basically the same api that `tidwall/wal` offers us. You `Open` a wal to load the state and it's ready for any action (from `Truncate`ing to `Replay`ing).

The difference with `Replay`ing at the `frostdb` user level is that they don't need to know about what is loaded and not loaded in the database and it is indeed an error to perform any operations against a database with an unloaded WAL.

db.go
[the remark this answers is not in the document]
asubiotto: I'm not sure I understand. Shouldn't it be sufficient to move this code to `ColumnStore.New`? The columnstore can't be considered ready until that returns.

db.go
thorfour: That's a good point, I'll refactor to that.

thorfour: Done

db.go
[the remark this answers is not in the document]
thorfour: Yea that's a good point, I'll move it there

db.go
```diff
@@ -236,7 +240,7 @@ func (s *ColumnStore) DatabasesDir() string {
 }
 
 // ReplayWALs replays the write-ahead log for each database.
```
asubiotto, line 242: nit: I should've probably renamed this to something like `recoverDBsFromStorage` since it's not just the WAL anymore.

db.go
```diff
@@ -259,11 +263,11 @@ func (s *ColumnStore) ReplayWALs(ctx context.Context) error {
 	for _, f := range files {
 		databaseName := f.Name()
 		g.Go(func() error {
-			db, err := s.DB(ctx, databaseName)
+			_, err := s.DB(ctx, databaseName)
```
asubiotto, line 266: nit: add a small comment explaining the side effects of this `DB` call. I don't think it'll be clear to a casual reader.

db.go
```diff
@@ -259,11 +263,11 @@ func (s *ColumnStore) ReplayWALs(ctx context.Context) error {
...
 			if err != nil {
 				return err
 			}
-			return db.recover(ctx)
+			return nil
```
asubiotto, line 270: `return err` directly (no need for the `if != nil` check above)

table.go
```diff
@@ -142,11 +143,7 @@ type WAL interface {
 	Close() error
 	Log(tx uint64, record *walpb.Record) error
 	LogRecord(tx uint64, table string, record arrow.Record) error
-	// Replay replays WAL records from the given first index. If firstIndex is
-	// 0, the first index read from the WAL is used (i.e. given a truncation,
-	// using 0 is still valid). If the given firstIndex is less than the WAL's
-	// first index on disk, the replay happens from the first index on disk.
-	Replay(firstIndex uint64, handler func(tx uint64, record *walpb.Record) error) error
+	Replay(tx uint64, handler wal.ReplayHandlerFunc) error
```
asubiotto, line 146: nit: keep the comment

wal/wal.go
```diff
@@ -408,15 +410,21 @@ func (w *FileWAL) Replay(firstIndex uint64, handler func(tx uint64, record *walp
...
+	tx := firstIndex
 	defer func() {
-		// recover the panic to print more context. Exit afterwards regardless.
-		if err := recover(); err != nil {
-			level.Error(w.logger).Log("msg", "replaying WAL failed", "path", w.path, "first_index", firstIndex, "last_index", lastIndex, "err", err)
-			panic(err)
+		// recover a panic of reading a transaction. Truncate the wal to the last valid transaction.
+		if r := recover(); r != nil {
+			level.Error(w.logger).Log("msg", "replaying WAL failed", "path", w.path, "first_index", firstIndex, "last_index", lastIndex, "offending_index", tx, "err", err)
+			if err = w.log.TruncateBack(tx - 1); err != nil {
```
asubiotto, line 418: Hmm, I understand the reasoning, but it feels very aggressive to `TruncateBack`. Don't think you should change anything, just a note. Should we have a metric for this?

wal/wal.go
```diff
@@ -408,15 +410,21 @@ func (w *FileWAL) Replay(firstIndex uint64, handler func(tx uint64, record *walp
 
 	level.Debug(w.logger).Log("msg", "replaying WAL", "first_index", firstIndex, "last_index", lastIndex)
 
+	tx := firstIndex
```
asubiotto, line 413: nit: might as well rename `firstIndex` to `tx`

asubiotto (approved):
Fantastic, thank you. Just a couple of nits

wal/wal.go
thorfour: The only thing we're missing out is the ability to manually copy out the WAL for debug purposes. We can track the number of times this occurs with the logging system.

PR #383, opened 2023-03-09
381 wal persist table config

### LINKED ISSUE #381: Table config not persisted in the WAL
The WAL replay will reload a schema in a replayed table but will not reload any table settings that were previously configured.

This can lead to scenario where blocks are being written with a certain row group size, but after a replay, the row group size limit is reset to 0 (unlimited) and blocks start to be written with singular row groups.

thorfour:
Closes #381

This adds the table config to the wal proto message to ensure that when the WAL is replay a table block creation message it copies in the table config as well so settings like `RowGroupSize` don't go missing.

The bulk of the changes in here are moving the table config proto definition into it's own file and having both the snapshot and wal proto's import it instead.


db_test.go
```diff
@@ -403,18 +403,20 @@ func Test_DB_ColdStart(t *testing.T) {
...
-			newColumnstore: func(t *testing.T) *ColumnStore {
-				c, err := New(
-					WithLogger(logger),
-					WithBucketStorage(bucket),
-					WithWAL(),
-					WithStoragePath(t.TempDir()),
-				)
-				require.NoError(t, err)
-				return c
+		/*
```
brancz, line 406: Anything you wanted to finish here?

brancz (approved):
Looks good other than the comment. Nice find!

proto/frostdb/snapshot/v1alpha1/snapshot.proto
```diff
@@ -13,16 +12,7 @@ message FooterData {
...
-    oneof schema {
-      frostdb.schema.v1alpha1.Schema deprecated_schema = 1;
-      frostdb.schema.v1alpha2.Schema schema_v2 = 2;
-    }
-    int64 row_group_size = 3;
-    int64 block_reader_limit = 4;
-    bool disable_wal = 5;
-  }
-  TableConfig config = 2;
+  frostdb.table.v1alpha1.TableConfig config = 5;
```
asubiotto, line 15: I believes this can stay as field number 2 (we don't care about backwards compatibility yet)

asubiotto (approved):
LGTM minus the proto comments

proto/frostdb/snapshot/v1alpha1/snapshot.proto
thorfour: That's fair. I tried to "do the right thing" but you're probably right.

db_test.go
thorfour: lol forgot that when fixing.

PR #393, opened 2023-03-14
*: miscellaneous recovery fixes

### LINKED ISSUE #390: wal: ensure wal can be appended to after loading snapshots
I believe that if we load a snapshot at txn `n`, but we have a truncated wal with the last index at txn `n-k`, we could get into a situation where no new records are appended to the WAL, which is not good since the log request queue could grow unboundedly given the wal's nextTx is at `n-k`, but this transaction will never be appended to the log.

I think this is currently happening in production given that the WALs are tiny.

asubiotto:
See commits for details

Closes #390

This PR fixes the referenced issue because when loading from a snapshot at txn n and a wal at txn n+k, the tables would be created with a noop wal that was never updated to point to the db wal. This means that no more records will ever be logged to the WAL.

The issue that is mentioned in 390 that somehow we get into a state with a WAL that has a lastIndex < highWatermark is theoretically impossible. WALs will only ever be replayed from a snapshot at txn n, and TruncateBack will therefore only ever be called on txn n+i, but n+i >= n is always true, so we will never load a snapshot to update our last index past the end of the WAL (unless we manually delete files).


asubiotto:
I think I'd still like to add a test for this.

db.go
```diff
@@ -448,11 +451,34 @@ func (db *DB) recover(ctx context.Context, wal WAL) error {
...
+			"msg", "failed to get WAL first index",
+			"err", err)
+	}
+	lastIndex, err := wal.LastIndex()
+	if err != nil {
+		level.Info(db.logger).Log(
+			"msg", "failed to get WAL last index",
+			"err", err)
+	}
+	snapshotLogArgs := make([]interface{}, 0)
```
metalmatze, line 466: Isn't this `any` over `interface` now?
```suggestion
	snapshotLogArgs := make([]any{}, 0)
```

db_test.go
```diff
@@ -1523,6 +1522,28 @@ func TestDBRecover(t *testing.T) {
...
+	for i := range samples {
+		samples[i].Timestamp = numInserts
+	}
+	buf, err := samples.ToBuffer(table.schema)
+	require.NoError(t, err)
+
+	writeTx, err := table.InsertBuffer(ctx, buf)
+	require.NoError(t, err)
+
+	require.Eventually(t, func() bool {
```
metalmatze, line 1542: TIL `require.Eventually`. Cool!

metalmatze (approved):
Very nice!

db.go
asubiotto: Fair point, will change

PR #395, opened 2023-03-15
*: fix not deleting snapshot dir on db.Close

asubiotto:
This commit deletes the snapshot dir alongside the wal dir if blocks are successfully persisted on Close. Not doing so could cause a recovery from this state to never log to the WAL since the recovered db tx from the snapshot could be much higher than the new WAL's last index.

Additionally, this commit enhances recovery testing with other scenarios.


metalmatze (approved):
:+1:

PR #396, opened 2023-03-15
Split large aggregations into multipl records

thorfour:
This adds a hedge into the binary opt builder to check if an append will cause the offsets to overflow an int32 (which is what the arrow arrays use for offsets) if so it returns an error indicating the max array size has been reached.

This updates the hash aggregator to check if this max array size error has been returned during aggregation; if it has, it will rollback the row appends and start a new set of arrays to aggregate onto. At the end it will create multiple records out of these aggregates and send them as the callback.

This does have a knock on effect of Hyrum's law, where previously a user may have correctly assumed that an aggregate would only every call the callback function once. However this is definitely no longer true, for example in [Parca](https://github.com/parca-dev/parca/blob/c1f03a1fc5b8bd9831aa5620d1bfeca9334be7c4/pkg/parcacol/querier.go#L347-L350) we currently have this problem. 

So before clients of this code update they should take care to ensure they are correctly handling aggregate callbacks. That said; it's unlikely to hit because it requires a buffer overflow which means one would be querying lots of data.


query/physicalplan/aggregate.go
```diff
@@ -443,62 +451,116 @@ func (a *HashAggregate) Callback(ctx context.Context, r arrow.Record) error {
...
 
-	groupByFields := make([]arrow.Field, 0, numCols)
-	groupByArrays := make([]arrow.Array, 0, numCols)
-	for _, fieldName := range a.colOrdering {
-		groupByCol, ok := a.groupByCols[fieldName]
-		if !ok {
-			return fmt.Errorf("unknown field name: %s", fieldName)
+		groupByCol, found := aggregate.groupByCols[fieldName]
+		if !found {
+			groupByCol = builder.NewBuilder(a.pool, groupByFields[i].Type)
```
asubiotto, line 463: This is where I would say we should create a new opt large binary builder (or even just default to arrow's large binary builder to start with). It's essentially just the hash aggregate result that we care about handling this case.

asubiotto (commented):
Could you run the benchmarks to get an idea if this impacts query perf given the changes to the opt binary builder?

I also wonder if you considered just creating a different type of optbuilder that uses int64s (similar to how arrow gives you the option to build `LargeBinary` arrays). I think that this is my preferred solution instead of having to worry about overflow in the execution code.

thorfour:
```
Aggregation/sum-10      1.29ms ± 1%    1.31ms ± 0%  +1.33%  (p=0.000 n=10+10)
Aggregation/count-10    1.29ms ± 0%    1.31ms ± 0%  +1.17%  (p=0.000 n=10+9)
Aggregation/max-10      1.30ms ± 0%    1.31ms ± 0%  +1.05%  (p=0.000 n=10+8)

Aggregation/sum-10      1.39MB ± 0%    1.39MB ± 0%  +0.04%  (p=0.000 n=9+10)
Aggregation/count-10    1.39MB ± 0%    1.39MB ± 0%  +0.04%  (p=0.000 n=10+9)
Aggregation/max-10      1.39MB ± 0%    1.39MB ± 0%  +0.03%  (p=0.000 n=10+9)

Aggregation/sum-10       20.9k ± 0%     20.9k ± 0%  +0.11%  (p=0.000 n=10+10)
Aggregation/count-10     20.9k ± 0%     20.9k ± 0%  +0.11%  (p=0.000 n=10+10)
Aggregation/max-10       20.9k ± 0%     20.9k ± 0%  +0.11%  (p=0.000 n=8+9)
```

Here are the current benchmarks. Currently looking into bench marking large aggregations to see if it's going to make a difference to switch to the large binary arrays

thorfour:
```bash
thor@thors-MacBook-Pro physicalplan % benchstat largebinary_overflow.txt splitlarge_overflow.txt
name                         old time/op    new time/op    delta
_Aggregate_ArrayOverflow-10      102s ± 5%       63s ± 1%  -38.56%  (p=0.000 n=9+7)

name                         old alloc/op   new alloc/op   delta
_Aggregate_ArrayOverflow-10     245GB ± 0%     206GB ± 0%  -16.06%  (p=0.000 n=10+10)

name                         old allocs/op  new allocs/op  delta
_Aggregate_ArrayOverflow-10     3.23M ± 0%     3.23M ± 0%   +0.11%  (p=0.000 n=10+10)
```

Wrote a benchmark where it aggregates a ton of data to see what the difference between using large binary vs. splitting

| implementation | memory | cpu |
|----|---- |-----|
| large binary array | https://pprof.me/b73bdb6/ | https://pprof.me/a397fdf/ |
| split binary array | https://pprof.me/ab7d3d0/ | https://pprof.me/ae474c6/ |

I'm a little surprised by these results, I thought there would be only a small amount of memory overhead with the large binary arrays @asubiotto wdyt?

asubiotto:
Thanks for running the experiments!

I do expect more memory used since we're adding 4 bytes per value, but probably not as large as you're seeing. BTW, I think there is still a difference in your second benchmark, just that benchstat probably determined there were not enough samples to display a statistically significant diff.

Here's one last suggestion:
In the Cockroach vectorized engine (and other columnar databases), there's a notion of a global "batch" size (https://github.com/cockroachdb/cockroach/blob/8b9c7dd37f2caafabe15167ef8410620415ee071/pkg/col/coldata/batch.go#L84) for execution that operators try to respect. I think we eventually want to have something like this so that buffering operators (like the hash agg) start emitting data early to reduce memory usage and unblock downstream operators. Maybe we should introduce a batch size and have the hash aggregate respect it (eventually moving all the other buffering operators to do so), so the split would happen a lot earlier than when we reach overflow, which also makes this behavior easier to test. The downside is that we definitely need to change the consumers of this API.

thorfour:
Yea you're definitely right about not running enough samples. The time to run these benchmarks is long so I definitely shorted it on that trial.

I like the idea of batch sizes and considered it initially, but it had two draw backs I could determine. 

1. If we emit data early in the hash aggregate won't that mean we could potentially not be performing the full aggregation? Effectively kicking the can down the road. 
2. It still doesn't solve the overflow problem, even if the batch size is small a binary column could use large enough values that it overflows the array without exceeding the batch size.

asubiotto:
> Yea you're definitely right about not running enough samples. The time to run these benchmarks is long so I definitely shorted it on that trial.
> 
> I like the idea of batch sizes and considered it initially, but it had two draw backs I could determine.
> 
> 1. If we emit data early in the hash aggregate won't that mean we could potentially not be performing the full aggregation? Effectively kicking the can down the road.

Yes, the "emitting data early" part was more for other operators (e.g. ordered agg). In the hash agg's case it would still buffer all results (but a separate record per group, so it's not as big a problem as using a single record for all results), and split in the same path you do (i.e. in the `Finish` call before returning all results).

> 2. It still doesn't solve the overflow problem, even if the batch size is small a binary column could use large enough values that it overflows the array without exceeding the batch size.

It could, but it's a lot less likely. It depends how big we think a stacktrace will be. I've only seen 4KiB stacktraces max in prod, and with an 8192 batch size for example, we'd need every stacktrace to be around 250KiB.

PR #397, opened 2023-03-16
*: call closeInternal on db startup error rather than Close

asubiotto:
Close expects some fields to be non-nil (e.g. metrics when triggering a snapshot on close), so this commit introduces a lighter-weight closeInternal that is called on a db setup error, which is all that is needed. Not doing so was causing a nil pointer exception.


metalmatze (approved):
LGTM, linting fails though.

Issue #366, opened 2023-03-06
frostdb: keep track of concrete dynamic columns in schema


asubiotto:
In FrostDB, we allow a user to have a dynamic columns in the schema. This means that they can add concrete columns to the schema at runtime. This leads to a lot of places in our codebase where we have to compute the concrete dynamic columns (e.g. comparisons, execution). Instead of computing the added columns at runtime, we should keep track of the distinct set of dynamic columns in the schema, which will make a lot of our code simpler and faster. Insertions will probably suffer a small performance hit, but it's probably worth it given we can even it out or even improve upon that hit by improving comparison performance (creating the schema with dynamic columns is a large overhead of that).

thorfour:
One way I've thought of doing this since writing the SchemaV2 PoC is to keep track of the whole schema as the proto definition by way of proto merges like
https://github.com/polarsignals/frostdb/blob/9c79b8d7ff437bd91914ab70e832d4d0115b0a32/dynparquet/schema.go#L965

asubiotto:
That's a very interesting idea. Only thing I'd be concerned about is performance since we're probably going to do this on every insert (unless we mitigate by having a difference check pre-merge).

gernest:
With the new api for building `arrow.Record` We already keep tab on dynamic columns per `.Append` so its pretty much there already.

Considering  `.Append` works with batches, ` proto.Merge` won't be a big issue during insert , it will be done once per batch insert.

@asubiotto  do you have something in mind on how this info should be exposed ?

Personally I  think that now we have moved to `arrow.Record`, they should be exposed  as `*arrow.Schema` , which is very robust and supports attaching metadata as well. There won't be any need to check `.` separator for dynamic columns instead the info will be part of the column metadata. 

There is a bigger problem with this approach, to get up to date state one will need access to `(*Build[T])` instance.

gernest:
> I don't think I understand why this is the case. The way I envisioned it is that the concrete dynamic columns will be stored as part of the table schema on InsertRecord which is post Build[T], right?

Correct.  I understand your point, that tracking should be at table level , once record is passed to `InsertRecord`.

> As mentioned above, maybe that's the dynparquet.Schema attached to the table?

Sounds like the right place to start . Last I checked it , it was very much coupled with  `parquet`.

I'm more exited to lean more into  `arrow` and anything parquet specific should just be derived from `arrow` constructs.

asubiotto:
[...]

PR #399, opened 2023-03-16
394 project null columns

### LINKED ISSUE #394: Project null column if selecting column that doesn't exist
We need this for https://github.com/parca-dev/parca/pull/2667

So we can write a query like this:

```
labels.a == "test" AND pprof_labels.a IS NULL
or
labels.a is NULL AND pprof_labels.a == "test"
```

The problem is that this doesn't work currently since projections don't return a column if it doesn't actually exist.

thorfour:
Closes #394 

So I don't think we actually need to project the null column here to solve this problem, we just actually needed to handle the missing column case during binary scalar expression evaluations.

### DIFF
--- a/binaryscalarexpr.go
+++ b/binaryscalarexpr.go
@@ -49,6 +49,15 @@ func (e BinaryScalarExpr) Eval(p Particulate) (bool, error) {
 	// existant columns or null values. I'm pretty sure this is completely
 	// wrong and needs per operation, per type specific behavior.
 	if !exists {
+		if e.Right.IsNull() {
+			switch e.Op {
+			case logicalplan.OpEq:
+				return true, nil
+			case logicalplan.OpNotEq:
+				return false, nil
+			}
+		}
+
 		// only handling string for now.
 		if e.Right.Kind() == parquet.ByteArray || e.Right.Kind() == parquet.FixedLenByteArray {
 			switch {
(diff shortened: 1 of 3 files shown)


brancz (approved):
Awesome!! lgtm

asubiotto:
I'm curious what the use-case is (i.e. why we want to run queries on a column that doesn't exist), haven't thought about it much.

I'm not sure whether it's a good idea to do anything other than return an error when a query operates on a column that doesn't exist. Granted, we're now currently in a state where it's difficult to determine if a dynamic column doesn't exist *at all* in the table vs it's just not present in the row group/arrow record.

If we care about cases where the latter is true (running these queries on columns not present in a row goup/record), I believe that the long-term approach would be a side-effect of implementing static schemas (https://github.com/polarsignals/frostdb/issues/366). We would "fill in" the columns not found physically with a virtual null arrow column and not have to change anything in the execution engine.

thorfour:
> I'm curious what the use-case is (i.e. why we want to run queries on a column that doesn't exist), haven't thought about it much.
> 
> I'm not sure whether it's a good idea to do anything other than return an error when a query operates on a column that doesn't exist. Granted, we're now currently in a state where it's difficult to determine if a dynamic column doesn't exist _at all_ in the table vs it's just not present in the row group/arrow record.
> 
> If we care about cases where the latter is true (running these queries on columns not present in a row goup/record), I believe that the long-term approach would be a side-effect of implementing static schemas (#366). We would "fill in" the columns not found physically with a virtual null arrow column and not have to change anything in the execution engine.
> 
> In any case, if we need this behavior now I think it's fine to take your proposed approach.

I initially started down the road of physically projecting a null column but we really just wanted to be able to filter on the concept of a null column so this was the better way

PR #404, opened 2023-03-22
new granules atomic pointer

thorfour:
New granules is concurrently accessed, it's written by the compaction routine, and it's accessed by both reads and writes to propagate after compactions. 

Due to this it should be an atomic pointer so that access is thread safe.


brancz (approved):
Nice catch!

PR #407, opened 2023-03-23
Add table size on record insert

thorfour:
Change how record sizes are calculated by using an approximation based on number of rows in the slice and the overall record size


asubiotto:
Would it make sense to add regression test that checks that the expected sizes grow by a reasonable amount when inserting data?

bufutils/arrow.go
```diff
@@ -0,0 +1,19 @@
+package bufutils
+
+import (
+	"github.com/apache/arrow/go/v10/arrow"
+)
+
+// RecordSize returns the arrow record size in bytes.
+func RecordSize(r arrow.Record) int64 {
```
asubiotto, line 8: nit: this might make more sense in the `arrowutils` package

table_test.go
```diff
@@ -1419,3 +1420,97 @@ func Test_RowWriter(t *testing.T) {
...
+
+// Test_Table_Size ensures the size of the table increases by the size of the inserted data.
+func Test_Table_Size(t *testing.T) {
+	test := func(isArrow bool) {
+		c, table := basicTable(t)
+		defer c.Close()
+
+		before := table.ActiveBlock().Size()
+
+		samples := dynparquet.Samples{{
```
asubiotto, line 1432: nit: why not just use `dynparquet.NewTestSamples` and modify those if you need to? It's less verbose

asubiotto (approved):
LGTM minus nits

table_test.go
thorfour: simply forgot about it

bufutils/arrow.go
thorfour: bah humbug I forgot about that

PR #408, opened 2023-03-23
rewrite record insert to insert larger slices

### LINKED ISSUE #405: Arrow ingestion run away memory
image

After a short time of running Parca with the `--experimental-arrow` settings on we see compaction on a run away train consuming all the memory. Current suspicion is something is incorrect with the size calculations of records during compaction and is causing it to continuously compact granules.

thorfour:
Closes #405 

image

Previously we were splitting every inserted record into N parts where N was the number of rows in the record.
Now we'll insert contiguous rows together in the same part if they belong to the same granule.


brancz (approved):
Am I mistaken or should this also be faster? We're essentially binary searching from both ends right? Very nice work! 💯

PR #409, opened 2023-03-24
frostdb: add more granular column chunk filtering of NULL values

asubiotto:
The deleted comment did not hold true for me: it looks like the number of nulls is returned correctly from the column chunks


metalmatze (approved):

asubiotto:
Looks like `NumNulls` returns `0` for logic tests as the previous comment said. I'll dig in at some point.

asubiotto:
It was due to specifying a wrong definition level for nulls when inserting in logic tests (purely a test-related issue). This should now be fixed.

Issue #414, opened 2023-03-28
[arrow ingestion] panic on ingest


thorfour:
```
panic: arrow/array: index out of range

goroutine 6210 [running]:
github.com/apache/arrow/go/v10/arrow/array.(*Binary).Value(...)
	github.com/apache/arrow/go/v10@v10.0.1/arrow/array/binary.go:53
github.com/polarsignals/frostdb/pqarrow.appendToRow({0xc028532000, 0x14, 0x22}, {0x51d8440?, 0xc028499d60?}, 0x132, 0xc000096800?, 0xc02852c480?, 0x0?)
	github.com/polarsignals/frostdb@v0.0.0-20230323200733-2d118da30218/pqarrow/parquet.go:56 +0x12fa
github.com/polarsignals/frostdb/pqarrow.getRecordRow(0x51d4a88?, 0xc?, {0x51d4748, 0xc0284862a0}, 0xc001647140?, {0xc02852c480, 0x22, 0xc0284862d0?}, {0xc001515000, 0x22, ...})
	github.com/polarsignals/frostdb@v0.0.0-20230323200733-2d118da30218/pqarrow/parquet.go:85 +0x15a
github.com/polarsignals/frostdb/pqarrow.RecordToRow(0x5190a20?, 0xc02846f080, {0x51d4748, 0xc0284862a0}, 0xd?)
	github.com/polarsignals/frostdb@v0.0.0-20230323200733-2d118da30218/pqarrow/parquet.go:71 +0xcd
github.com/polarsignals/frostdb/pqarrow.RecordToDynamicRow(0xc0015281c0, {0x51d4748, 0xc0284862a0}, 0x132)
	github.com/polarsignals/frostdb@v0.0.0-20230323200733-2d118da30218/pqarrow/parquet.go:153 +0x111
github.com/polarsignals/frostdb.(*TableBlock).insertRecordToGranules(0xc000ffc000, 0xa, {0x51d4748, 0xc0284862a0})
	github.com/polarsignals/frostdb@v0.0.0-20230323200733-2d118da30218/table.go:1249 +0xa5
github.com/polarsignals/frostdb.(*TableBlock).InsertRecord(0xc000ffc000, {0x51bd8c0?, 0xc0263fed20?}, 0x3f6c762?, {0x51d4748, 0xc0284862a0})
	github.com/polarsignals/frostdb@v0.0.0-20230323200733-2d118da30218/table.go:1062 +0xcb
[...]

PR #415, opened 2023-03-28
fix arrow row conversion panic on nil dictionaries

thorfour:
We weren't handling null entries in dictionaries correctly when constructing rows from arrow records

### DIFF
--- a/pqarrow/parquet.go
+++ b/pqarrow/parquet.go
@@ -37,6 +37,11 @@ func ArrowScalarToParquetValue(sc scalar.Scalar) (parquet.Value, error) {
 }
 
 func appendToRow(row []parquet.Value, c arrow.Array, index, rep, def, col int) ([]parquet.Value, error) {
+	if c.IsNull(index) {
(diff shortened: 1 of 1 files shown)


thorfour:
Closes #414

pqarrow/parquet.go
```diff
@@ -37,6 +37,11 @@ func ArrowScalarToParquetValue(sc scalar.Scalar) (parquet.Value, error) {
 }
 
 func appendToRow(row []parquet.Value, c arrow.Array, index, rep, def, col int) ([]parquet.Value, error) {
+	if c.IsNull(index) {
+		row = append(row, parquet.ValueOf(nil).Level(rep, def, col))
```
asubiotto, line 41: Will it be possible to append a value with a def level > 0? I just ran into a subtle bug where we were doing this in logic test value creation. Parquet considers values null if the definition level != max definition level: https://github.com/segmentio/parquet-go/blob/main/column_buffer.go#L297

thorfour: Yes, will fix.

PR #418, opened 2023-04-02
update cols after updating aggregates

thorfour:
This fixes the disjointed label sets we were seeing with arrow ingestion. The refactor of the split-records moved the code that added nils to the arrays to be after the aggregate addition when it was supposed to be before. This undos that accidental part of the refactor.


thorfour:
> Is there an easy way to tickle this bug with a logic test to avoid potential future regressions? Might not be if we need to force a split

I wrote a couple tests that I thought would but was unable to get it to recreate locally.

PR #421, opened 2023-04-12
Data sources and sinks

thorfour:
This replaces the objstore bucket interface with a slightly more specific interface for remote storage called sources and sinks. 

This allows pluggable sources/sinks for remote storage for FrostDB to write and read to/from. This allows us to remove the "ignoreObjstoreOnQuery" flag and instead one could simply configure a sink but not a source to achieve the same goal. The separate set of sources and sinks allows for a more flexible system where one could perform migrations from old storage by configuring a new sink but keep a previous source to continue to read from.

I've also updated the Visitor pattern interface to include an in-order visitor.

### DIFF
--- a/db.go
+++ b/db.go
@@ -56,6 +56,9 @@ type ColumnStore struct {
 	indexDegree int
 	// splitSize is the number of new granules that are created when granules are split (default =2)
 	splitSize int
+
+	sources []DataSource
+	// TODO: sinks []DataSink
 }
 
 type metrics struct {
@@ -177,6 +180,19 @@ func WithStorage(bucket storage.Bucket) Option {
 	}
 }
 
+func WithDataSource(ds DataSource) Option {
+	return func(s *ColumnStore) error {
+		s.sources = append(s.sources, ds)
+		return nil
+	}
+}
+
+func WithDataSink(ds interface{}) Option {
+	return func(s *ColumnStore) error {
+		return nil
+	}
+}
+
 func WithWAL() Option {
 	return func(s *ColumnStore) error {
 		s.enableWAL = true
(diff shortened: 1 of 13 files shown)


db.go
```diff
@@ -293,10 +285,12 @@ type DB struct {
...
 
-	storagePath          string
-	wal                  WAL
-	bucket               storage.Bucket
-	ignoreStorageOnQuery bool
+	storagePath string
+	wal         WAL
+
+	sources []DataSource
+	sinks   []DataSink
```
asubiotto, line 292: Why a slice? Maybe worth commenting?

db.go
```diff
@@ -313,6 +307,19 @@ type DB struct {
 	metrics *dbMetrics
 }
 
+// DataSource is remote source of data that can be queried.
+type DataSource interface {
+	fmt.Stringer
+	Scan(ctx context.Context, prefix string, schema *dynparquet.Schema, filter logicalplan.Expr, lastBlockTimestamp uint64, stream chan<- any) error
```
asubiotto, line 313: Maybe it makes sense to pass a callback instead of a channel directly to allow the caller to choose how to process the scan results.

asubiotto (commented):
Very cool. I like this.

The only thing I wonder if we should change is whether the user should be aware of the source/sink differentiation. It might be unclear that you have to pass two options in order to read and write. Maybe we could have something like:
`WithReadWriteSource` (probably a better name, not just Source) which would set up the bucket as both a source and a sink under the hood.
And `WithReadOnlySource` which sets up a bucket only as a source.

db.go
thorfour: Done.

asubiotto (approved):

PR #424, opened 2023-04-24
wal: add metrics to track corruption on recovery

asubiotto:


metalmatze (approved):
Very nice work.
Overall it's fine to merge!

My only suggestion is to use a CounterVec here (`frostdb_wal_repairs_total`) too. Similar to what we recently discussed in the Parca Agent: https://github.com/parca-dev/parca-agent/pull/1526
I'm not sure which other metric we would compare `frostdb_wal_repairs_total` aginst. It might be worth cleaning this up a bit.

asubiotto:
I'm not sure I follow why a CounterVec would be a better option. As far as I understand it's basically adding dimensions to a counter (so you can record stuff like error or success). What would the labels be?
The intention is that the number of repairs is looked at independently. Ideally it stays at 0 but otherwise we have a general idea of whether we are truncating data to repair a WAL.

asubiotto:
Ah, do you mean we would have one counter `wal_repairs` and the CounterVec would be `wal_repairs{value="truncations"}` and `wal_repairs{value="lost_records"}`?

asubiotto:
Hmm, but it doesn't make sense to me what `wal_repairs` would be in that case though

PR #426, opened 2023-04-25
db: move snapshots and wal directories to trash

asubiotto:
Previously, db.Close would call os.RemoveAll on the snapshots and wal
directories. This is an O(n) operation, which could lead to some leftover
files if we exceed the graceful shutdown period.

This commit changes the removal to move both directories to a trash/ directory,
which is an O(1) operation. This trash/ directory can be removed safely on
startup if it still exists.

Additionally, this commit moves closeInternal to before removing the
directories. This ensures that the WAL is flushed and closed before deletion.

### DIFF
--- a/db.go
+++ b/db.go
@@ -376,6 +376,9 @@ func (s *ColumnStore) DB(ctx context.Context, name string) (*DB, error) {
 	}
 
 	if dbSetupErr := func() error {
+		if err := os.RemoveAll(db.trashDir()); err != nil {
+			return err
+		}
 		db.txPool = NewTxPool(db.highWatermark)
 		db.compactorPool = newCompactorPool(db, s.compactionConfig)
 		if len(db.sources) != 0 {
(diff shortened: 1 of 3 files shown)


db.go
```diff
@@ -767,17 +779,49 @@ func (db *DB) Close() error {
...
+			if err := os.RemoveAll(db.snapshotsDir()); err != nil {
+				return fmt.Errorf("%v: %v", moveErr, err)
+			}
+			if err := os.RemoveAll(db.walDir()); err != nil {
+				return fmt.Errorf("%v: %v", moveErr, err)
+			}
+			return moveErr
 		}
-		if err := os.RemoveAll(db.walDir()); err != nil {
+		if err := os.RemoveAll(trashDir); err != nil {
```
metalmatze, line 820: Just for my understanding: This is what might take too long and we don't finish in time to leave data dangling?

metalmatze (approved):
Very neat approach! :+1:

PR #428, opened 2023-04-26
wal: swap out tidwall/wal for polarsignals/wal

asubiotto:
We've seen a lot of corruption issues in prod using tidwall/wal.
polarsignals/wal is a fork of hashicorp/raft-wal which is a little more
production ready and conscious of storage medium failure modes, which hopefully
will reduce instances of WAL corruption on prod.


metalmatze (approved):
Awesome! Looking forward to seeing if there's a difference. Also incredible how few changes are actually needed :smiley:

asubiotto:
Added some debug logging to WAL shutdown to get some more insight into the CI failure. Tests are running fine locally.

asubiotto:
It looks like the issue is a race condition and Github CI is slow enough that it triggers easily 😂 . We truncate at a txn n after it's written to the queue but before writing it to the WAL, so the lastIndex of the WAL is 0 (after truncation) and we expect the nextTx to be 1, rather than n. Now reproducing this in a unit test so will fix locally.

asubiotto:
Changed our WAL implementation to return an error when it is asked to truncate outside the range: https://github.com/polarsignals/wal/pull/5

This is what the previous WAL implementation did, which brings up some interesting edge cases. It is possible for truncations to fail when we don't expect because of concurrency issues, leading to possible unclaimed space. Will create an issue and tackle in a separate PR.

Issue #425, opened 2023-04-25
pqarrow: panic on serializing arrow to parquet


asubiotto:
I see the following panic after we enabled arrow ingestion:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0x17d289f]

goroutine 10483 [running]:
github.com/apache/arrow/go/v10/arrow/array.(*Dictionary).Dictionary(...)
	github.com/apache/arrow/go/v10@v10.0.1/arrow/array/dictionary.go:232
github.com/polarsignals/frostdb/pqarrow.appendToRow({0xc0e42f0000, 0x1, 0x29}, {0x2c133a0, 0xc1527539a0?}, 0xc, 0xc000d80c00?, 0xc1086f1b80?, 0x0?)
	github.com/polarsignals/frostdb@v0.0.0-20230424150726-5288c29ad8a0/pqarrow/parquet.go:59 +0xcff
github.com/polarsignals/frostdb/pqarrow.getRecordRow(0xd?, 0xd?, {0x2c105c8, 0xc120178d20}, 0xa?, {0xc1086f1b80, 0x29, 0x3?}, {0xc037651000, 0x29, ...})
	github.com/polarsignals/frostdb@v0.0.0-20230424150726-5288c29ad8a0/pqarrow/parquet.go:90 +0x15a
github.com/polarsignals/frostdb/pqarrow.RecordToFile(0xc19b2ea700, 0xc14224baa0, {0x2c105c8, 0xc120178d20})
	github.com/polarsignals/frostdb@v0.0.0-20230424150726-5288c29ad8a0/pqarrow/parquet.go:179 +0x2ef
github.com/polarsignals/frostdb/parts.(*Part).SerializeBuffer(0xc19b2ea700?, 0x2bd9fc0?, 0xc1512679b0?)
	github.com/polarsignals/frostdb@v0.0.0-20230424150726-5288c29ad8a0/parts/part.go:53 +0x33
github.com/polarsignals/frostdb/parts.(*Part).AsSerializedBuffer(0xc152769040, 0xc19b2ea700)
	github.com/polarsignals/frostdb@v0.0.0-20230424150726-5288c29ad8a0/parts/part.go:69 +0x10a
[...]

Issue #419, opened 2023-04-04
physicalplan: Test_Aggregate_ArrayOverflow flake


asubiotto:
Running this test on a local branch caused this failure:
```
--- FAIL: Test_Aggregate_ArrayOverflow (8.76s)
    aggregate_test.go:55:
        	Error Trace:	/Users/asubiotto/Developer/github.com/polarsignals/frostdb/query/physicalplan/aggregate_test.go:55
        	            				/Users/asubiotto/Developer/github.com/polarsignals/frostdb/query/physicalplan/physicalplan.go:68
        	            				/Users/asubiotto/Developer/github.com/polarsignals/frostdb/query/physicalplan/aggregate.go:556
        	            				/Users/asubiotto/Developer/github.com/polarsignals/frostdb/query/physicalplan/aggregate_test.go:113
        	Error:      	Not equal:
        	            	expected: 952
        	            	actual  : 953
        	Test:       	Test_Aggregate_ArrayOverflow
```

Seems like the test is flaky cc @thorfour

PR #429, opened 2023-04-26
Fix 425 arrow panic

thorfour:
This addresses the arrow panics reported in #425 

It has compaction obtain a write block which causes the block rotation to wait for a compaction to finish. This will prevent the serialize command from accessing arrow records that have been released by compaction.

It also addresses a unit test failure reported in #419 

Fixes the array overflow mismatch by correctly rolling back the initially created aggregation, whereas before it was leaving that extra aggregation even though it had to create a new set.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -137,9 +137,14 @@ func (c *compactorPool) compactLoop(ctx context.Context) {
 			c.db.mtx.RUnlock()
 
 			for _, table := range tablesToCompact {
-				if err := table.ActiveBlock().compact(c.cfg); err != nil {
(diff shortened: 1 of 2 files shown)


asubiotto (approved):

compaction.go
```diff
@@ -137,9 +137,14 @@ func (c *compactorPool) compactLoop(ctx context.Context) {
 			c.db.mtx.RUnlock()
 
 			for _, table := range tablesToCompact {
-				if err := table.ActiveBlock().compact(c.cfg); err != nil {
+				block, done, err := table.ActiveWriteBlock() // obtain a write block to prevent compaction and persistence from running at the same time
```
asubiotto, line 140: Looking at this again: should we do the same for snapshots?

PR #434, opened 2023-04-27
dynparquet: reduce allocations

asubiotto:
This PR is the result of a Let's Profile stream where we looked at CPU usage in the frostdb ingestion path. The changes improve insertion benchmarks by ~50%.

The only thing I'm not quite sure about is that the third commit release pooled `parquet.Schema`s with a potential use after a release (e.g. storing the schema somewhere). I *think* this is fine because schemas are immutable once created, but it would be nice to have someone else's opinion.

### DIFF
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -116,6 +116,7 @@ type Schema struct {
 	writers        *sync.Map
 	buffers        *sync.Map
 	sortingSchemas *sync.Map
+	parquetSchemas *sync.Map
 }
 
 // IsDynamicColumn returns true if the passed in column is a dynamic column.
@@ -546,6 +547,7 @@ func newSchema(
 		writers:        &sync.Map{},
 		buffers:        &sync.Map{},
 		sortingSchemas: &sync.Map{},
+		parquetSchemas: &sync.Map{},
 	}
 
 	for i, col := range columns {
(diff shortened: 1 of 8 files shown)


dynparquet/schema.go
```diff
@@ -1077,6 +1092,56 @@ func (s *Schema) GetWriter(w io.Writer, dynamicColumns map[string][]string) (*Po
...
 
+type PooledParquetSchema struct {
+	pool   *sync.Pool
+	Schema *parquet.Schema
+}
+
+// GetParquetSortingSchema returns a parquet schema of the sorting columns and
+// the given dynamic columns.
+// The difference with GetDynamicParquetSchema is that non-sorting columns are elided.
+func (s *Schema) GetParquetSortingSchema(dynamicColumns map[string][]string) (*PooledParquetSchema, error) {
```
thorfour, line 1103: This isn't being used anywhere that I can tell. Do we need it?

pqarrow/parquet.go
```diff
@@ -150,12 +150,13 @@ func RecordToDynamicRow(schema *dynparquet.Schema, record arrow.Record, index in
 		return nil, io.EOF
 	}
 
-	ps, err := schema.DynamicParquetSchema(RecordDynamicCols(record))
+	ps, err := schema.GetDynamicParquetSchema(RecordDynamicCols(record))
```
thorfour, line 153: I think we can improve this further by pulling the dynamic schema out of the `RecordToDynamicRow` function since it calls that function for each row in the record, as well as the `RecordToDynamicSchema` function

dynparquet/schema.go
metalmatze: It's used in `dynparquet/row.go`. See above.

thorfour: Doh! Github had collapsed that file

pqarrow/parquet.go
asubiotto: Done. RFAL.

asubiotto:
This is ready for another review.

brancz (approved):

PR #435, opened 2023-05-01
Repeated schema

thorfour:
Adds support for repeated columns into schema v1alpha1.

This only adds support for repeated binary columns that are dictionary encoded. But it adds the necessary scaffolding for other supported types in the future as needed.


asubiotto (approved):
LGTM, but I thought we were not going to support this in v1 and instead move forward with v2?

PR #436, opened 2023-05-02
wal: perform truncation in run

### LINKED ISSUE #431: wal: truncations can sometimes unexpectedly fail
In #428, CI uncovered an interesting edge case in our WAL usage. We have two goroutines: one in https://github.com/polarsignals/frostdb/blob/06fe1fc7c4904417cfbe9bf89fb4acb0e9b190e3/wal/wal.go#L178
and another arbitrary goroutine that can perform truncations concurrently:
https://github.com/polarsignals/frostdb/blob/06fe1fc7c4904417cfbe9bf89fb4acb0e9b190e3/wal/wal.go#L282

The edge case is that a goroutine logs a record with index `k` (e.g. block persistence) and then performs a truncation at that same index. If the `run` goroutine has not read and logged the record with this index, the truncation returns an `out of range` error, since the expected index has not been persisted yet.

To fix these cases, we should move truncation into the `run` goroutine so that it can be executed in the expected order.

asubiotto:
Previously, we would synchronously truncate even though we asynchronously wrote
to the WAL. This caused truncations to fail with high likelihood given that the
transaction to truncate at was likely not yet logged.

Closes #431

### DIFF
--- a/wal/wal.go
+++ b/wal/wal.go
@@ -71,10 +71,12 @@ type FileWAL struct {
 
 	metrics        *fileWALMetrics
 	logRequestCh   chan *logRequest
-	queue          *logRequestQueue
 	logRequestPool *sync.Pool
 	arrowBufPool   *sync.Pool
-	mtx            *sync.Mutex
+	protected      struct {
+		sync.Mutex
+		queue logRequestQueue
+	}
 
 	cancel     func()
 	shutdownCh chan struct{}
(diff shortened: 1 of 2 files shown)


wal/wal.go
```diff
@@ -71,10 +71,17 @@ type FileWAL struct {
...
-	mtx            *sync.Mutex
+	protected      struct {
+		sync.Mutex
+		queue logRequestQueue
+		// truncateTx is set when the caller wishes to perform a truncation. The
+		// WAL will keep on logging records up to and including this txn and
+		// then perform a truncation. If another truncate call occurs in the
+		// meantime, the highest txn will be used.
+		truncateTx uint64
+	}
```
metalmatze, line 84: Interesting pattern!

metalmatze (approved):

PR #441, opened 2023-05-09
*: clean up and add more snapshot metrics

### LINKED ISSUE #391: snapshot: add more metrics to snapshots
It would be good to have an idea of how long snapshots are taking as well as the resulting size

asubiotto:
Closes #391

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -13,6 +13,9 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
 	"github.com/apache/arrow/go/v10/arrow/ipc"
 	"github.com/go-kit/log/level"
 	"github.com/google/btree"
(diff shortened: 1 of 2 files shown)


asubiotto:
Snapshot duration is recorded in a Gauge. Does that sound good? Felt like a histogram doesn't really fit here.

brancz:
A gauge is not super useful, as it can only represent the latest snapshot, and we can't tell when that snapshot happened. A histogram will tell us the bucket that the snapshot duration fell into and at which time the snapshot happened (since buckets are counters and it would count up at that time which we can detect with an `increase` query). I think a histogram makes more sense.

snapshot.go
```diff
@@ -66,6 +69,33 @@ const (
...
+			[]string{"success"},
+		),
+		snapshotFileSizeBytes: promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+			Name: "file_size_bytes",
+			Help: "Size of snapshots in bytes",
+		}),
+		snapshotDurationHistogram: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+			Name:    "duration_seconds",
+			Help:    "Duration of snapshots in seconds",
+			Buckets: prometheus.ExponentialBucketsRange(0.1, 120, 30),
```
brancz, line 94: That's a lot of buckets for a metric that we have per database, we might want to adapt this after we get a feeling for the snapshot durations in prod.

The number of series we get out of this is `30 * number of databases * nodes in the system`, so even now this is already ~1000 series. It's negligible for now but something to keep in mind.

brancz (approved):

PR #444, opened 2023-05-10
Update bufbuild plugin references

mfridman:
Hey there, I noticed you were referencing old Remote Plugins in the `buf.gen.yaml` file. There are upcoming [Remote Plugin changes](https://buf.build/blog/remote-packages-remote-plugins-approaching-v1) and this PR ensures you don't get affected.

In summary,

1. Update plugins to the Buf-managed ones. The build source for these plugins is now open-sourced at [github.com/bufbuild/plugins](https://github.com/bufbuild/plugins)
2. Run `buf generate` with `buf` CLI version 1.17.0


buf.gen.yaml
```diff
@@ -6,12 +6,12 @@ managed:
 
 plugins:
   # renovate: datasource=github-releases depName=protocolbuffers/protobuf-go
-  - remote: buf.build/protocolbuffers/plugins/go:v1.28.0-1
+  - plugin: buf.build/protocolbuffers/go:v1.30.0
```
mfridman, line 9: Newer versions of the `buf` CLI use the `plugin` key for both local and remote plugin references.

This also updates both plugins to the latest available upstream versions.

metalmatze (approved):
Oh! Thank you so much! We updated Parca but seems like we forgot about FrostDB...

Issue #452, opened 2023-05-18
logictest: run tests with arrow storage


asubiotto:
Currently, our logic test framework writes all data in tests as parquet rows:
https://github.com/polarsignals/frostdb/blob/4c82bcbc7f806bd2fae6ec54d6a802dbc253f7d8/logictest/runner.go#L221

We should run another subtest with a flag on the logictest runner that inserts data as arrow records instead of row groups.

PR #453, opened 2023-05-18
run all logic tests as arrow

thorfour:
In addition to running the logic tests with Parquet insertions we now also run the tests with arrow insertion


brancz (approved):

thorfour:
Closes #452

PR #454, opened 2023-05-19
Fix: remove leftover incorrect append

thorfour:


brancz:
Is there maybe a test that we could write for this?

PR #459, opened 2023-05-31
use arrow pkg util for total record size

thorfour:
Use the provided `TotalRecordSize` util from the arrow package to correctly calculate record sizes.


asubiotto (approved):
Fantastic, thanks. Great that we can reuse an arrow function

PR #461, opened 2023-06-15
pqarrow/arrowutils: Add SortRecord and make MergeRecords accept a limit

metalmatze:
Add `SortRecord` to sort a single arrow record by a single column. This currently only supports `int64`, `string`, and `bytes`. 
This works nicely when having multiple records returned, to sort them individually and afterward pass them to MergeRecords where they get sorted with heapsort. This now supports an optional limit where the sorting stops once the limit has been reached.


thorfour:
Is this what we want or would it make more sense to contribute back to the arrow project to implement the sort record function?

metalmatze:
This was mostly a hack for only what we need. I'm not sure it's generic enough for the upstream package. We can definitely take this forward and contribute it upstream, if we put some more time and effort into this.

asubiotto:
I think one thing that is missing that I recall the upstream implementation did (and I think is probably a good idea) is that the result of the sort is a slice of indices into the array, rather than the rearranged array itself. Long term I think we want to have this idea of a "selection" vector accompanying records as well since it'll make things like filtering cheaper.

metalmatze:
Makes sense Alfonso! 
I'll take a closer look at https://arrow.apache.org/blog/2022/11/07/multi-column-sorts-in-arrow-rust-part-1/ eventually. For now, I am going to send an update that splits the `SortRecord` into `SortRecord` that returns the indices and will add another helper function that `ReorderRecord` takes the indices and copies the actual rows in the correct order.

metalmatze:
This branch contains multiple unrelated features at this point. Closing this PR and opening newer smaller ones.

PR #463, opened 2023-06-22
aggregate: directly hash the list values

thorfour:
Previously we were appending all the list values and then hashing them. This becomes extremely inefficient when using really large lists. Instead we just hash the list values directly.

Below is a benchmark for a normal aggregate query against 12h of real profiling data.

```bash
name              old time/op    new time/op    delta
_PreAggregate-10     6.56s ± 2%     4.81s ± 3%  -26.76%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op   delta
_PreAggregate-10    33.1GB ± 0%    14.6GB ± 0%  -55.94%  (p=0.000 n=8+10)

name              old allocs/op  new allocs/op  delta
_PreAggregate-10      174M ± 0%      162M ± 0%   -6.99%  (p=0.000 n=10+9)
```

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -12,6 +12,7 @@ import (
 	"github.com/apache/arrow/go/v13/arrow/math"
 	"github.com/apache/arrow/go/v13/arrow/memory"
 	"github.com/apache/arrow/go/v13/arrow/scalar"
+	"github.com/cespare/xxhash/v2"
 	"github.com/dgryski/go-metro"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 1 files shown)


query/physicalplan/aggregate.go
```diff
@@ -281,18 +282,20 @@ func hashArray(arr arrow.Array) []uint64 {
...
 	res := make([]uint64, arr.Len())
+	digest := xxhash.New()
 	for i := 0; i < arr.Len(); i++ {
-		list := []byte{}
 		if err := arrowutils.ForEachValueInList(i, arr, func(_ int, v any) {
 			switch val := v.(type) {
 			case []byte:
-				list = append(list, val...)
+				// NOTE: We can ignore the return here because from the xxhash Write function is says: It always returns len(b), nil.
+				_, _ = digest.Write(val)
```
asubiotto, line 291: Just double checking: would e.g. `[a, b, c]` and `[b, c, a]` hash to different values?

asubiotto (approved):

query/physicalplan/aggregate.go
thorfour: Correct; can see here: https://go.dev/play/p/rKO6WL2JK54

PR #464, opened 2023-06-22
Query limits

thorfour:
This introduces a new type of `memory.Allocator` that acts as a wrapper around the arrow `GoMemoryAllocator` which will account for the number of bytes allocated but not yet freed and will panic if that limit is exceeded.

In addition it adds a recovery step into the query engine, where if a child go routine panics in the query engine, the engine will capture that panic and return an error instead.

### DIFF
--- a/query/memory.go
+++ b/query/memory.go
@@ -1,0 +1,53 @@
+package query
+
+import (
+	"sync/atomic"
+
+	"github.com/apache/arrow/go/v13/arrow/memory"
+)
+
+const PanicMemoryLimit = "memory limit exceeded"
+
+var _ memory.Allocator = (*LimitAllocator)(nil)
+
+// LimitAllocator is a wrapper around a memory.Allocator that panics if the memory usage exceeds the defined limit.
+type LimitAllocator struct {
+	limit     int64
+	allocated *atomic.Int64
+	allocator memory.Allocator
+}
+
+func NewLimitAllocator(limit int64, allocator memory.Allocator) *LimitAllocator {
+	return &LimitAllocator{
+		limit:     limit,
+		allocated: &atomic.Int64{},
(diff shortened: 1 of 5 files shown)


brancz (approved):
Very nice!

query/memory.go
```diff
@@ -0,0 +1,53 @@
...
+
+	allocated := a.allocated.Add(int64(size - len(b)))
+	if allocated > a.limit {
+		panic(PanicMemoryLimit)
+	}
+	buf := a.allocator.Reallocate(size, b)
+	return buf
+}
+
+func (a *LimitAllocator) Free(b []byte) {
```
asubiotto, line 50: Something I would love to do (maybe not in this PR) is to add some sort of verifier similar to `goleak`'s leak detection in `TestMain` that somehow verifies that we aren't forgetting to release anything. I think running this e.g. with the logic tests would help us be confident here.

recovery/recovery.go
```diff
@@ -0,0 +1,22 @@
+package recovery
+
+import (
+	"fmt"
+)
+
+// Do is a function wrapper that will recover from a panic and return the error.
```
asubiotto, line 7: I think this is very nice since it theoretically means our query engine cannot crash a node anymore. But are there maybe some panics we do want to make sure go through? At cockroach we took the approach that all panics from the query engine are treated as errors and that's it. One way to "filter" is to panic with a specific object (e.g. ErrOutOfMemory) that is then checked here.

asubiotto (approved):

recovery/recovery.go
thorfour: I think at present we don't have a use case to let a panic through so can leave it for now. But we certainly can add future handling here if need be.

PR #470, opened 2023-07-11
Fix limit leaks

thorfour:
This adds a couple unit tests, but also adds a allocator leak check to the logic tests.

It fixes a lot of the leaks that were uncovered when adding the unit tests.

NOTE: that the check is skipped for the ordered aggregate operator because I was unable to get that one to work.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1741,3 +1741,50 @@ func TestReplayBackwardsCompatibility(t *testing.T) {
 	require.NoError(t, err)
 	defer c.Close()
 }
+
+func Test_DB_Limiter(t *testing.T) {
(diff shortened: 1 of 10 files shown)


asubiotto (approved):

metalmatze (approved):
Wow...
Were we leaking all of these records that have not been released?

thorfour:
> Wow... Were we leaking all of these records that have not been released?

So leaking is a bit of an overloaded term here. There was no physical memory that was being held forever since we use the default allocator which is just the built in Go allocation. So while we weren't always calling free on memory that we had allocated we would eventually lose all references to it and Go would clean it up for us.

PR #471, opened 2023-07-13
pqarrow/builder: Add Value to OptInt64Builder

metalmatze:
Turns out I had this locally all the time and never made the PR. 
This is needed for https://github.com/parca-dev/parca/pull/3448.

### DIFF
--- a/pqarrow/builder/optbuilders.go
+++ b/pqarrow/builder/optbuilders.go
@@ -394,6 +394,11 @@ func (b *OptInt64Builder) Add(i int, v int64) {
 	b.data[i] += v
 }
 
+// Value returns the ith value of the builder.
+func (b *OptInt64Builder) Value(i int) int64 {
+	return b.data[i]
+}
+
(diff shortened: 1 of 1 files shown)


pqarrow/builder/optbuilders.go
```diff
@@ -81,6 +81,11 @@ func (b *builderBase) appendValid(n int) {
 	b.length += n
 }
 
+// IsNull returns whether the element at index i is null.
+func (b *builderBase) IsNull(i int) bool {
```
asubiotto, line 85: I think we should add these utility functions to the optbuilder interface, what do you think?

asubiotto (approved):
LGTM. Are you still planning on adding the Value and IsNull to the interface?

metalmatze:
I did. Sadly that's the one I accidentially pushed to `main` directly: https://github.com/polarsignals/frostdb/commit/1a323c8b843ff9d13f7e1f4b108b53cbfee09686

As for the `Value` I don't see how that can be added to the interface with different return types.

PR #476, opened 2023-07-24
support map as dynamic type

albertlockett:
I would like to propose consider supporting `map[string]...` as a type that can be inserted into a dynamic column.

Currently using a map as the value for a dynamic column will fail:
```golang
func main() {
	columnstore, _ := frostdb.New()
	db, _ := columnstore.DB(context.Background(), "test")
	table, _ := db.Table("test", frostdb.NewTableConfig(&schemapb.Schema{
		Columns: []*schemapb.Column{
			{
				Name: "name",
					Type: schemapb.StorageLayout_TYPE_STRING,
				},
			},
			{
				Name: "attributes",
					Type: schemapb.StorageLayout_TYPE_STRING,
				},
				Dynamic: true,
			},
		},
	}))

	record := struct {
		Name       string
		Attributes map[string]string
	}{
		Name: "albert",
		Attributes: map[string]string{
			"age": "9999",
		},
	}

	_, err := table.Write(context.Background(), record) // <---- Panics
```

Here is the panic. I would like to also suggest improving the error message to include the column name.
```
2023-07-24 17:43:01.303927 I | err happen
panic: unsupported dynamic type

goroutine 1 [running]:
main.main()
        /home/albertlockett/Development/otel-logs-frostdb/example/main.go:48 +0x5fb
exit status 2
```

### DIFF
--- a/table.go
+++ b/table.go
@@ -608,6 +608,14 @@ func ValuesToBuffer(schema *dynparquet.Schema, vals ...any) (*dynparquet.Buffer,
 						pair := reflect.ValueOf(dynVals.Index(j).Interface())
 						dynamicColumns[col.Name] = append(dynamicColumns[col.Name], ToSnakeCase(fmt.Sprintf("%v", pair.Field(0))))
 					}
+				case reflect.Map:
+					dynVals := reflect.ValueOf(cv)
+					for _, key := range dynVals.MapKeys() {
+						if key.Kind() != reflect.String {
+							return nil, fmt.Errorf("unsupported dynamic map key type for column")
+						}
+						dynamicColumns[col.Name] = append(dynamicColumns[col.Name], ToSnakeCase(fmt.Sprintf("%v", key.String())))
+					}
 				default:
 					return nil, fmt.Errorf("unsupported dynamic type")
 				}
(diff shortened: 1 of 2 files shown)


table.go
```diff
@@ -608,8 +608,16 @@ func ValuesToBuffer(schema *dynparquet.Schema, vals ...any) (*dynparquet.Buffer,
 						pair := reflect.ValueOf(dynVals.Index(j).Interface())
 						dynamicColumns[col.Name] = append(dynamicColumns[col.Name], ToSnakeCase(fmt.Sprintf("%v", pair.Field(0))))
 					}
+				case reflect.Map:
+					dynVals := reflect.ValueOf(cv)
+					for _, key := range dynVals.MapKeys() {
+						if key.Kind() != reflect.String {
```
thorfour, line 614: Can we make this a switch statement and just have the default case return the error?

thorfour (commented):
Nice! Thanks for this addition!

table.go
albertlockett: Sure thing -- made the fix. Thanks @thorfour

thorfour:
Thanks again for this contribution!

PR #477, opened 2023-07-25
Limit allocation fixes for panic

thorfour:
This adds fixes for leaking allocation when the limit allocator panics.

Of note it adds the `Close()` method to output plans that is defer called during Execute to ensure resources are Released.
It also fixes a couple of other corner cases of leaking resources on panic.


db_test.go
```diff
@@ -1778,12 +1778,6 @@ func Test_DB_Limiter(t *testing.T) {
 
 	for i := 0; i < 1024; i++ {
```
asubiotto, line 1779: Would choosing `k` random limits in this range rather than running the test for every single limit in range suffice?

asubiotto (approved):
very nice

db_test.go
thorfour: I fear that might bury a regression, and the test is cheap to run

PR #478, opened 2023-07-26
fix panic on write when dynamic column not defined

albertlockett:
When writing to some table, if the schema has a dynamic column and the value being written does not define this column, then the write will panic

```golang
table, _ := db.Table("test", frostdb.NewTableConfig(&schemapb.Schema{
Columns: []*schemapb.Column{
	{
		Name: "name",
			Type: schemapb.StorageLayout_TYPE_STRING,
		},
	},
	{
		Name: "attributes",
			Type: schemapb.StorageLayout_TYPE_STRING,
		},
		Dynamic: true,
	},
},
}))

record := struct {
	Name string
	// 'attributes' is not a member of the struct
}{
	Name: "albert",
}

_, err := table.Write(context.Background(), record) // <---- panics
```

This PR fixes the panic

### DIFF
--- a/table.go
+++ b/table.go
@@ -594,6 +594,9 @@ func ValuesToBuffer(schema *dynparquet.Schema, vals ...any) (*dynparquet.Buffer,
 		val := reflect.ValueOf(v)
 		for _, col := range schema.Columns() {
 			cv := findColumn(val, col.Name, v)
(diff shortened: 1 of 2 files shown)


thorfour (approved):
Thanks for the fix!

PR #482, opened 2023-08-04
fixes issues w/ too many arrow Release

### LINKED ISSUE #403: Fix Release() arrow callbacks
Our usage of Release of arrow records during callback is inconsistent. If operators are utilized they won't release the callback record, but will create a new one which the final callback can release. 

However if there is no operators and it's just straight iteration then a release on the final callback will cause too many releases to occur.

This is easy to reproduce by running unit tests with `-tags assert`

The final callback should always be able to safely release the given record.

We should detect this automatically but adding `-tags assert` to our github workflow.

albertlockett:
Fixes #403 

The issue here was that when the physical plan executes concurrently, the `Close` method on our operations would get called multiple times. HashAggregation was reasonably assuming the `Close` method would only get called once, but since the instance after the synchronizer is shared between concurrent plans, this was not the case. This PR fixes it so the syncrhonizer only calls `next.Close` once.


query/physicalplan/synchronize.go
```diff
@@ -62,5 +63,12 @@ func (m *Synchronizer) Draw() *Diagram {
 }
 
 func (m *Synchronizer) Close() {
-	m.next.Close()
+	if !m.closed {
```
thorfour, line 66: So I think this is fine, but could we keep this close logic consistent with the finish logic where it waits to call close until it has received all the expected close calls

albertlockett: ohhh yeah sure I'll make it so 👍

albertlockett: done :+1:

PR #487, opened 2023-08-09
wal: allow Truncating past end of WAL

### LINKED ISSUE #412: db: handle cases where WAL last index != snapshot txn
There might be cases where the WAL hasn't caught up to a given snapshot txn for whatever reason (e.g. WAL corruption causes an end truncation). In cases where the WAL's last index is < the snapshot txn during recovery, we should remove all WAL files and create a new WAL file with the given snapshot's last index. This will ensure that the WAL can be appended to.

asubiotto:
This PR implements the logic to allow truncating past the end of the WAL,
i.e. resetting the WAL entirely. The only condition is that the next record
logged MUST be the txn immediately following the txn at which the WAL was
truncated.

This is very useful for recovery where we have a snapshot with a txn k and the
last index of a WAL is j < k-1, so we don't really care about the WAL entries
anymore.

Closes #412


brancz (approved):
Neat! Actually makes the code easier to follow than before.

PR #489, opened 2023-08-10
wal: time out Close 

### LINKED ISSUE #486: wal: time out Close
I've found on many instances that the wal package is stuck closing because the next minimum transaction is not the next expected transaction, causing an endless loop waiting for the WAL to drain:
https://github.com/polarsignals/frostdb/blob/336457ecf250723103a7ede2a011377280143206/wal/wal.go#L185-L204

I think it would be nice to put a timeout (maybe after a second of no new records logged) so that the WAL closes and logs an error even in this invalid state. It might be nice to also have metrics on this invalid close, because it's usually indicative of a bug.

Having this hard shutdown helps when running frostdb on pre-emptible machines that have a max time to shutdown. If the WAL doesn't shutdown in time, the whole machine is shut down, skipping other steps of a graceful shutdown which are nice to have (and would even obviate this invalid WAL state, e.g. if we take a snapshot or persist the data).

albertlockett:
Fixes #486

### DIFF
--- a/wal/wal_test.go
+++ b/wal/wal_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


wal/wal_test.go
```diff
@@ -269,3 +271,43 @@ func TestWALTruncate(t *testing.T) {
...
+		prometheus.NewRegistry(),
+		dir,
+	)
+	require.NoError(t, err)
+	defer w.Close()
+
+	closed := atomic.Bool{}
+	closed.Store(false)
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
```
asubiotto, line 288: nit: rather than setting up the cancel infra and running the wal yourself, I think it would be cleaner to call `w.RunAsync` here and defer `w.Close`. It's fine to fail the test due to timeout.

albertlockett: ah of course, good call will fix

asubiotto (approved):

asubiotto:
Thanks for the contribution!

PR #490, opened 2023-08-14
*: reset the WAL to zero in resetToTxn

asubiotto:
Previously, we would use the given txn to truncate the WAL. However, this txn could be less than the last index, causing no records to be logged until the high watermark caught up with the last index.

This commit performs an unconditional truncation using the max uint64, resulting in a full reset of the WAL.


asubiotto:
Still need to do some more modifications before merging, I think a test should fail as well.

asubiotto:
This is ready for a look now

brancz (approved):

PR #491, opened 2023-08-14
Store snapshot

thorfour:
Small refactor inject a writer dependency into `snapshotAtATX` which allows for a new function `StoreSnapshot` that can write an external snapshot into the snapshot directory.


asubiotto (approved):
Very nice

PR #498, opened 2023-08-23
use RemoveAll instead of Remove

thorfour:


db.go
```diff
@@ -1117,5 +1117,5 @@ func validateName(name string) bool {
 
 // DropStorage drops all data in the storage.
 func (db *DB) DropStorage() error {
-	return os.Remove(db.storagePath)
```
asubiotto, line 1120: Can we successfully log to the WAL after this?

PR #501, opened 2023-08-25
Aggregation dyn cols

thorfour:
Adds support for performing aggregations on dynamic columns.

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -2,6 +2,8 @@ package frostdb
 
 import (
 	"context"
+	"fmt"
+	"math/rand"
 	"sort"
 	"strconv"
 	"testing"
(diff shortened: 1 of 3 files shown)


aggregate_test.go
```diff
@@ -335,3 +337,75 @@ func BenchmarkAggregation(b *testing.B) {
 		})
 	}
 }
+
+func Test_Aggregation_DynCol(t *testing.T) {
```
asubiotto, line 341: Can we write this as a logic test? I think we can reuse the default schema and find the max label or something

thorfour: Would have to add support for aggregation on strings which is why I wrote it as this. Was just more work to write as a logictest

asubiotto (approved):

Issue #347, opened 2023-02-15
Support grouping multiple rows of first match in an interval


metalmatze:
Right now Parca queries FrostDB for all the timestamps and on the server side chooses to ignore metric samples when a sample has been found in the step's bucket.

https://github.com/parca-dev/parca/blob/ed90dbeb684186e9cdb295bc0f62c723ed3c5a9f/pkg/parcacol/querier.go#L571-L579

This isn't ideal since the data is still queried...

FrostDB should add support to find all rows for a bucket of interval and only then start aggregating. 

The row data would look like this:
| timestamp 	| stacktrace 	| value 	|
|-----------	|------------	|-------	|
| 1         	| stack1     	| 3     	|
| 1         	| stack2     	| 5     	|
| 1         	| stack3     	| 8     	|
| 2         	| stack1     	| 2     	|
| 2         	| stack2     	| 3     	|

What we want. We want to only return the sum(value) for the first timestamp we find in the bucket that ranges from 0-9. 
So what we want as a result is:
| timestamp 	| sum(value) 	|
|-----------	|------------	|
| 1         	| 16          	|

What we currently get is however the very first row that falls within that bucket `(1, stack1, 3)` and then the sum of it's value, which is basically a noop:
| timestamp 	| sum(value) 	|
|-----------	|------------	|
| 1         	| 3          	|

metalmatze:
This issue is especially bad for Parca itself. On our cloud with the aggregate_view table, it shouldn't be such a bad problem for now.

albertlockett:
@metalmatze As far as I an tell the `Duration` aggregation does what we want. I added a test to demonstrate.
https://github.com/polarsignals/frostdb/pull/503/files

That said, I think this is different than what we do here in parca, where we take the first result in the interval.
https://github.com/parca-dev/parca/issues/2598

I'm wondering if maybe what we want is a `First` aggregation instead like this?
https://github.com/polarsignals/frostdb/pull/504

PR #504, opened 2023-08-28
Added first aggregation

albertlockett:
Adds a `First` aggregation.. 

```golang
_ = engine.ScanTable("test").
Aggregate(
	[]logicalplan.Expr{
		logicalplan.First(logicalplan.Col("timestamp")),
	},
	[]logicalplan.Expr{
		logicalplan.Duration(time.Second),
	},
).
Execute(context.Background(), func(_ context.Context, r arrow.Record) error {
	r.Retain()
	results = append(results, r)
	return nil
})
```

this can be used to get the First value in some interval, which I think is what we might want to do [here](https://github.com/parca-dev/parca/issues/2598) (I'm not sure about this, so opening this as a draft while clarifying behaviour)

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -248,7 +248,7 @@ func TestAggregationProjection(t *testing.T) {
 	require.True(t, record.Schema().HasField("max(value)"))
 }
 
-func TestAggregationLimit(t *testing.T) {
+func TestAggregationTakeLimit(t *testing.T) {
 	config := NewTableConfig(dynparquet.SampleDefinition())
 	logger := newTestLogger(t)
 	c, err := New(WithLogger(logger))
@@ -333,8 +333,7 @@ func TestAggregationLimit(t *testing.T) {
 	record := records[0]
 
 	require.True(t, record.Schema().HasField("timestamp"))
-	// TODO --
-	// require.True(t, record.Schema().HasField("limit(value, 2)"))
+	require.True(t, record.Schema().HasField("take(value, 2)"))
 
 	require.Equal(t, int64(2), record.NumCols())
 	require.Equal(t, int64(3), record.NumRows())
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -2,7 +2,6 @@ package frostdb
 
 import (
 	"context"
-	"fmt"
 	"math/rand"
 	"sort"
 	"strconv"
@@ -323,7 +322,6 @@ func TestAggregationTakeLimit(t *testing.T) {
 		Execute(context.Background(), func(ctx context.Context, ar arrow.Record) error {
 			records = append(records, ar)
 			ar.Retain()
-			fmt.Printf("%s", ar)
 			return nil
 		})
 
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -2,6 +2,7 @@ package frostdb
 
 import (
 	"context"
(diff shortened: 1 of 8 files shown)


asubiotto:
I think I would prefer implementing a more general Limit operator. This First agg func seems to specific to me.

albertlockett:
> I think I would prefer implementing a more general Limit operator. This First agg func seems too specific to me. I might be misunderstanding the problem.

ah thanks @asubiotto yes that's a good call

metalmatze:
Oh hey, @albertlockett. Nice to see you around! :wave: 

I think the overall approach here looks good to me. 
Not sure about the naming either. Making it a more generic limit operator seems like a good call.

I would need to read up some more on those too, but this looks pretty much like it indeed: https://stackoverflow.com/questions/2129693/using-limit-within-group-by-to-get-n-results-per-group

albertlockett:
Oh hi, @metalmatze 👋 

Of course, I'll change the implementation to be a more generic limit of group! It's more flexible and effectively supports the  `First` use case when the Limit=1.

The impetus for adding this is https://github.com/polarsignals/frostdb/issues/347 . Maybe it's not required for this PR, if this is a feature we want anyway, but wondering if you can clarify what's needed here? The frostdb issue reads like we actually want to use the `Sum` aggregation expression w/ the duration group expression, but the parca code reads like we'd want to use a Limit=1 aggregation expression.

albertlockett:
@metalmatze @asubiotto I made the change we discussed above here -- to have a more general aggregation function.

Hoping to get your feedback on the approach ...

I called the aggregation function "take" instead of "limit". I thought it was maybe a better name because:
- we might want to reserve `logicalplan.Limit` to create an expression that can limit the number of results we return from the query
- "limit" to me implies were getting all results to begin with and wish to limit to only some, but this isn't really the case here.
That said, I don't feel strongly about what we call it.

Here's a simple of how it can be used:
```golang
// get the first "value" at each timestamp
err = engine.ScanTable("test").
Aggregate(
	[]logicalplan.Expr{
		logicalplan.Take(logicalplan.Col("value"), 1),
	},
	[]logicalplan.Expr{
		logicalplan.Duration(time.Second),
	},
).
// ...
```

And here's how it's indended to be used in parca:

https://github.com/parca-dev/parca/blob/838e7bda405d281c265a484f71de5cda8eea0e40/pkg/parcacol/querier.go#L531-L556

One tricky thing about this is that the actual `AggregationFunction` in the physical plan needs some extra arguments passed from the logicalplan (in this case, how many results to take). The way I approcahed this is passing around a type called `AggregationArguments` (which is a map). This probably isn't the only approach we could take, so if you guys don't like how this is done happy to change.

asubiotto:
Hi @albertlockett, I think there's been some confusion. I did not mean to imply that we needed a limit *aggregation function*, rather a limit *operator*, which is what you describe in your second bullet point (i.e. the reason to reserve the `logicalplan.Limit` function). If a general-purpose limit operator is not enough/doesn't work for what we want to de here then I think we should go back to the initial intention of this PR which is to create a `First` aggregation function.

I appreciate all the work you've put into this and I apologize for the confusion.

albertlockett:
Hey @asubiotto it's all good no need to apologize. I'll try to get this PR fixed up when I have some time to come back to it!

PR #511, opened 2023-08-31
*: listen to context cancellation when collecting records

asubiotto:
There are cases where the context gets canceled and granule collection would not react.


asubiotto:
Also added another commit to change the `IterateTables` interface.

PR #524, opened 2023-09-07
Pre hash columns

thorfour:
This adds the ability to specify a column should be prehashed before inserted into the table. 

This allows to aggregator to skip performing the HashArray function for already prehashed columns. This comes with a small alloc cost for aggregations on non pre-hashed data.

```
Aggregation/sum-10       867µs ± 1%     874µs ± 2%    ~     (p=0.075 n=10+10)
Aggregation/count-10     869µs ± 1%     872µs ± 1%    ~     (p=0.218 n=10+10)
Aggregation/max-10       871µs ± 1%     876µs ± 1%  +0.62%  (p=0.003 n=10+9)

Aggregation/sum-10      1.16MB ± 0%    1.16MB ± 0%  +0.22%  (p=0.000 n=10+10)
Aggregation/count-10    1.16MB ± 0%    1.16MB ± 0%  +0.22%  (p=0.000 n=8+10)
Aggregation/max-10      1.16MB ± 0%    1.16MB ± 0%  +0.23%  (p=0.000 n=10+10)

Aggregation/sum-10         825 ± 0%       844 ± 0%  +2.30%  (p=0.000 n=9+10)
Aggregation/count-10       824 ± 0%       843 ± 0%  +2.31%  (p=0.000 n=10+9)
Aggregation/max-10         824 ± 0%       844 ± 0%  +2.37%  (p=0.000 n=10+9)
```


thorfour:
Comparison on large aggregations 
```
name              old time/op    new time/op    delta
_PreAggregate-10     300ms ± 2%     289ms ± 4%  -3.62%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op   delta
_PreAggregate-10     603MB ± 0%     612MB ± 0%  +1.37%  (p=0.000 n=10+8)

name              old allocs/op  new allocs/op  delta
_PreAggregate-10     5.80M ± 0%     5.28M ± 0%  -9.02%  (p=0.000 n=10+10)
```

thorfour:
Comparison when the data is prehashd
```
Aggregation/sum-10       379µs ± 0%     299µs ± 3%  -21.04%  (p=0.000 n=9+10)
Aggregation/count-10     374µs ± 1%     293µs ± 2%  -21.81%  (p=0.000 n=9+10)
Aggregation/max-10       384µs ± 1%     301µs ± 0%  -21.69%  (p=0.000 n=9+9)

Aggregation/sum-10       400kB ± 0%     401kB ± 0%   +0.06%  (p=0.000 n=9+9)
Aggregation/count-10     401kB ± 0%     401kB ± 0%   +0.06%  (p=0.000 n=10+10)
Aggregation/max-10       400kB ± 0%     401kB ± 0%   +0.06%  (p=0.000 n=10+10)

Aggregation/sum-10         686 ± 0%       686 ± 0%     ~     (all equal)
Aggregation/count-10       685 ± 0%       685 ± 0%     ~     (p=0.087 n=10+10)
Aggregation/max-10         686 ± 0%       686 ± 0%     ~     (all equal)
```

thorfour:
Latest benchmark
```
thor@thors-MacBook-Pro frostdb % benchstat BenchmarkAggregation BenchmarkAggregation-prehashed
Aggregation/sum-10       406µs ± 1%     338µs ± 1%  -16.55%  (p=0.000 n=10+10)
Aggregation/count-10     400µs ± 1%     332µs ± 0%  -16.97%  (p=0.000 n=9+9)
Aggregation/max-10       409µs ± 0%     342µs ± 0%  -16.27%  (p=0.000 n=10+10)

Aggregation/sum-10       426kB ± 0%     433kB ± 0%   +1.68%  (p=0.000 n=10+7)
Aggregation/count-10     424kB ± 0%     432kB ± 0%   +1.82%  (p=0.000 n=10+10)
Aggregation/max-10       426kB ± 0%     434kB ± 0%   +1.70%  (p=0.000 n=8+10)

Aggregation/sum-10         788 ± 0%       853 ± 0%   +8.25%  (p=0.000 n=10+10)
Aggregation/count-10       787 ± 0%       853 ± 0%   +8.39%  (p=0.000 n=10+10)
Aggregation/max-10         788 ± 0%       853 ± 0%   +8.25%  (p=0.000 n=10+10)
```

Got a small bit worse due to some bug fixes

db_test.go
```diff
@@ -1986,3 +1992,104 @@ func Test_DB_SnapshotOnClose(t *testing.T) {
 	}))
 	require.True(t, found)
 }
+
+func Test_DB_PrehashedStorage(t *testing.T) {
```
asubiotto, line 1996: Can this be a logic test?

proto/frostdb/schema/v1alpha1/schema.proto
```diff
@@ -20,6 +20,8 @@ message Column {
     StorageLayout storage_layout = 2;
     // Whether the column can dynamically expand.
     bool dynamic = 3;
+    // Prehash the column before storing it. This is an optimization to speed up aggregation queries when this column is often aggregated.
```
asubiotto, line 23: nit: specify where the prehash is stored (i.e. a separate non-dynamic column)

query/logicalplan/optimize.go
```diff
@@ -8,12 +8,18 @@ type Optimizer interface {
...
-	&PhysicalProjectionPushDown{},
-	&FilterPushDown{},
-	&DistinctPushDown{},
-	&ProjectionPushDown{},
+func DefaultOptimizers() []Optimizer {
+	return []Optimizer{
+		&AverageAggregationPushDown{},
+		&PhysicalProjectionPushDown{
+			defaultProjections: []Expr{
+				RegExpNotColumnMatch("^hashed."),
```
asubiotto, line 16: Is the cost of running this extra projection what you measured?

query/physicalplan/aggregate.go
```diff
@@ -424,6 +324,13 @@ func (a *HashAggregate) Callback(ctx context.Context, r arrow.Record) error {
 				groupByFields = append(groupByFields, field)
 				groupByArrays = append(groupByArrays, r.Column(i))
 
+				if a.finalStage { // in the final stage expect the hashes to already exist, so only need to combine them as normal hashes
+					groupByFieldHashes = append(groupByFieldHashes,
+						&uint64HashCombine{value: scalar.Hash(a.hashSeed, scalar.NewStringScalar(field.Name))},
```
asubiotto, line 329: My concern here is that won't the seeds across nodes/restarts be unstable and therefore the hashes won't match? Might be missing something.

query/logicalplan/optimize.go
thorfour: The measurements are prior to this change and then against this change with prehashed columns.

For schemas without prehashed columns this projection shouldn't result in much of anything

thorfour:
```
Aggregation/sum-10       406µs ± 1%     403µs ± 1%  -0.53%  (p=0.008 n=10+9)
Aggregation/count-10     400µs ± 1%     397µs ± 0%  -0.59%  (p=0.000 n=9+10)
Aggregation/max-10       409µs ± 0%     407µs ± 1%  -0.45%  (p=0.001 n=10+10)

Aggregation/sum-10       426kB ± 0%     426kB ± 0%    ~     (p=0.190 n=10+10)
Aggregation/count-10     424kB ± 0%     425kB ± 0%    ~     (p=0.536 n=10+7)
Aggregation/max-10       426kB ± 0%     426kB ± 0%    ~     (p=0.122 n=8+10)

Aggregation/sum-10         788 ± 0%       788 ± 0%    ~     (all equal)
Aggregation/count-10       787 ± 0%       787 ± 0%    ~     (all equal)
Aggregation/max-10         788 ± 0%       788 ± 0%    ~     (all equal)
```

Here's the benchmark run before and after this change without modifications to the benchmark

asubiotto: Do we have measurements of a vanilla selection with and without this projection? I agree with you though, I don't think it should be that high overhead given it's per-record

query/physicalplan/aggregate.go
thorfour: This seed is just for the hash combiner. Which yes will be inconsistent between restarts, but consistent for any given query so it isn't a problem since it'll combine them in the same way at each stage.

query/logicalplan/optimize.go
thorfour: I don't have one no

proto/frostdb/schema/v1alpha1/schema.proto
thorfour: done.

query/physicalplan/aggregate.go
thorfour: Technically the final stage doesn't even need to use the hashcombiner since it could just re-use the hashes from the previous stages. But I didn't see any performance gains when I tried this.

thorfour: However with a distributed query engine it could be problematic to generate different combined hashes on each instance

asubiotto: Ahh gotcha. So since only the final stage generates a seed this is not an issue since the distributor will be the only one to do so.

asubiotto (approved):
LGTM but I'd like you to confirm my last doubt on the seed

query/physicalplan/aggregate.go
thorfour: We inject a single seed for the whole query. But the final stage drops the combined hashes

PR #525, opened 2023-09-07
table/close: snapshot if not persisting

thorfour:


snapshot.go
```diff
@@ -115,6 +115,10 @@ func getTxFromSnapshotFileName(fileName string) (uint64, error) {
 // started (i.e. no other snapshot was in progress). When the snapshot
 // goroutine successfully completes a snapshot, onSuccess is called.
 func (db *DB) asyncSnapshot(ctx context.Context, onSuccess func()) {
+	db.snapshot(ctx, true, onSuccess)
+}
+
+func (db *DB) snapshot(ctx context.Context, async bool, onSuccess func()) {
```
asubiotto, line 121: Could we return an error from snapshot and move the callback to just the async version? Not sure if it's too complex with the txn beginning

db.go
```diff
@@ -841,6 +841,17 @@ func (db *DB) Close(options ...CloseOption) error {
 	}
 	level.Info(db.logger).Log("msg", "closed all tables")
 
+	if !shouldPersist && db.columnStore.snapshotTriggerSize != 0 {
+		db.snapshot(context.Background(), false, func() {
```
asubiotto, line 845: I think it would be nice to info log the duration this took

thorfour: Done.

snapshot.go
thorfour: The refactor gets a little hairy with the tx and metadata. It's gonna require a bigger refactor than I think is worth it right now.

asubiotto (approved):

PR #530, opened 2023-09-13
Project all

thorfour:
Since fixing the projection pushdowns to also apply to in-memory arrow records we no longer have a way to wildcard select all columns.

This adds a new logicalplan Expr that allows projection of all columns.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1986,3 +1986,82 @@ func Test_DB_SnapshotOnClose(t *testing.T) {
 	}))
 	require.True(t, found)
 }
+
+func Test_DB_All(t *testing.T) {
+	config := NewTableConfig(
+		dynparquet.SampleDefinition(),
+	)
+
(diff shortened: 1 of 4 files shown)


db_test.go
```diff
@@ -1986,3 +1986,82 @@ func Test_DB_SnapshotOnClose(t *testing.T) {
...
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	_, err = table.InsertRecord(ctx, r)
+	require.NoError(t, err)
+
+	pool := memory.NewGoAllocator()
+	engine := query.NewEngine(pool, db.TableProvider())
+	err = engine.ScanTable(t.Name()).
+		Project(logicalplan.All()).
```
asubiotto, line 2059: Is there no way for this to be implicit?

asubiotto (approved):

PR #532, opened 2023-09-15
dynparquet: add Size method

asubiotto:


asubiotto:
Merging since this shouldn't be controversial.

PR #533, opened 2023-09-18
db: start compactor pool before replaying WAL

asubiotto:
Previously, the compactor pool was not running during WAL replay. This could cause memory blow-ups on the first compactions after a WAL replay since there would be a lot of l0 parts processed at the same time.

Starting the compactor pool before WAL replay causes WAL replay to slow down a bit but with the good property that data is compacted during WAL replay and memory spikes are avoided. Also, recovery code already expected the compactor pool to be started.

This commit also removes the CompactAfterRecovery option given it is not needed any more.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -29,9 +29,6 @@ type CompactionConfig struct {
 	concurrency          int
 	interval             time.Duration
 	l1ToGranuleSizeRatio float64
-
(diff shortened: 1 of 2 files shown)


asubiotto:
Tested this with an oqton wal and replay time went from 39 to 49s. Interestingly, queries were a lot faster on data that was compacted during WAL replay.

brancz (approved):

PR #543, opened 2023-09-21
LSM index design proposal

thorfour:


brancz (approved):

design/lsm.md
```diff
@@ -0,0 +1,50 @@
...
+| Parquet |--->| Parquet |--->| Parquet |   L1: Compacted L0 in Parquet files.
+|         |    |         |    |         |
+|_________|    |_________|    |_________|
+
+                Fig 1.
+
+
+### Compaction
+
+Compaction would work much like it does today except instead of compacting at the Granule level it takes all of L0 and compacts those records into a Parquet file (potentially in the future another larger arrow record).
```
asubiotto, line 45: Again, kind of a moot point since we shove all records/row groups into the query engine, but the big difference here is that the memtable/l0 will be much larger than what we're currently used to in a single compaction given it will be global. How will this affect memory usage?

Additionally, this compaction will also need to merge memtable/l0 with any globally overlapping l1 parts.

design/lsm.md
```diff
@@ -0,0 +1,50 @@
...
+## Proposal
+
+Instead of using a Btree where we sort the individual rows of each record into granules, we instead use an LSM tree where the entire record is simply appended to the first level
+of the LSM index. Because of the immutable nature of the data in FrostDB we do not require all the features of an LSM trree. But are instead lifting the ideas of having different
+levels of compacted data. With the first level being a raw list of arrow records that have been written, and subsequent levels being compacted records.
+
+
+|---------|    |---------|    |---------|
+|         |    |         |    |         |
+|  Record |--->|  Record |--->|  Record |   L0: Raw written Arrow records
```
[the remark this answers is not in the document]
thorfour: That's fair. I didn't because we're not really implementing an LSM but I think it has enough parallels that it's fine

design/lsm.md
thorfour: I don't think the memotables is much larger it would just be a compaction setting that we can tune to adjust our tradeoffs of memory vs CPU much like compaction today

thorfour: Also this compaction won't merge across level boundaries.

asubiotto: > I don't think the memotables is much larger it would just be a compaction setting that we can tune to adjust our tradeoffs of memory vs CPU much like compaction today

But currently we only look at the l0 parts of a single granule for a given compaction. The memtable will be equivalent to compacting all l0 parts across all granules at the same time. I agree this would be something to tune, but I think it's important to keep in mind how compactions change.

> Also this compaction won't merge across level boundaries.

OK. Maybe this is a good thing, but again we'll need to change this if we ever move towards more of a strict LSM implementation since with this you're forced to do a linear scan over the whole level. Again, we already do this so I don't think it's an issue. This might be nice to also reduce compaction memory usage. Also if we can guarantee l0 won't overlap by maybe throwing away dynamic columns as sorting columns that'll be even better.

thorfour: Agreed. Definitely something to revisit in the future should we change the querying

design/lsm.md
```diff
@@ -0,0 +1,50 @@
...
+## Abstract
+
+FrostDB was built using a BTree as it's index. This decision was made based on early assumptions and before the query engine
+had taken shape. Now with a better understanding of how FrostDB works in practice it is my recommendation that we replace the Btree index with
+an LSM index.
+
+## Problem
+
+Insertion spends the largest portion of CPU cycles on sorting the individual rows of records into their corresponding granules. However due to the multi-key primary key 
+schemas coupled with dynamic columns, the query engine does not leverage the sorted nature of rows during query time. The scan layer simply passes the records into the query engine.
```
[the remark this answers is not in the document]
asubiotto: > I'm not proposing we have a full LSM. I'm only proposing we structure our index in a similar fashion.

OK, this makes a lot more sense. Could you make this clearer in the design doc and give a quick overview of how it differs from a traditional LSM?

PR #544, opened 2023-09-21
remove InsertBuffer

thorfour:
Insertion via Parquet files is now officially unsupported

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -14,25 +14,6 @@ import (
 	"github.com/polarsignals/frostdb/parts"
 )
 
-// insertSamples is a helper function to insert a deterministic sample with a
-// given timestamp. Note that rows inserted should be sorted by timestamp since
-// it is a sorting column.
-func insertSamples(ctx context.Context, t *testing.T, table *Table, timestamps ...int64) uint64 {
-	t.Helper()
-	samples := make([]any, 0, len(timestamps))
-	for _, ts := range timestamps {
-		samples = append(samples, dynparquet.Sample{
-			Labels: []dynparquet.Label{
-				{Name: "label1", Value: "value1"},
-			},
-			Timestamp: ts,
-		})
-	}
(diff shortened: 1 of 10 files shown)


brancz (approved):
🥳

logictest/runner.go
```diff
@@ -255,24 +254,18 @@ func (r *Runner) handleInsert(ctx context.Context, c *datadriven.TestData, arrow
 
 	buf.Sort()
 
-	if arrow {
-		converter := pqarrow.NewParquetConverter(memory.NewGoAllocator(), logicalplan.IterOptions{})
-		defer converter.Close()
+	converter := pqarrow.NewParquetConverter(memory.NewGoAllocator(), logicalplan.IterOptions{})
```
asubiotto, line 257: We're still creating records from parquet buffers here. Maybe change this to construct arrow records directly in this PR or leave that as a TODO.

compaction_test.go
```diff
@@ -329,9 +310,7 @@ func TestCompaction(t *testing.T) {
 							"tests assume only a single granule as input",
 						)
 						cfg := table.db.columnStore.compactionConfig
-						if asArrow {
-							cfg.l1ToGranuleSizeRatio = 0.6 // use a different ratio for arrow records
-						}
+						cfg.l1ToGranuleSizeRatio = 0.6 // use a different ratio for arrow records
```
asubiotto, line 313: nit: is the comment still relevant?

asubiotto (approved):
Very nice to see this. Could you create an issue to revisit the `ParquetConverter` usages to see if we can create records directly?

logictest/runner.go
thorfour: Good call. I'll add an issue and leave it as todo

compaction_test.go
thorfour: Kind of? The test needs that ratio but it isn't using a different one since now there's only the one

PR #553, opened 2023-10-06
dynparquet: add unique primary index option

asubiotto:
This commit adds the option to specify that the primary index should be unique. This results in dropping duplicate rows during compaction. Tables with this option might experience slower compactions.


dynparquet/schema.go
```diff
@@ -1231,27 +1235,38 @@ func (s *Schema) NewWriter(w io.Writer, dynamicColumns map[string][]string) (*pa
...
 			serializeDynamicColumns(dynamicColumns),
 		),
-		parquet.SortingWriterConfig(
-			parquet.SortingColumns(cols...),
-		),
-	), nil
+	}
+	sortingOptions := []parquet.SortingOption{parquet.SortingColumns(cols...)}
+	if s.uniquePrimaryIndex {
+		sortingOptions = append(sortingOptions, parquet.DropDuplicatedRows(true))
```
thorfour, line 1249: Nit Is the conditional necessary? Could just be `parquet.DropDuplicatedRows(s.uniquePrimaryIndex)`

thorfour (approved):

PR #557, opened 2023-10-10
lsm: hold read lock on iterate

asubiotto:
Otherwise, compaction races with snapshots.

### DIFF
--- a/index/lsm.go
+++ b/index/lsm.go
@@ -193,6 +193,8 @@ func (l *LSM) Prefixes(ctx context.Context, prefix string) ([]string, error) {
 }
 
 func (l *LSM) Iterate(iter func(node *Node) bool) {
+	l.RLock()
+	defer l.RUnlock()
 	l.levels.Iterate(iter)
 }
 
--- a/store.go
+++ b/store.go
@@ -48,7 +48,10 @@ func (t *TableBlock) Persist() error {
 		}
 
 		if err != nil {
-			return fmt.Errorf("failed to serialize block: %v", err)
+			if deleteErr := sink.Delete(context.Background(), fileName); deleteErr != nil {
+				err = fmt.Errorf("%v failed to delete file on error: %w", err, deleteErr)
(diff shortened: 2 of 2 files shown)


table.go
```diff
@@ -1113,6 +1113,12 @@ func (t *Table) externalParquetCompaction(writer io.Writer) func(compact []*part
 // compactParts will compact the given parts into a Parquet file written to w.
 // It returns the size in bytes of the compacted parts.
 func (t *Table) compactParts(w io.Writer, compact []*parts.Part) (int64, error) {
+	block, done, err := t.ActiveWriteBlock()
```
asubiotto, line 1116: It seems like the issue here is that we close a table before persisting to object storage, which means that this `ActiveWriteBlock` will fail with `ErrTableClosing` when persisting on close. How did we get around this before?

asubiotto: I think the answer is that we had some compaction-like code that did not acquire the write lock on the block

thorfour: Didn't it acquire the read lock instead? I'd have to look again, but I'm pretty sure that's the piece I overlooked

asubiotto: No, it was a write block (I'm pretty sure it has to right?) 
https://github.com/polarsignals/frostdb/blob/c9100f2ac9c7aca111e1362be4a8a67b85b6b44b/compaction.go#L154

The compaction on serialization was different code: https://github.com/polarsignals/frostdb/blob/cbeaf0d100edc9476bb27b013cbd023f02c78ced/table.go#L1172

asubiotto: Updated this PR to fall back to the previous behavior on `ErrTableClosing` but I'm curious: should we be acquiring a write block earlier on (e.g. in merge) since we'll be updating the index?

asubiotto:
OK updated again with the approach we commented live (holding a read lock in `Iterate`)

thorfour (approved):
LGTM!

PR #560, opened 2023-10-11
*: do not use SortingWriter for compaction

asubiotto:
This commit removes SortingWriter for compaction, given it increases our memory usage. Since we now rely on l0 parts being sorted, we can use a more efficient non-sorting writer for compaction.

On a prod wal replay:
```
benchstat benchmain_mem benchnew_mem
goos: darwin
goarch: arm64
          │ benchmain_mem │           benchnew_mem            │
          │    sec/op     │   sec/op    vs base               │
Replay-12      34.99 ± 0%   23.21 ± 1%  -33.66% (p=0.002 n=6)

          │ benchmain_mem │            benchnew_mem             │
          │     B/op      │     B/op      vs base               │
Replay-12    50.54Gi ± 2%   42.07Gi ± 4%  -16.77% (p=0.002 n=6)

          │ benchmain_mem │           benchnew_mem            │
          │   allocs/op   │  allocs/op   vs base              │
Replay-12     131.9M ± 1%   125.8M ± 1%  -4.66% (p=0.002 n=6)
```

### DIFF
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -1267,47 +1267,6 @@ type ParquetWriter interface {
 	Reset(writer io.Writer)
 }
 
-func (s *Schema) NewSortingWriter(w io.Writer, dynamicColumns map[string][]string) (ParquetWriter, error) {
-	ps, err := s.GetDynamicParquetSchema(dynamicColumns)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


metalmatze (approved):
Sweet.

PR #566, opened 2023-10-19
*: trigger snapshots based on uncompressed inserts

asubiotto:
This commit changes the WithSnapshotTriggerSize option to specify a size in bytes of uncompressed inserts that will trigger a snapshot of the whole database, rather than the active block size, which was the previous trigger. Active memory *compressed* data, while snapshots are now triggered based on the *uncompressed* data inserted into the database. The reason this choice was made is that if a database instance crashes, it is forced to reread all uncompressed inserts since the last snapshot from the WAL, which could potentially lead to unrecoverable OOMs on startup. Defining the snapshot trigger in terms of uncompressed bytes limits the memory usage on recovery to at most the snapshot trigger size (as long as snapshots were successful).


brancz (approved):
Nice!

PR #568, opened 2023-10-23
Add constant value to make the code more readable

elliotchenzichang:
Add constant value to make the code more readable. I believe it should be more elegant. 
Could you help review it when you have time? Thanks!

### DIFF
--- a/db.go
+++ b/db.go
@@ -37,11 +37,11 @@ import (
 )
 
 const (
-	B  = 1
-	KB = 1024 * B
-	MB = 1024 * KB
-	GB = 1024 * MB
-	TB = 1024 * GB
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -36,6 +36,14 @@ import (
 	"github.com/polarsignals/frostdb/wal"
 )
 
+const (
+	B  = 1
+	KB = 1024 * B
```
thorfour, line 41: I think this is a great change,  could use change it to use the units here of KiB, MiB etc.

elliotchenzichang: Sure. Updated.

metalmatze (approved):
Very neat!

PR #577, opened 2023-10-26
Fix: Data not found on rotate

thorfour:
Hedge against a scenario where the new active block and the preivously uploaded block have the same timestamp in their ulid. This caused queries to discard the uploaded block erroniously.

### DIFF
--- a/table.go
+++ b/table.go
@@ -497,6 +497,7 @@ func (t *Table) RotateBlock(ctx context.Context, block *TableBlock, skipPersist 
 
 	id := generateULID()
 	for id.Time() == block.ulid.Time() { // Ensure the new block has a different timestamp.
+		runtime.Gosched()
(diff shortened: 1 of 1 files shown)


table.go
```diff
@@ -496,6 +496,9 @@ func (t *Table) RotateBlock(ctx context.Context, block *TableBlock, skipPersist
 	defer commit()
 
 	id := generateULID()
+	for id.Time() == block.ulid.Time() { // Ensure the new block has a different timestamp.
+		id = generateULID()
```
asubiotto, line 500: I like to put a `runtime.Gosched` in these loops to yield the cpu. Up to you

asubiotto (approved):

table.go
thorfour: good call

PR #578, opened 2023-10-31
moved parquet-tool to use cobra cli

thorfour:
Added find command to search for values in parquet files

image


brancz:
How come cobra if everywhere else we use kong?

PR #582, opened 2023-11-03
db: add GetDB and DropDB

asubiotto:
This allows checking whether a DB exists, and dropping a DB.

### DIFF
--- a/db.go
+++ b/db.go
@@ -599,6 +599,30 @@ func (s *ColumnStore) DB(ctx context.Context, name string, opts ...DBOption) (*D
 	return db, nil
 }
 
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -599,6 +599,30 @@ func (s *ColumnStore) DB(ctx context.Context, name string, opts ...DBOption) (*D
 	return db, nil
 }
 
+func (s *ColumnStore) GetDB(name string) (*DB, error) {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+	db, ok := s.dbs[name]
```
thorfour, line 605: What about read only dbs, does that need to be considered here?

thorfour (approved):
One question but seems fine

PR #599, opened 2023-11-28
pqarrow: change back to row-by-row conversion but write directly

asubiotto:
Previously, we were building a scratchRows buffer column by column before writing to a writer. Profiles show that this buffer is not very efficiently reused, most likely since its scope is limited to a single compaction. These allocations end up being 7% of allocations on prod. This commit changes the arrow->parquet conversion back to row-by-row but avoids allocations. Benchmarks show that the runtime is not affected that much:
```
goos: darwin
goarch: arm64
pkg: github.com/polarsignals/frostdb/pqarrow
                 │ benchmainnoscratch │           benchnew           │
                 │       sec/op       │   sec/op     vs base         │
RecordsToFile-12          278.9µ ± 3%   278.5µ ± 1%  ~ (p=0.818 n=6)

                 │ benchmainnoscratch │              benchnew               │
                 │        B/op        │     B/op      vs base               │
RecordsToFile-12        350.53Ki ± 0%   62.81Ki ± 0%  -82.08% (p=0.002 n=6)

                 │ benchmainnoscratch │              benchnew              │
                 │     allocs/op      │  allocs/op   vs base               │
RecordsToFile-12          4.873k ± 0%   3.850k ± 0%  -20.99% (p=0.002 n=6)
```

### DIFF
--- a/pqarrow/parquet.go
+++ b/pqarrow/parquet.go
@@ -131,6 +131,14 @@ func recordToRows(w dynparquet.ParquetWriter, dcv dynamicColumnVerifier, record 
 				default:
 					row = append(row, parquet.ValueOf(dict.GetOneForMarshal(vidx)).Level(0, def, j))
 				}
+			case *array.Int32:
+				row = append(
+					row, parquet.Int32Value(arr.Value(indexIntoRecord)).Level(0, def, j),
+				)
+			case *array.Int64:
+				row = append(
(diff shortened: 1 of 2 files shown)


thorfour (approved):
LGTM

asubiotto:
Also added another commit to avoid calling `GetOneForMarshal` on int32/64 types since that allocates

PR #613, opened 2023-12-06
Make  Expr implement  fmt.Stringer interface

### LINKED ISSUE #521: Replace `Name()` with `fmt.Stringer` in `Expr`
It's kind of annoying to print an `Expr` because you need to call `Name()`. Expr should just implement the fmt.Stringer interface instead.

gernest:
Closes #521

### DIFF
--- a/query/logicalplan/builder.go
+++ b/query/logicalplan/builder.go
@@ -1,6 +1,8 @@
 package logicalplan
 
 import (
+	"fmt"
+
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/parquet-go/parquet-go"
 )
@@ -60,6 +62,7 @@ type Expr interface {
 	DataType(*parquet.Schema) (arrow.DataType, error)
 	Accept(Visitor) bool
 	Name() string
+	fmt.Stringer
 
 	// ColumnsUsedExprs extracts all the expressions that are used that cause
 	// physical data to be read from a column.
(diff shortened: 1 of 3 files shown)


gernest:
@brancz I had the same impression as you, since there was little details on the ticket I had to start somewhere.

`.String` can just alias `.Name` until the `Expr` implementation wants to display something else, however I don't mind it to stay this way because there is no such use case in the library. Having redundant methods that does the same thing seems extreme though.

One case might me `.Col().Name()` instead of `.Col().String()` which means we can just have `(*Column).Name` .

Either way, I'm  cool with whatever you guys thinks will be Ok. I just need to get familiar with the library.

thorfour:
Yea I should have done more diligence into how the `Name()` function was utilized for the ticket details, so sorry about the lack of clarity in there, I was mostly just sick of writing `fmt.Println(expr.Name())`. 

I think we can have the Expr support both `fmt.Stringer` and `Name() string` because as @brancz said, right now it's mostly coincidence that they are the same.

gernest:
@thorfour thanks for the feedback. I embedded `fmt.Stringer` in `Expr` and made changes for implementations to satisfy it.

For now we only alias `(Expr).Name`  PTAL and let me know if we need to render something else.

thorfour (approved):
Great! Thanks!

thorfour:
@gernest looks like it just needs a rebase and a small lint fix now that the expr has a stringer

gernest:
@thorfour  rebased, I can't see the lint message because of the workflow constraint

> 2 workflows awaiting approval

thorfour:
It just wants to use the new `String()` implementation here: 
image

gernest:
@thorfour I addressed all comments PTAL.

PR #617, opened 2023-12-07
Use buffered channels in synchronizer

### LINKED ISSUE #210: Use buffered channels in synchronizer
> Instead of using a mutex, I would use a buffered channel here (maybe with `concurrency` slots) with a synchronizer goroutine reading from the channel and pushing. The nice thing with a buffered channel is that upstream goroutines can push data to be read even if the downstream goroutine is not ready to read this data yet (e.g. so the upstream goroutine can perform more scan work concurrently), while still being able to apply backpressure if the downstream goroutine is super slow for some reason.

_Originally posted by @asubiotto in https://github.com/polarsignals/frostdb/pull/202#discussion_r976288777_

gernest:
Closes #210

We use a background goroutine for synchronizing the physical plan. This adds a requirement to ensure the passed `context.Context` is cancelled to avoid goroutine leaking calling `(*Synchronizer).Close` safely terminates the background goroutine..

### DIFF
--- a/query/engine.go
+++ b/query/engine.go
@@ -132,7 +132,8 @@ func (b LocalQueryBuilder) Project(
 func (b LocalQueryBuilder) Execute(ctx context.Context, callback func(ctx context.Context, r arrow.Record) error) error {
 	ctx, span := b.tracer.Start(ctx, "LocalQueryBuilder/Execute")
 	defer span.End()
-
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	phyPlan, err := b.buildPhysical(ctx)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


gernest:
There is elusive data race . Theoretically we only perform ops on the underlying physical plan in the same goroutine, yet we have a race here!

thorfour:
While I do blindly believe that this is indeed better, it would be good to see some before and after benchmarks run against this to prove that.

gernest:
Complexity introduced by the changes outweighs whatever gains will arise from this.

metalmatze:
Thank you for exploring this nonetheless!

PR #618, opened 2023-12-08
Add support for uint64 columns

metalmatze:
It would be nice to be able to store uint64.


metalmatze:
I had the feeling that this wouldn't be everything. It's all for ingesting the data, however, for querying we need to add the `scalar.Uint64` type to the `BinaryScalarOperation`.

thorfour:
Should we add support for ingesting something we can't query? I feel like this PR should be updated with the ability to query it too, wdyt?

metalmatze:
Oh, yes. That's what I meant with my above statement. 
I was adding this in the last hours and just pushed those changes.

asubiotto:
I think I tried to do this a while back but ran into the issue that parquet only supports int64: https://github.com/polarsignals/frostdb/pull/516. Is this no longer an issue?

metalmatze:
Very interesting. I didn't see that. Thanks for pointing out. 
During the tests, this worked fine.

Maybe we'll support uint64 via int64 casting? :thinking: 
It would be nice if FrostDB could abstract so the type cast would be transparent to FrostDB users.

metalmatze:
I can confirm (using our parquet tool to look at the blocks) that the underlying datatype in the parquet file is indeed an `int64`. 
For FrostDB users that might be fine, given that the schema can type-hint to FrostDB to convert the data to an `uint64`. If the data is read outside of FrostDB that might be a different story...

metalmatze:
Interesting. This page on [LogicalTypes](https://parquet.apache.org/docs/file-format/types/logicaltypes/) makes it sound like Parquet is potentially indirectly supporting unsigned integers via the metadata. I'm trying to find more information on that.

gernest (commented):
From my work on `parquet-go`  types supported are grouped into `physical` and `converted` . Physical types are `int32|int64|int96|float32|float64|ByteArray|FixedLenByteArray` 

Converted types are pretty much anything that can be converted to physical, but default ones are 

- `UTF8` -> `ByteArray` these are strings
- `Uint8` -> `int32` physically it is `int32` but logically we use it as `uint8` because it firts into the next smallest physical integer
- `Uint64` -> `int64` 
-  ... etc

To answer  @asubiotto 

>I think I tried to do this a while back but ran into the issue that parquet only supports int64: https://github.com/polarsignals/frostdb/pull/516. Is this no longer an issue?

Yes, as far as I understand it is not an issue. While consuming parquet, conversion is done transparently. You can safely store and consume a `uint64` the same way you do a `int64`

I will wait to review this, because I have the same opinion as  @thorfour 

> Should we add support for ingesting something we can't query? I feel like this PR should be updated with the ability to query it too, wdyt?

If we add new types to schema, we should be able to ingest and query them. I don't mind the ingestion and querying to be done separately but it matters to track them both( as to be upfront if this PR is only for ingestion so a folloup querying can be expected)

For the arrow counterpart I think there is all matching supported types for both physical and logical parquet types.

metalmatze:
@achille-roussel pointed out on Gopher Slack in the #parquet-go channel that this is where the underlying conversion is happening: https://github.com/parquet-go/parquet-go/blob/bec192ac8b062cde58c0b2e25757796bc60cf63b/value.go#L172-L215

I'm going to rebase this to move forward with uint64 support. 
We definitely want to also be able to query uint64s!

metalmatze:
This is ready for another round of reviews.

thorfour (approved):
LGTM

Issue #587, opened 2023-11-08
frostdb aggregation function is very strong，but aggregation function only support arrow.INT64, can we support arrow.FLOAT64?


jicanghaixb:
frostdb schema can support column type StorageLayout_TYPE_DOUBLE, but aggregation function can't apply it, can we support it ?

asubiotto:
There's no reason why not. We won't be able to get to this anytime soon but will be happy to review a PR.

jicanghaixb:
thank you for your quick response, aggregation functions is one of the most commonly used abstractions data analysis, float64 is more common than int64 in the field of data analysis, so if  aggregation function can support float64,  frostdb can adapt to a wider range of application scenarios.

PR #619, opened 2023-12-08
feat: initial work for aggregations with float64

garrensmith:
This is very much a work in progress and could also be all wrong. 
This is an attempt at adding support for float64 aggregations. At this point, the aggregation works but the values are all wrong. Somehow the conversion of the value to float is wrong and I can't quite see why that is. I've been using the `examples/aggregation.go` to test it and it gives results of `[1.5e-322 1.5e-322 2e-322]` so I've got something wrong. 

If you could point me in the correct direction I could make the changes, or if I'm completely wrong let me know a better way. 

Fixes https://github.com/polarsignals/frostdb/issues/587

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -97,30 +97,15 @@ func Aggregate(
 
 func chooseAggregationFunction(
 	aggFunc logicalplan.AggFunc,
-	dataType arrow.DataType,
+	_ arrow.DataType,
 ) (AggregationFunction, error) {
 	switch aggFunc {
 	case logicalplan.AggFuncSum:
-		switch dataType.ID() {
-		case arrow.INT64, arrow.FLOAT64:
-			return &SumAggregation{}, nil
-		default:
-			return nil, fmt.Errorf("unsupported sum of type: %s", dataType.Name())
-		}
+		return &SumAggregation{}, nil
 	case logicalplan.AggFuncMin:
-		switch dataType.ID() {
-		case arrow.INT64, arrow.FLOAT64:
-			return &MinAggregation{}, nil
-		default:
-			return nil, fmt.Errorf("unsupported min of type: %s", dataType.Name())
-		}
+		return &MinAggregation{}, nil
 	case logicalplan.AggFuncMax:
-		switch dataType.ID() {
-		case arrow.INT64, arrow.FLOAT64:
-			return &MaxAggregation{}, nil
-		default:
-			return nil, fmt.Errorf("unsupported max of type: %s", dataType.Name())
-		}
+		return &MaxAggregation{}, nil
 	case logicalplan.AggFuncCount:
 		return &CountAggregation{}, nil
 	default:
@@ -674,7 +659,7 @@ func (a *HashAggregate) finishAggregate(ctx context.Context, aggIdx int, aggrega
 
 type SumAggregation struct{}
 
-var ErrUnsupportedSumType = errors.New("unsupported type for sum aggregation, expected int64")
+var ErrUnsupportedSumType = errors.New("unsupported type for sum aggregation, expected int64 or float64")
 
 func (a *SumAggregation) Aggregate(pool memory.Allocator, arrs []arrow.Array) (arrow.Array, error) {
 	if len(arrs) == 0 {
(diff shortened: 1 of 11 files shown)


query/physicalplan/aggregate.go
```diff
@@ -717,6 +719,8 @@ func (a *Int64MinAggregation) Aggregate(pool memory.Allocator, arrs []arrow.Arra
 	switch typ {
 	case arrow.INT64:
 		return minInt64arrays(pool, arrs), nil
+	case arrow.FLOAT64:
```
garrensmith, line 722: Instead of creating a `Float64MinAggregation` I changed this one to handle float as well. For the `MaxAggregation` I created a separate struct. I prefer having one function that can all both types but I don't know this code base so I'm happy to change it.

garrensmith: Right. I wanted to check which was the better way.

thorfour:
Okay I took a quick look and it looks correct to me. I ran the example and I think where the breakdown was is the example is ingesting the snowfall in `int64` but the schema says it should be in `DOUBLE` and I don't believe that our converting handles converting types. So if we change the snowfall struct to be of type `float64` it correctly runs the min/max aggregations on it.

query/physicalplan/aggregate.go
metalmatze: Agreed. I was looking at this very same thing earlier for uint64. I think having the switch in the aggregation function is better in the long run. 
Sadly, generics are still not feasible with the arrow library... :confused:

garrensmith:
@thorfour ok great. Thanks for the help. I've implemented all the aggregations and they work via the example. I've tried to add some tests to the `logictest` but got a bit stuck. It's returning all 0 for the answers. So I'm guessing I have the conversion error again in the tests but I'm not quite sure.

thorfour:
@garrensmith Figured it out! You had the `floatvalue` listed at `Nullable: true` which it seems we don't correctly handle in the logictest runner. So we need to add 

```bash
diff --git a/logictest/runner.go b/logictest/runner.go
index e29baa0..294acf3 100644
--- a/logictest/runner.go
+++ b/logictest/runner.go
@@ -221,7 +221,15 @@ func (r *Runner) handleInsert(ctx context.Context, c *datadriven.TestData) (stri
                                if err != nil {
                                        return "", fmt.Errorf("insert: %w", err)
                                }
-                               rows[i] = append(rows[i], parquet.ValueOf(v).Level(0, 0, colIdx))
+                               if col.StorageLayout.Optional() {
+                                       if parquet.ValueOf(v).IsNull() {
+                                               rows[i] = append(rows[i], parquet.ValueOf(v).Level(0, 0, colIdx))
+                                       } else {
+                                               rows[i] = append(rows[i], parquet.ValueOf(v).Level(0, 1, colIdx))
+                                       }
+                               } else {
+                                       rows[i] = append(rows[i], parquet.ValueOf(v).Level(0, 0, colIdx))
+                               }
                                colIdx++
                                continue
                        }
```
                        
To correctly set the definition level for non-dynamic optional columns.

garrensmith:
@thorfour awesome thanks for that. I've added that in. Added some more tests. This should now be ready for a proper review.

query/physicalplan/aggregate.go
```diff
@@ -102,22 +102,22 @@ func chooseAggregationFunction(
 	switch aggFunc {
 	case logicalplan.AggFuncSum:
 		switch dataType.ID() {
-		case arrow.INT64:
-			return &Int64SumAggregation{}, nil
+		case arrow.INT64, arrow.FLOAT64:
```
thorfour, line 105: Since the aggregation functions now handle the type switch we should be ably to simplify this function and remove the redundant type switches here.

query/physicalplan/project.go
```diff
@@ -341,11 +341,20 @@ func (a *averageProjection) Project(mem memory.Allocator, r arrow.Record) ([]arr
...
+			Name: resultName,
+			Type: &arrow.Int64Type{},
+		})
+		columns = append(columns, avgInt64arrays(mem, sums, counts))
+	case arrow.FLOAT64:
+		fields = append(fields, arrow.Field{
+			Name: resultName,
+			Type: &arrow.Float64Type{},
+		})
+		columns = append(columns, avgFloat64arrays(mem, sums, counts))
```
thorfour, line 356: Can we add a `default:` here that will return an error for an unsupported type?

thorfour (commented):
Couple of smol QoL fixes and the linter wants some periods on the end of comments (I hate this linting requirement so much). 

Otherwise this looks ready to go! Thank you so much!

PR #621, opened 2023-12-08
Parquet file compaction

### LINKED ISSUE #605: L1+ disk compaction
It would be useful if we implemented a new compaction function that compacts parts into on disk parts, to reduce memory usage.

thorfour:
This adds an optional compaction function to FrostDB to allow for compaction to write to disk instead of an in-memory buffer. 

It appends compacted parts into an os temporary file, and will truncate the file when the level is compacted into the next level.

I plan to run benchmarks on this but initially ran with these settings on Parca and observed the expected reduction in memory usage, but would expect the query latency to increase as well as compaction times.

Closes #605


parts/file.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	buf, err := dynparquet.NewSerializedBuffer(file)
+	if err != nil {
+		return nil, err
+	}
+
+	rowBuf := &dynparquet.DynamicRows{Rows: make([]parquet.Row, 1)}
```
asubiotto, line 82: nit: I assume the `Least`/`Most` in-mem parquet part code is the same as this. Can we extract this into a `baseParquetPart` or even `basePart`?

table.go
```diff
@@ -1092,22 +1094,59 @@ func (t *Table) close() {
...
 	for _, level := range levels {
-		config = append(config, &index.LevelConfig{
-			Level:   level.Level,
-			MaxSize: level.MaxSize,
-			Compact: t.parquetCompaction, // NOTE: right now we only support parquet compaction but in the future this could compact parts into larger arrow records.
-		})
+		if level.Type == CompactionTypeParquetDisk {
+			fileCompaction = t.parquetFileCompaction(len(levels) - 1) // Create a file compaction for all but the last level.
+			t.closers = append(t.closers, fileCompaction)             // Append to closers so that thr underlying files are closed on table close.
+			break
```
asubiotto, line 1129: nit: is this break necessary?

parts/file.go
```diff
@@ -0,0 +1,162 @@
...
+// fileParquetPart implements the Part interface. Is is backed by a Parquet file.
+type fileParquetPart struct {
+	basePart
+
+	rdr     io.ReaderAt
+	size    int64
+	release func()
+}
+
+func NewFileParquetPart(tx uint64, rdr io.ReaderAt, size int64, release func(), options ...Option) Part {
```
asubiotto, line 22: Below I comment on sharing some code with the parquet part but thinking more about this I think we could just remove this new file part and have the existing parquet part use a disk-backed reader at, right? It's basically up to whoever creates the part to create the reader at, but the part itself is agnostic to the backing storage. I think we just need to add the release functionality.

table.go
thorfour: Yes, otherwise we may accidentally add multiple closers when we only want one.

parts/file.go
thorfour: So initially it made more sense as when I wrote this I had the `fileParquetPart` handling the file(s) themselves. But as it evolved it definitely makes less sense. 

There are some subtle differences between the two still that make me wonder if it's best to merge them. Such as the file part opens the file each time to avoid holding pages in memory, I'm wondering if that behavior makes sense for an already in memory part.

parts/file.go
thorfour: moved to common function

parts/file.go
asubiotto: Hmm, I didn't notice that. Are we sure we want to open the file every time? This also includes re-reading parts of the file that would have already been read by a previous open.

thorfour: I don't know yet. Need more real world data on that. Since the goal was to limit the amount of memory to store parts it made more sense to me to not hold onto any of that memory. 

Since in theory the file will only ever be opened on query and compaction.

table.go
thorfour: However clearly I made this more complicated than it needed to be. I change the `fileCompaction` struct to only handle a single level, and just had the configure lsm function create multiple of them.

parts/file.go
asubiotto: How much memory do we actually use with just keeping an open file around? I would assume variable-sized buffers are only used when reading/decoding. We would already be reducing the memory used just for the backing storage with the in-mem parquet part.

thorfour: You're right, tested it out and there was no noticeable overhead. I've removed the file part and replaced the usage with the normal Parquet part.

asubiotto (approved):
Thanks!

PR #622, opened 2023-12-11
Add Generic RecordBuilder to use with (*Table)InsertRecord

gernest:
I noticed in parca `(*Table)InsertRecord` is used. The `arrow.Record` is  built by hand. Considering the table is flat by nature, this is the part where a generic solution fits better. Hand building records can be error prone and bring long term maintenance burden.

This commit adds `(*Build[T])` that takes a struct and returns an api for appending `T` and retrieving the final `arrow.Record`

Tags are used to describe the schema(This is a common pattern, works very well on parquet-go).

Here is example of Sample schema annotated with tags

```go
type Sample struct {
	ExampleType string      `frostdb:"example_type,rle_dict,asc(0)"`
	Labels      []Label     `frostdb:"labels,rle_dict,null,dyn,asc(1),null_first"`
	Stacktrace  []uuid.UUID `frostdb:"stacktrace,rle_dict,asc(3),null_first"`
	Timestamp   int64       `frostdb:"timestamp,asc(2)"`
	Value       int64       `frostdb:"value"`
}
```

Which is the same as

```go
func SampleDefinition() *schemapb.Schema {
	return &schemapb.Schema{
		Name: "test",
		Columns: []*schemapb.Column{{
			Name: "example_type",
				Type:     schemapb.StorageLayout_TYPE_STRING,
				Encoding: schemapb.StorageLayout_ENCODING_RLE_DICTIONARY,
			},
			Dynamic: false,
		}, {
			Name: "labels",
				Type:     schemapb.StorageLayout_TYPE_STRING,
				Nullable: true,
				Encoding: schemapb.StorageLayout_ENCODING_RLE_DICTIONARY,
			},
			Dynamic: true,
		}, {
			Name: "stacktrace",
				Type:     schemapb.StorageLayout_TYPE_STRING,
				Encoding: schemapb.StorageLayout_ENCODING_RLE_DICTIONARY,
			},
			Dynamic: false,
		}, {
			Name: "timestamp",
				Type: schemapb.StorageLayout_TYPE_INT64,
			},
			Dynamic: false,
		}, {
			Name: "value",
				Type: schemapb.StorageLayout_TYPE_INT64,
			},
			Dynamic: false,
		}},
		SortingColumns: []*schemapb.SortingColumn{{
			Name:      "example_type",
			Direction: schemapb.SortingColumn_DIRECTION_ASCENDING,
		}, {
			Name:       "labels",
			Direction:  schemapb.SortingColumn_DIRECTION_ASCENDING,
			NullsFirst: true,
		}, {
			Name:      "timestamp",
			Direction: schemapb.SortingColumn_DIRECTION_ASCENDING,
		}, {
			Name:       "stacktrace",
			Direction:  schemapb.SortingColumn_DIRECTION_ASCENDING,
			NullsFirst: true,
		}},
	}
}
```

Example usage

```go
		b := NewBuild[Sample](memory.DefaultAllocator)
		defer b.Release()
		samples := NewTestSamples()
		b.Append(samples...)
		r := b.NewRecord()
```

Please see the `Build` godoc comment for more details and limitations.

### DIFF
--- a/dynparquet/record_builder.go
+++ b/dynparquet/record_builder.go
@@ -1,0 +1,597 @@
+package dynparquet
+
+import (
+	"reflect"
+	"slices"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/apache/arrow/go/v14/arrow/array"
+	"github.com/apache/arrow/go/v14/arrow/memory"
+	"github.com/google/uuid"
+	schemapb "github.com/polarsignals/frostdb/gen/proto/go/frostdb/schema/v1alpha1"
+)
+
+const (
+	TagName = "frostdb"
+)
+
+// Build is a generic arrow.Record builder that ingests structs of type T. The
+// generated record can be passed to (*Table).InsertRecord.
+//
+// Struct tag `frostdb` is used to pass options for the schema for T and use
+// (*Build[T]).Schema to obtain schema v1alpha1.
+//
+// This api is opinionated.
+//
+//   - Dynamic columns are supported but only for  maps and struct slices
+//
+//   - Repeated Columns are not supported
+//
+//   - Nested Columns are not supported
+//
+// # Tags
+//
+// Use `frostdb` to define tags that customizes field values. You can express
+// everything needed to construct schema v1alpha1.
+//
+// Tags are defined as a comma separated list. The first item is the column
+// name. Column name is optional, when omitted it is derived from the field name
+// (snake_cased)
+//
+// Supported Tags
+//
+//	    delta_binary_packed | Delta binary packed encoding.
+//	                 brotli | Brotli compression.
+//	                    asc | Sorts in ascending order.Use asc(n) where n is an integer for sorting order.
+//	                   gzip | GZIP compression.
+//	                 snappy | Snappy compression.
+//	delta_length_byte_array | Delta Length Byte Array encoding.
+//	       delta_byte_array | Delta Byte Array encoding.
+//	                   desc | Sorts in descending order.Use desc(n) where n is an integer for sorting order
+//	                lz4_raw | LZ4_RAW compression.
+//	               pre_hash | Prehash the column before storing it.
+//	                    dyn | Whether the column can dynamically expand.
+//	                   null | Nullable column.
+//	             null_first | When used wit asc nulls are smallest and with des nulls are largest.
+//	                   zstd | ZSTD compression.
+//	               rle_dict | Dictionary run-length encoding.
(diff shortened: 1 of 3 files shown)


dynparquet/record_builder.go
```diff
@@ -0,0 +1,597 @@
...
+// generated record can be passed to (*Table).InsertRecord.
+//
+// Struct tag `frostdb` is used to pass options for the schema for T and use
+// (*Build[T]).Schema to obtain schema v1alpha1.
+//
+// This api is opinionated.
+//
+//   - Dynamic columns are supported but only for  maps and struct slices
+//
+//   - Repeated Columns are not supported
```
thorfour, line 31: Is this a fundamental issue or just not something that is supported today?

thorfour (commented):
Wow! Really cool PR love this!

dynparquet/record_builder.go
thorfour: So I asked because `v1alpha1` technically does support the repeated column type using https://github.com/polarsignals/frostdb/blob/bc23e42eda40e0d86d629aa7837e23bfa52f5b2c/proto/frostdb/schema/v1alpha1/schema.proto#L90

gernest: Oops! I missed that. Still, we probably need to keep this PR simple  for reviewing. I will prepare a follow up patch when this lands.

Unless if this is necessary. My focus was to handle https://github.com/parca-dev/parca/blob/main/pkg/profile/schema.go which doesn't have repeated fields.

gernest: @thorfour since you will be the one reviewing, do you think I should add repeated fields here or on a followup PR?

If here, I was wondering how repeated field behaves for dynamic columns ?

thorfour: We can do it in a follow up that's no problem. Just wanted to raise it here so that it was documented.

I think we'd want to slightly change how dynamic columns are handled. I think we should probably require them to be of type `map[string]any` where the key in the map would be the fully qualified dynamic column name. 

So we'd want to change `Labels` in the `Sample` struct into 
```go
type Sample struct {
...
    Labels map[string]string
...
}
```

Which I think is probably better behavior anyways, it was kind of a strange thing to assume that the `Name` field of the struct should exist for a dynamic column.

thorfour:
@gernest I'm happy to merge this if you are.

gernest:
@thorfour  Thanks, I'm happy with the scope and the changes, please merge.

I am currently working on a patch to define dynamic columns as `map[string]T`  where `T` is of base types `int64|bool|float64|string` as per your comment on dynamic columns. Will submit it as a separate PR

PR #624, opened 2023-12-12
Support repeated columns for Generic Record Builder

gernest:
Support repeated columns for all base types `[]int64` `[]float64`  `[]bool` `[]string`.   Use all supported tags on this column like other base columns( set encoding ,compression etc)

Since we deal with slices I took the liberty to make repeated fields nullable by default. 

Example

```go
		type Example struct {
			StringDict []string `frostdb:",rle_dict"`
		}
```


thorfour (approved):
Great thanks!

PR #627, opened 2023-12-13
Non-temporary file compaction support.

thorfour:
If FrostDB is configured with disk compaction and a storage path it will use persistent files instead of temporary files.

Which means during WAL recovery it will recover compacted sections straight from the LSM level files. 

The replay maintains a list of sections that are recorded into the WAL during the first pass. During the second pass when a new table is created the final list of compacted parts are inserted into the index into that table. Any parts that have been compacted into this list are then skipped during replay and only the uncompacted parts are replayed.


proto/frostdb/wal/v1alpha1/wal.proto
```diff
@@ -17,12 +17,27 @@ message Record {
...
   message Write {
+    reserved 3; // Previously used for Arrow indication, but all writes are now Arrow.
     // Table name of the write.
     string table_name = 1;
     // Data is the data of the write.
     bytes data = 2;
-    // Arrow indicates that data is stored in arrow record format.
-    bool arrow = 3;
+    // Tx is the transaction ID of the write.
+    uint64 tx = 4;
```
asubiotto, line 26: I don't think it's necessary to add the transaction ID into the Write message. The transaction ID is already implicitly the WAL index.

asubiotto (commented):
It doesn't feel right to me to write compaction information into the write-ahead log because compactions are internal database reorganizations of existing data. To me, we should only ever write the actual data being written (and actually do this synchronously) before the operation is applied to the DB. I guess we already kind of do this (e.g. persistence) but not sure that belongs in the WAL either.

Apart from this, my main concern is that these WAL writes are done after compaction and they're not atomic, so what happens if we write the compaction, but not the WAL record (i.e. we crash?).

One other thing I'm wondering about with removing snapshots in favor of this is when is the WAL truncated? It seems to me like what we're trying to do here is ignore a set of transactions that are recovered some other way, which is very similar to truncation, just that with truncation we never read these txns in the first place.

I know we talked about deprecating snapshots, but I wonder what you think about this idea: the big problem we have is that compaction acts on a set of transactions per-table while the WAL contains all transactions in a database. So during a compaction, we only know that we can ignore the set of compacted transactions that were applied to a given table, but not all transactions less than the max transaction since there could be earlier transactions applied to a *different* table whose write only exists in the WAL. This requires us to store this metadata somewhere.

The nice thing about snapshots is that we can confidently create a snapshot and truncate the WAL and then recover from a snapshot that has this compacted data without doing complex operations with the WAL. My thinking is that maybe we should consider extending snapshots with file compactions, rather than replacing them. Snapshots could, for example, add a new part proto that points to a filename which is the compacted part and on recovery, simply create the part with a handle to that file. This would avoid having to write complex conditionals based on WAL records (and we also avoid the extra disk writes per compaction).

thorfour:
> It doesn't feel right to me to write compaction information into the write-ahead log because compactions are internal database reorganizations of existing data. To me, we should only ever write the actual data being written (and actually do this synchronously) before the operation is applied to the DB. I guess we already kind of do this (e.g. persistence) but not sure that belongs in the WAL either.

This feels like an arbitrary distinction to me. But originally I wanted to write the compactions to the WAL but that wasn't going to be possible with the way the WAL works today.

> Apart from this, my main concern is that these WAL writes are done after compaction and they're not atomic, so what happens if we write the compaction, but not the WAL record (i.e. we crash?).

This is fine, we simply wouldn't know we compacted, and would replay the writes as normal, and would just overwrite the compaction the next go around.

> One other thing I'm wondering about with removing snapshots in favor of this is when is the WAL truncated? It seems to me like what we're trying to do here is ignore a set of transactions that are recovered some other way, which is very similar to truncation, just that with truncation we never read these txns in the first place.

This was more of a side effect of the change, and I agree it's not ideal.

> I know we talked about deprecating snapshots, but I wonder what you think about this idea: the big problem we have is that compaction acts on a set of transactions per-table while the WAL contains all transactions in a database. So during a compaction, we only know that we can ignore the set of compacted transactions that were applied to a given table, but not all transactions less than the max transaction since there could be earlier transactions applied to a different table whose write only exists in the WAL. This requires us to store this metadata somewhere.
The nice thing about snapshots is that we can confidently create a snapshot and truncate the WAL and then recover from a snapshot that has this compacted data without doing complex operations with the WAL. My thinking is that maybe we should consider extending snapshots with file compactions, rather than replacing them. Snapshots could, for example, add a new part proto that points to a filename which is the compacted part and on recovery, simply create the part with a handle to that file. This would avoid having to write complex conditionals based on WAL records (and we also avoid the extra disk writes per compaction).

This implies we ignore on disk LSM entirely unless we have a snapshot though.

asubiotto:
> > One other thing I'm wondering about with removing snapshots in favor of this is when is the WAL truncated? It seems to me like what we're trying to do here is ignore a set of transactions that are recovered some other way, which is very similar to truncation, just that with truncation we never read these txns in the first place.
> 
> This was more of a side effect of the change, and I agree it's not ideal.
> 

Hmm, I think this is a big problem because it implies we can only ever truncate the WAL once all tables are persisted including small tables that don't grow much (which we do have in production).

> > I know we talked about deprecating snapshots, but I wonder what you think about this idea: the big problem we have is that compaction acts on a set of transactions per-table while the WAL contains all transactions in a database. So during a compaction, we only know that we can ignore the set of compacted transactions that were applied to a given table, but not all transactions less than the max transaction since there could be earlier transactions applied to a different table whose write only exists in the WAL. This requires us to store this metadata somewhere.
> > The nice thing about snapshots is that we can confidently create a snapshot and truncate the WAL and then recover from a snapshot that has this compacted data without doing complex operations with the WAL. My thinking is that maybe we should consider extending snapshots with file compactions, rather than replacing them. Snapshots could, for example, add a new part proto that points to a filename which is the compacted part and on recovery, simply create the part with a handle to that file. This would avoid having to write complex conditionals based on WAL records (and we also avoid the extra disk writes per compaction).
> 
> This implies we ignore on disk LSM entirely unless we have a snapshot though.

I think this is not a big problem in practice. We currently ignore any compacted data if it is not recorded in a snapshot and these on-disk LSM files would just be part of that. We have an upper bound of ingested data before we trigger a snapshot so the "compaction" progress lost also has an upper bound. With your changes, I think snapshots will also become a lot cheaper to take (since we avoid multi-MB copies), so we can even reduce this bound if necessary.

thorfour:
> I think this is not a big problem in practice. We currently ignore any compacted data if it is not recorded in a snapshot and these on-disk LSM files would just be part of that. We have an upper bound of ingested data before we trigger a snapshot so the "compaction" progress lost also has an upper bound. With your changes, I think snapshots will also become a lot cheaper to take (since we avoid multi-MB copies), so we can even reduce this bound if necessary.

Hmm okay I see your point. I think I'm okay with this. Will close this in favor of recording compacted regions in snapshots.

PR #628, opened 2023-12-14
pqarrow/arrowutils: Add SortRecord and ReorderRecord

metalmatze:
This is extract from a previous PR #461.

### DIFF
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -64,7 +64,7 @@ func ReorderRecord(r arrow.Record, indices arrow.Array) (arrow.Record, error) {
 	return res.(*compute.RecordDatum).Value, nil
 }
 
-type orderedArray[T constraints.Ordered] interface {
+type orderedArray[T int64 | float64 | string] interface {
 	Value(int) T
 	IsNull(int) bool
 	Len() int
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -10,7 +10,6 @@ import (
 	"github.com/apache/arrow/go/v14/arrow/array"
 	"github.com/apache/arrow/go/v14/arrow/compute"
 	"github.com/apache/arrow/go/v14/arrow/memory"
-	"golang.org/x/exp/constraints"
 )
 
 // SortRecord sorts the given record's rows by the given column. Currently only supports int64, string and binary columns.
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -70,7 +70,7 @@ type orderedArray[T int64 | float64 | string] interface {
 	Len() int
 }
 
-type orderedSorter[T constraints.Ordered] struct {
+type orderedSorter[T int64 | float64 | string] struct {
 	array   orderedArray[T]
 	indices []int64
 }
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -103,14 +103,11 @@ func (s binarySort) Len() int {
 }
 
 func (s binarySort) Less(i, j int) bool {
-	if s.array.IsNull(int(s.indices[i])) && !s.array.IsNull(int(s.indices[j])) {
+	if s.array.IsNull(int(s.indices[i])) {
 		return false
 	}
-	if !s.array.IsNull(int(s.indices[i])) && s.array.IsNull(int(s.indices[j])) {
+	if s.array.IsNull(int(s.indices[j])) {
 		return true
-	}
-	if s.array.IsNull(int(s.indices[i])) && s.array.IsNull(int(s.indices[j])) {
-		return false
 	}
 	// we need to read the indices from the indices slice, as they might have already been swapped.
 	return bytes.Compare(s.array.Value(int(s.indices[i])), s.array.Value(int(s.indices[j]))) == -1
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -2,33 +2,36 @@ package arrowutils
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"sort"
 
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/apache/arrow/go/v14/arrow/array"
+	"github.com/apache/arrow/go/v14/arrow/compute"
 	"github.com/apache/arrow/go/v14/arrow/memory"
 	"golang.org/x/exp/constraints"
-
-	"github.com/polarsignals/frostdb/pqarrow/builder"
 )
 
 // SortRecord sorts the given record's rows by the given column. Currently only supports int64, string and binary columns.
-func SortRecord(r arrow.Record, cols []int) ([]int, error) {
+func SortRecord(mem memory.Allocator, r arrow.Record, cols []int) (*array.Int64, error) {
 	if len(cols) > 1 {
 		return nil, fmt.Errorf("sorting by multiple columns isn't implemented yet")
 	}
+	indicesBuilder := array.NewInt64Builder(mem)
+
 	if r.NumRows() == 0 {
-		return nil, nil
+		return indicesBuilder.NewInt64Array(), nil
 	}
 	if r.NumRows() == 1 {
-		return []int{0}, nil
+		indicesBuilder.Append(0)
+		return indicesBuilder.NewInt64Array(), nil
 	}
 
-	indices := make([]int, r.NumRows())
+	indices := make([]int64, r.NumRows())
 	// populate indices
 	for i := range indices {
-		indices[i] = i
+		indices[i] = int64(i)
 	}
 
 	switch c := r.Column(cols[0]).(type) {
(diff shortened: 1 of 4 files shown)


asubiotto (commented):
This is a good start. I think what is missing is:
1) Order by a set of columns (e.g. change the `SortRecord` parameter from `col int` to `cols int`)
2) Support ordering directions (ascending vs descending)
3) Support NULLs.

metalmatze:
I 100% agree with your list. 
Would you like us to add those before merging this PR? 
Even with this small feature set some use cases benefit from this already. 

For now, I'll make the parameter a `cols []int` for the future.

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// ReorderRecord reorders the given record's rows by the given indices.
+func ReorderRecord(mem memory.Allocator, r arrow.Record, indices []int) (arrow.Record, error) {
+	// if the indices are already sorted, we can return the original record to save memory allocations
+	if sort.SliceIsSorted(indices, func(i, j int) bool { return indices[i] < indices[j] }) {
+		return r, nil
+	}
+
+	recordBuilder := builder.NewRecordBuilder(mem, r.Schema())
```
gernest, line 55: `compute.Take` supports `arrow.Record` I was wondering why you chose not to use it.

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/apache/arrow/go/v14/arrow/array"
+	"github.com/apache/arrow/go/v14/arrow/memory"
+	"golang.org/x/exp/constraints"
+
+	"github.com/polarsignals/frostdb/pqarrow/builder"
+)
+
+// SortRecord sorts the given record's rows by the given column. Currently only supports int64, string and binary columns.
+func SortRecord(r arrow.Record, cols []int) ([]int, error) {
```
gernest, line 17: Can we also add `direction []int` which will behave similar  `bytes.Compare` does

- `-1` for ascending
- `0` no direction
- `1` for descending

This will simply sorting , in the less function you can return `bytes.Compare(...)==direction`

pqarrow/arrowutils/sort.go
metalmatze: I wasn't aware of it and this is an old version from months ago. We should take the time to refactor it!

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	indicesBuilder.AppendValues(indices, nil)
+	return indicesBuilder.NewInt64Array(), nil
+}
+
+// ReorderRecord reorders the given record's rows by the given indices.
+// This is a wrapper around compute.Take which handles the type castings.
+func ReorderRecord(r arrow.Record, indices arrow.Array) (arrow.Record, error) {
+	res, err := compute.Take(
```
gernest, line 55: Awesome, I also wanted to note, this will work majority of the time. We support dictionaries and there is no dictionary kernel for `Take` yet.

We need to optimise

-  If record has no dictionary column. Use `compute.Take` on  hte record
-  If there is a dictionary column. Use `compute.Take` on individual non dictionary columns and fall back to manual taking for the dictionary and then assembling the record (preferably concurrently).


This can be done on a separate PR. Unless I'm mistaken, the PR just adds these functions but there is no expected call site for them in yet( There will be a need for a lot of changes for this to be used in  my logictest PR). 

I'm on mobile now , so I cant write what is needed for the logictest to use these functions.

metalmatze: Oh, that's nasty and a really good comment! 
We should make sure this is supported in follow-up PRs!

metalmatze:
I've updated to additionally add support for sorting `NULL`. It always gets sorted to the back.

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,124 @@
...
+func (s orderedSorter[T]) Less(i, j int) bool {
+	if s.array.IsNull(int(s.indices[i])) && !s.array.IsNull(int(s.indices[j])) {
+		return false
+	}
+	if !s.array.IsNull(int(s.indices[i])) && s.array.IsNull(int(s.indices[j])) {
+		return true
+	}
+	if s.array.IsNull(int(s.indices[i])) && s.array.IsNull(int(s.indices[j])) {
+		return false
+	}
```
gernest, line 91:
```suggestion
	if s.array.IsNull(int(s.indices[i])) {
		return false
	}
	if s.array.IsNull(int(s.indices[j])) {
		return true
	}
```

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,124 @@
...
+func (s binarySort) Less(i, j int) bool {
+	if s.array.IsNull(int(s.indices[i])) && !s.array.IsNull(int(s.indices[j])) {
+		return false
+	}
+	if !s.array.IsNull(int(s.indices[i])) && s.array.IsNull(int(s.indices[j])) {
+		return true
+	}
+	if s.array.IsNull(int(s.indices[i])) && s.array.IsNull(int(s.indices[j])) {
+		return false
+	}
```
gernest, line 117:
```suggestion
	if s.array.IsNull(int(s.indices[i])) {
		return false
	}
	if s.array.IsNull(int(s.indices[j])) {
		return true
	}
```

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,124 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/apache/arrow/go/v14/arrow/array"
+	"github.com/apache/arrow/go/v14/arrow/compute"
+	"github.com/apache/arrow/go/v14/arrow/memory"
+	"golang.org/x/exp/constraints"
```
gernest, line 13:
```suggestion
```

gernest: No need to pull up this constraint. It contains `~string` while we only use `string`.

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,124 @@
...
+		compute.NewDatum(r),
+		compute.NewDatum(indices),
+	)
+	if err != nil {
+		return nil, err
+	}
+	return res.(*compute.RecordDatum).Value, nil
+}
+
+type orderedArray[T constraints.Ordered] interface {
```
gernest, line 67:
```suggestion
type orderedArray[T int64 | float64 | string] interface {
```

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,124 @@
...
+	Value(int) T
+	IsNull(int) bool
+	Len() int
+}
+
+type orderedSorter[T constraints.Ordered] struct {
```
gernest, line 73:
```suggestion
type orderedSorter[T int64 | float64 | string] struct {
```

gernest (changes requested):
Looks good, I left a few suggestions. 

Thinking about it, do we really need to handle `*array.Binary`? we are not consistent with how we handle `string` schema column. I see a  lot of variations ,some cases `*array.String` some cases `*array.Binary`.

I was wondering maybe we can consolidate this, have `*array.String` for `string` column and when we add `binary` columns then they can be `*array.Binary`.

This is enough foundation. I will do folloup patches to cover

- Multi column sort
- Support direction (ascending and descending)
- NullFirst (for ascending and descending)
- Pooling/ Reusing the sorting objects ( indices and the indices array builder should be reused)

pqarrow/arrowutils/sort.go
metalmatze: It seemed a bit bloated. This is a lot nicer. Thanks!

metalmatze:
You're correct with the `*array.Binary` column. It's not covered in the unit tests (yet). 
This was added for an internal experiment. I'm fine removing it from the sort code in this PR for now.

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if r.NumRows() == 0 {
+		return indicesBuilder.NewInt64Array(), nil
+	}
+	if r.NumRows() == 1 {
+		indicesBuilder.Append(0)
+		return indicesBuilder.NewInt64Array(), nil
+	}
+
+	indices := make([]int64, r.NumRows())
```
asubiotto, line 29: I think we have one too many copies than necessary here: one alloc here, then a copy of all indices in `AppendValues` below. I would suggest instead to do:
`indicesBuilder.Reserve(r.NumRows())`
To keep a single allocation then `indicesBuilder.AppendValue` one by one.

pqarrow/arrowutils/sort.go
```diff
@@ -0,0 +1,90 @@
...
+
+	indicesBuilder.AppendValues(indices, nil)
+	return indicesBuilder.NewInt64Array(), nil
+}
+
+// ReorderRecord reorders the given record's rows by the given indices.
+// This is a wrapper around compute.Take which handles the type castings.
+func ReorderRecord(r arrow.Record, indices arrow.Array) (arrow.Record, error) {
+	res, err := compute.Take(
+		context.Background(),
```
asubiotto, line 52: nit: pass in the context

asubiotto (approved):
LGTM, we can follow up with the remaining work

gernest (approved):

metalmatze:
Thanks for all the valuable comments and feedback!

PR #629, opened 2023-12-14
Write directly to arrow.Record in logictest

### LINKED ISSUE #548: Logictest: Build arrow records directly
Logictests build arrow records by building a parquet record and then using the converter to convert it into an arrow record. Since we no longer support inserting by Parquet records, we should just build the arrow record directly instead.

gernest:
Closes #548 

Waiting for #628  We need the ability to sort the `arrow.Record` before insertion into the table.

### DIFF
--- a/logictest/logic_test.go
+++ b/logictest/logic_test.go
@@ -140,12 +140,3 @@ func TestLogic(t *testing.T) {
 		}
 	})
 }
-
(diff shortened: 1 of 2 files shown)


gernest:
Closing this for now. Will focus on smaller patches.

PR #631, opened 2023-12-14
[WIP] Add implementation of  TimescaleDB time_bucket

gernest:
Inspired by this blogpost from timescaledb  [Aggregate time-series data with time_bucket](https://docs.timescale.com/use-timescale/latest/time-buckets/use-time-buckets/#aggregate-time-series-data-with-time_bucket) .

It is similar but simplified due to the nature of our api. I had a need to be able to compute aggregates for `daily`, `weekly`, `monthly` and `yearly` reports for my now defunct web analytics project.

Some underlying code is direct port from `c` implementation on `timescaledb` see function `timeBucket` function.  There is a long way to go untill this is ready for review.

### DIFF
--- a/query/logicalplan/expr.go
+++ b/query/logicalplan/expr.go
@@ -663,6 +663,91 @@ func (d *DurationExpr) Value() time.Duration {
 	return d.duration
 }
 
+type Period interface {
+	internal()
+}
+
+type Day int
+
+var _ Period = (*Day)(nil)
+
+func (Day) internal() {}
+
+type Week int
+
+var _ Period = (*Week)(nil)
+
+func (Week) internal() {}
+
+type Month int
+
+var _ Period = (*Month)(nil)
+
+func (Month) internal() {}
+
+type Year int
+
+var _ Period = (*Year)(nil)
+
+func (Year) internal() {}
+
+type TimeBucketExpr struct {
+	period Period
+}
+
+func TimeBucket(period Period) *TimeBucketExpr {
+	return &TimeBucketExpr{period: period}
+}
+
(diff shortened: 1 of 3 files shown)


thorfour:
Neat!

gernest:
Closing this for now. Will focus on smaller patches ( perf stuff are more interesting)

metalmatze:
It would still be very cool to get to this (even in smaller chunks as you wrote). 

Parca currently does this in the application layer: https://github.com/parca-dev/parca/blob/718d30a137dbd5bef8db3f58c798c7fa6e4e4113/pkg/parcacol/querier.go#L628-L641

gernest:
@metalmatze I see there is a TODO on the link you provided. 

Implementing only that use case is possible. I mean only accepting a step which is  `time.Duration` ,  and time representation `ms/s/ns` .

It wont cover `time_bucket` but it can be something close to what `Duration` expression does.  `time_bucket` is way nuanced than that, it takes into account for days of the week/month etc .

Happy to resurrect this given enough interest. I am just worried that I will invest a lot of time on things that `frostdb` team might deem not important, or give maintainers headache reviewing large diffs of irrelevant stuff.

I pretty much just take small tasks that can be reviewed while having a cup of coffee.

Addressing this will also eat a chunk of your time as well, we will probably need to discuss tradeoffs and see what works well with `frostdb` and `parca`.

metalmatze:
I see.
This looked like the same problem to me or at least similar enough to reuse it. On second thought it seems to be ever so slightly different as you pointed out. 
Let's keep both problems separate then. 

Nonetheless, the feature in this PR seems interesting!

PR #632, opened 2023-12-15
Define dynamic column as map[string]T

gernest:
No special tag is needed any field of type `map[string]T` where `T` is
`int64|bool|float64|string` is a dynamic column.

Example

```go
type Sample struct {
	Labels      map[string]string `frostdb:"labels,rle_dict,asc(1),null_first"`
}
```


thorfour (approved):
Great thanks!

PR #633, opened 2023-12-15
pqarrow/arrowutils: Add optional limit to MergeRecords

metalmatze:
This is helpful for merging multiple sorted records but only keeping the first e.g. 1000 rows.

### DIFF
--- a/pqarrow/arrowutils/merge.go
+++ b/pqarrow/arrowutils/merge.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"container/heap"
 	"fmt"
+	"math"
 
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/apache/arrow/go/v14/arrow/array"
@@ -14,11 +15,15 @@ import (
 
 // MergeRecords merges the given records. The records must all have the same
 // schema. orderByCols is a slice of indexes into the columns that the records
-// and resulting records are ordered by. Note that the given records should
-// already be ordered by the given columns.
+// and resulting records are ordered by. While ordering the limit is checked before appending more rows.
+// If limit is less than 1, no limit is applied.
+// Note that the given records should already be ordered by the given columns.
 // WARNING: Only ascending ordering is currently supported.
 func MergeRecords(
-	mem memory.Allocator, records []arrow.Record, orderByCols []int,
+	mem memory.Allocator,
+	records []arrow.Record,
+	orderByCols []int,
+	limit uint64,
 ) (arrow.Record, error) {
 	h := cursorHeap{
 		cursors:     make([]cursor, len(records)),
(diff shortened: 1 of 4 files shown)


pqarrow/arrowutils/merge.go
```diff
@@ -14,11 +15,15 @@ import (
...
+// and resulting records are ordered by. While ordering the limit is checked before appending more rows.
+// If limit is less than 1, no limit is applied.
+// Note that the given records should already be ordered by the given columns.
 // WARNING: Only ascending ordering is currently supported.
 func MergeRecords(
-	mem memory.Allocator, records []arrow.Record, orderByCols []int,
+	mem memory.Allocator,
+	records []arrow.Record,
+	orderByCols []int,
+	limit int64,
```
gernest, line 26: > If limit is less than 1, no limit is applied.

Can we just say `0` . I checked  `MergeRecords` call site the only cases with no limit uses `-1` . In this case `0` pretty much means no limit, we can leverage the type system and use `uint64`

```suggestion
	limit uint64,
```

metalmatze: Sounds like a plan :+1:

metalmatze:
Finally resolved the small mistake and this should be ready for another review with `uint64` now.

gernest (changes requested):
Looks good, I left a suggestion.

So, I have a question about use case for `limit`. The way I see it is we limit how many records gets merged.

What happens when the limit is exceeded? There is no way for the caller to get feedback on which records were not merged.

Is this intended behaviour? That caller

metalmatze:
Thinking about this some more, and I'm undecided what the correct API is. We have two options, I suppose. 
1. We keep the returned types like they are (`arrow.Record, error`) and create a specific error that can be treated as a warning using `errors.Is`
2. We add another type to the return that either `arrow.Record, string, error` or even `arrow.Record, error, error` where the second returned type would always be a warning if not `nil`. For what it's worth, this is what Prometheus does in some APIs. 

Which one do you prefer?

gernest:
@metalmatze  I'm not sure about both options, to me it is matter of practicality. Just informing something was left out is not actionable. I need to know which was left out so I can chart other course of action ( retry / drop  ...etc ).

The passed input records may not be processed in order they come in, so we can rule out returning number of records processed.

The only viable option is at the end of processing, collect all records still in the cursor and return them back to the caller together with the merged record ( yes! It is ugly!). `( merged arrow.Record, remaining []arrow.Record, err error)`

So,

- If the the goal is to just pick first `n` and ignore everything else. I think the current state of changes is fine ( We can just add more info on the comment regarding this behaviour)
- If the the goal is to merge first `n` records, we need to provide a way for the caller to identify which records were never merged. 

Either way, this is just my thought on this, it shouldn't block the PR from being merged,

metalmatze:
Talking to the others, we only have use cases where we might want to limit to the first `n` rows and drop the rest. 
We're open to reconsidering if we ever have the use case of still wanting to know the dropped records.

PR #634, opened 2023-12-15
query: Add Contains operator

metalmatze:
Adding a `|=` operator to FrostDB.
It's basically `LIKE '%$foo%'` in SQL. It'll return true whenever the string contains the substring.

### DIFF
--- a/logictest/testdata/exec/filter/filter
+++ b/logictest/testdata/exec/filter/filter
@@ -129,3 +129,14 @@ select labels, timestamp, value where doesntexist >= 4
 exec
 select labels, timestamp, value where doesntexist <= 4
 ----
+
+exec
+select stacktrace, value where stacktrace LIKE 'ack'
+----
+stack1  1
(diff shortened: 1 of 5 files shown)


thorfour:
Do we need this if we could do this with regexp matching?

metalmatze:
Interesting idea to use the regexp. 
We still need this as it uses substring matching rather than full regexp which I assume to be _a lot_ cheaper. We could benchmark this.

thorfour:
> Interesting idea to use the regexp. We still need this as it uses substring matching rather than full regexp which I assume to be _a lot_ cheaper. We could benchmark this.

Instead of regexp matching instead have the regexp operator downgrade itself into a `Contains` matcher if it detects it's regexp is a substring matcher.

metalmatze:
I'll start fresh from the regexp angle.

brancz:
What Thor said would be a great logical optimizer.

PR #635, opened 2023-12-16
Upgrade  arrowutils.SortRecord

gernest:
Followup to #628

This commit enhances `arrowutils.SortRecord` to support

- Multi column sort
- Column direction (ascending, descending)
- NullFirst (for both ascending and descending)

Also optimizes `arrowutils.ReorderRecord` to support dictionary columns.


pqarrow/arrowutils/sort_test.go
```diff
@@ -11,73 +11,356 @@ import (
 )
 
 func TestSortRecord(t *testing.T) {
```
asubiotto, line 13: I think it would also be great if we could add some benchmarks for these.

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,303 @@ package arrowutils
...
+// the column index.
+//
+// Comparison is made sequentially by each column. When rows are equal in the
+// first column we compare the rows om the second column and so on and so forth
+// until rows that are not equal are found.
+func SortRecord(
+	mem memory.Allocator,
+	r arrow.Record,
+	columns []int,
+	direction []int,
```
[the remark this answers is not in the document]
metalmatze: Passing these as structs would be nice!
I was already playing with the idea during the previous PR but it seemed overkill for what it was.

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,303 @@ package arrowutils
...
+// column.
+// By default nullFirst is false for all sorting columns
+//
+// Both direction and nullsFirst are optional, nil is allowed as value. For non
+// nil values they must be the same length as columns where each index maps to
+// the column index.
+//
+// Comparison is made sequentially by each column. When rows are equal in the
+// first column we compare the rows om the second column and so on and so forth
+// until rows that are not equal are found.
```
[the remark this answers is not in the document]
metalmatze: Some explanation would be good, yep! 
To be totally honest, to this day I don't get why index columns in arrow aren't unsigned by default too.

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,303 @@ package arrowutils
...
+	for i := 0; i < int(r.NumCols()); i++ {
+		_, hasDictionary = r.Column(i).(*array.Dictionary)
+		if hasDictionary {
+			break
+		}
+	}
+	if !hasDictionary {
+		res, err := compute.Take(
+			ctx,
+			compute.TakeOptions{BoundsCheck: true},
```
metalmatze, line 131: Isn't that the default? 
https://github.com/apache/arrow/blob/go/v14.0.0/go/arrow/compute/expression.go#L448

metalmatze (commented):
Amazing work! 💯 
Really liking the overall direction you're taking this in!

pqarrow/arrowutils/sort.go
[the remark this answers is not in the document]
gernest: done ff6d120

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,303 @@ package arrowutils
...
+		nullsFirst = make([]bool, len(columns))
+	}
+	if len(direction) != len(columns) {
+		return nil, fmt.Errorf("pqarrow/arrowutils: expected %d directions got %d", len(columns), len(direction))
+	}
+	if len(nullsFirst) != len(columns) {
+		return nil, fmt.Errorf("pqarrow/arrowutils: expected %d nullsFirst got %d", len(columns), len(nullsFirst))
+	}
+
+	build := array.NewInt32Builder(mem)
```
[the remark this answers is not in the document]
gernest: done 2dcf9be

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,303 @@ package arrowutils
...
-		return indicesBuilder.NewInt64Array(), nil
+		build.Append(0)
+		return build.NewInt32Array(), nil
 	}
 
-	indices := make([]int64, r.NumRows())
-	// populate indices
-	for i := range indices {
-		indices[i] = int64(i)
+	ms := &multiColSorter{
```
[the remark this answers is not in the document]
gernest: done ca45c04

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,303 @@ package arrowutils
...
-	)
+// ReorderRecord uses indices to create a new record that is sorted according to
+// the indices array.
+func ReorderRecord(ctx context.Context, mem memory.Allocator, r arrow.Record, indices *array.Int32) (arrow.Record, error) {
+	ctx = compute.WithAllocator(ctx, mem)
+	// compute.Take doesn't support dictionaries. Use take on r when r does not have
+	// dictionary column.
+	var hasDictionary bool
+	for i := 0; i < int(r.NumCols()); i++ {
+		_, hasDictionary = r.Column(i).(*array.Dictionary)
```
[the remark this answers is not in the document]
gernest: done 87436e1

pqarrow/arrowutils/sort_test.go
gernest: I'm pretty new to `frostdb`, this probably my second or third week on the code base. While I'm comfortatable writing tests, I think I'm not qualified to write benchmarks yet.

I hope someone more knowledgable and better qualified should help with benchmarks. What I'm saying benchmarking is beyond the scope of this PR.

pqarrow/arrowutils/sort.go
gernest: Yes, the arrow api is pretty much broken on this regard `compute.DefaultTakeOptions` is a pointer while `compute.Take` accepts a non pointer. Doing pointer derefence for arguments in this case is just not okay.

gernest:
@metalmatze @asubiotto thanks for the feedback, PTAL. I addressed all the comments.

Ignore `fix lint` commits, the linter action was generating partial recommendations, I had to fight for my life to make it happy.

pqarrow/arrowutils/sort_test.go
asubiotto: How about now? 😂 I feel you are more than qualified given your recent contributions, but up to you.

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,306 @@ package arrowutils
...
-	if len(cols) > 1 {
-		return nil, fmt.Errorf("sorting by multiple columns isn't implemented yet")
+func (d Direction) comparison() int {
+	switch d {
+	case Ascending:
+		return -1
+	case Descending:
+		return 1
+	default:
+		return 0
```
asubiotto, line 32: I would panic in this default case since it doesn't make sense

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,306 @@ package arrowutils
...
+}
+
+// SortRecord sorts given arrow.Record by columns.
+// Returns  *array.Int32 of  indices to sorted rows or record r.
+//
+// Comparison is made sequentially by each column. When rows are equal in the
+// first column we compare the rows om the second column and so on and so forth
+// until rows that are not equal are found.
+func SortRecord(
+	indicesBuilder *builder.OptInt32Builder,
```
asubiotto, line 50: I don't like the API of having to instantiate a builder and then passing it in to `SortRecord`. Any reason you did it this way? Can we remove and instantiate within the `SortRecord` function?

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,306 @@ package arrowutils
...
 
+// SortingColumn describes a sorting column on a arrow.Record.
+type SortingColumn struct {
+	Index      int
+	Direction  Direction
+	NullsFirst bool
+}
+
+// SortRecord sorts given arrow.Record by columns.
+// Returns  *array.Int32 of  indices to sorted rows or record r.
```
asubiotto, line 44: nit while you're here: extra spaces in comment

pqarrow/builder/optbuilders.go
```diff
@@ -637,6 +638,22 @@ func (b *OptInt32Builder) Set(i int, v int32) {
...
+
+// UnsafeSet sets v at index i without bounds check.
+func (b *OptInt32Builder) UnsafeSet(i int, v int32) {
+	b.data[i] = v
+	bitutil.SetBit(b.validityBitmap, i)
+}
+
+// UnsafeValue returns value at index i without bounds check.
+func (b *OptInt32Builder) UnsafeValue(i int) int32 {
+	return b.data[i]
```
asubiotto, line 654: I don't think we need a bounds check in `Value` since this is performed implicitly by the runtime (just that the panic would be different. We can probably remove this `UnsafeValue` and remove the `if` from `Value`. What do you think? cc @metalmatze since I think you introduced this code.

pqarrow/builder/optbuilders.go
```diff
@@ -637,6 +638,22 @@ func (b *OptInt32Builder) Set(i int, v int32) {
...
 
+// UnsafeSwap swaps without bounds check.
+func (b *OptInt32Builder) UnsafeSwap(i, j int) {
+	b.data[i], b.data[j] = b.data[j], b.data[i]
+}
+
+// UnsafeSet sets v at index i without bounds check.
+func (b *OptInt32Builder) UnsafeSet(i int, v int32) {
+	b.data[i] = v
+	bitutil.SetBit(b.validityBitmap, i)
```
asubiotto, line 649: Interesting, is the fact that we don't set the null bitmap in `Set` a bug? Same idea here as below in `UnsafeValue`: I don't think the application-level bounds check is necessary so we could probably merge

pqarrow/arrowutils/sort_test.go
gernest: haha! lol! I think I can do it, will make sure we have a benchmark in place.

pqarrow/arrowutils/sort.go
gernest: done c087044

pqarrow/arrowutils/sort.go
gernest: done 18f38f1

pqarrow/arrowutils/sort.go
gernest: done c00f9d4

gernest:

**NOTE** . I have left `*OptInt32Builder` api comments because changing api or api expectation is sensitive and they will need more context and input from @metalmatze . I can handle whatever feedback in followup PR.  I have also decided to leave benchmark out.

pqarrow/arrowutils/sort.go
```diff
@@ -2,93 +2,306 @@ package arrowutils
...
+		if err != nil {
+			return err
+		}
+		arr[idx] = r
+		return nil
 	}
+}
 
-	return indicesBuilder.NewInt64Array(), nil
+func takeDictColumn(ctx context.Context, a *array.Dictionary, idx int, arr []arrow.Array, indices *array.Int32) func() error {
```
asubiotto, line 138: nit: how about simply returning an error instead of a callback in all these `take` functions?? You'd have to call this within a `func() error { }` block in the `Go` above, but I think it's nicer given that it gives the option to directly execute this function

pqarrow/builder/optbuilders.go
```diff
@@ -688,3 +705,9 @@ func (b *OptInt32Builder) ResetToLength(n int) {
 	b.data = b.data[:n]
 	b.validityBitmap = resizeBitmap(b.validityBitmap, n)
 }
+
+func (b *OptInt32Builder) ReserveToLength(n int) {
```
asubiotto, line 709: can we just call this `Reserve` to be more in line with the arrow API?

asubiotto (commented):
Just a couple of nits, but this is looking good!

pqarrow/arrowutils/sort.go
gernest: done 7be2e2b

pqarrow/builder/optbuilders.go
gernest: done c236674

gernest:
@asubiotto  thanks for the review. I addressed all comments PTAL.

Hopeful we put this into a practical use. Areas I think it will have impact

- logictest : we can now build records directly and skip parquet rows construction

- in parca

https://github.com/parca-dev/parca/blob/4f536249850d27a79b365d3d796ca8ff692896d5/pkg/parcacol/ingest.go#L63-L107
Now we have a generic builder that constructs arrow records directly, sorting was the only missing piece.

asubiotto (approved):
Absolutely, this is hugely impactful work, thanks for your contribution!

pqarrow/builder/optbuilders.go
metalmatze: Not exactly sure anymore. Let's see how it goes now that it's merged. Happy to revise in the future.

metalmatze:
Fantastic work! :100:

pqarrow/builder/optbuilders.go
asubiotto: The problem is that we now have duplicate methods that do the same thing (and one version seems to have a bug with setting nulls IIRC from another comment on this PR). We should probably fix this sooner rather than later.

gernest: @metalmatze @asubiotto I created a PR for this #680 we can move this discussion there.

PR #638, opened 2023-12-17
Support nullable base types for (*Build[T])

gernest:
Before, we had `null` tag that set schema column to nullable. However the base types allowed were `int64|float64|bool|string`.

This begged the question, what is `null`? for a `nullable` field of type `int64` do we interpret `0` as `null`?, probably not. We were offering option for something that was impossible to represent.

Fortunately, we can safely represent null values with out type system. For
base types   `*int64|*float64|*bool|*string` defines nullable base types.

This commit add support for both `int64|float64|bool|string` and `*int64|*float64|*bool|*string` for base types . This works with dynamic columns too.

With this change `null` tag is redundant so It was removed. Fields of type `*int64|*float64|*bool|*string` will automatically generate nullable schema
 column.

NOTE: I also discovered a bug where when `T` has dynamic column and you append multiple `T` without any dynamic column followed by T with dynamic column caused a panic. The issue was how we adjusted dynamic columns to match current `.Append` rows state. The fix is included in this commit because I needed this behavior in tests.

### DIFF
--- a/dynparquet/record_builder.go
+++ b/dynparquet/record_builder.go
@@ -50,7 +50,6 @@ const (
 //	                   desc | Sorts in descending order.Use desc(n) where n is an integer for sorting order
 //	                lz4_raw | LZ4_RAW compression.
 //	               pre_hash | Prehash the column before storing it.
-//	                   null | Nullable column.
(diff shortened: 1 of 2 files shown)


thorfour (approved):
Awesome work thanks!

PR #639, opened 2023-12-17
Delete query/physicalplan/physicalplan.test

meson10:
Looks like an accidentally committed file.


gernest (approved):
LGTM . Confirmed there is no reference to this file.

thorfour (approved):

brancz:
We need the SLA to be signed. Otherwise lgtm, thank you!

PR #642, opened 2023-12-20
Add Implementation of  L1 arrow compaction

### LINKED ISSUE #433: L1 arrow compaction
It may be useful to have the option to compact L0 arrow records into L1 arrow records instead of Parquet.

gernest:
Closes #433

### DIFF
--- a/parts/arrow.go
+++ b/parts/arrow.go
@@ -2,8 +2,12 @@ package parts
 
 import (
 	"bytes"
+	"sort"
 
 	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/apache/arrow/go/v14/arrow/array"
+	"github.com/apache/arrow/go/v14/arrow/memory"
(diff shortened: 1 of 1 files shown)


gernest:
Parts are not guaranteed to have the same number of columns( when dyncamic columns are involved). For each compaction call we collect all columns observed across all parts.  For dynamic columns, to ensure all columns have the same number of rows, we fill them with nulls when they are not found in a part.

Initially I had thought since we expect parts to be sorted , then  merging them should be enough. However, I have doubts about this.

- We support multiple columns sort (including dynamic columns), the new part will contain nulls in  no guaranteed order which messes up sorting expectation.
- We operate under assumptions that main sorting column behaves like `timestamp` which is always increasing. When using different main sorting column. It doesn't matter much if the part was sorted before. New record must be sorted for sorting to make sense.

gernest:
Seems to be bigger than I anticipated. Closing this for now, will focus on smaller patches.

PR #644, opened 2023-12-20
Performance improvement on pqarrow.RecordDynamicCols

gernest:
This function is called on a hot path. I couldn't resist the temptation to squeeze some performance gains from it.

Benchmark results:

```
goos: darwin
goarch: amd64
pkg: github.com/polarsignals/frostdb/pqarrow
                    │    old.txt    │             new.txt             │
                    │    sec/op     │    sec/op     vs base           │
RecordDynamicCols-8   1470.0n ± ∞ ¹   792.6n ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                    │   old.txt   │            new.txt             │
                    │    B/op     │    B/op      vs base           │
RecordDynamicCols-8   936.0 ± ∞ ¹   536.0 ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                    │   old.txt    │            new.txt             │
                    │  allocs/op   │  allocs/op   vs base           │
RecordDynamicCols-8   14.000 ± ∞ ¹   6.000 ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

### DIFF
--- a/pqarrow/parquet.go
+++ b/pqarrow/parquet.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"sort"
 	"strings"
 
 	"github.com/apache/arrow/go/v14/arrow"
@@ -11,7 +12,6 @@ import (
 	"github.com/apache/arrow/go/v14/arrow/scalar"
 	"github.com/parquet-go/parquet-go"
 
-	"github.com/polarsignals/frostdb/bufutils"
 	"github.com/polarsignals/frostdb/dynparquet"
 	"github.com/polarsignals/frostdb/pqarrow/arrowutils"
 )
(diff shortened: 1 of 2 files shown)


pqarrow/parquet.go
```diff
@@ -154,17 +154,23 @@ func recordToRows(w dynparquet.ParquetWriter, dcv dynamicColumnVerifier, record
...
 		af := record.Schema().Field(i)
-		parts := strings.SplitN(af.Name, ".", 2)
-		if len(parts) == 2 { // dynamic column
-			dyncols[parts[0]] = append(dyncols[parts[0]], parts[1])
+		if strings.Contains(af.Name, ".") {
+			dyncols[af.Name] = struct{}{}
 		}
 	}
-
-	return bufutils.Dedupe(dyncols)
```
thorfour, line 167: Is dedupe used anywhere now?

thorfour (approved):
Awesome!

pqarrow/parquet.go
gernest: @thorfour  yes only one place in `ValuesToBuffer`, but when we merge #645

Then I will submit a PR to remove it as well. Because there will  beno any reference to it in `frostdb`.

gernest: I just rebased to main (In case you are wondering about the forced push)

PR #646, opened 2023-12-20
parquet/converter: New Record call ResetFull on dictionary builders

thorfour:
This frees the dictionaries between record creations so that stale dictionary values aren't being held.

We were seeing huge storage spikes resulting from this dictionary bloat.

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -359,6 +359,7 @@ func (c *ParquetConverter) NumRows() int {
 
 func (c *ParquetConverter) NewRecord() arrow.Record {
 	if c.builder != nil {
+		defer c.builder.Reset()
 		return c.builder.NewRecord()
 	}
 
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -359,11 +359,16 @@ func (c *ParquetConverter) NumRows() int {
 
 func (c *ParquetConverter) NewRecord() arrow.Record {
 	if c.builder != nil {
-		defer c.builder.Reset()
 		return c.builder.NewRecord()
 	}
 
 	return nil
+}
+
+func (c *ParquetConverter) Reset() {
+	if c.builder != nil {
+		c.builder.Reset()
+	}
 }
 
 func (c *ParquetConverter) Close() {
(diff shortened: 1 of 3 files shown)


pqarrow/arrow.go
```diff
@@ -359,6 +359,7 @@ func (c *ParquetConverter) NumRows() int {
 
 func (c *ParquetConverter) NewRecord() arrow.Record {
 	if c.builder != nil {
+		defer c.builder.Reset()
```
asubiotto, line 362: Do we want to make this part of NewRecord?

pqarrow/builder/recordbuilder.go
```diff
@@ -117,3 +117,14 @@ func (b *RecordBuilder) ExpandSchema(schema *arrow.Schema) {
 
 	b.schema = schema
 }
+
+// Reset will call ResetFull on any dictionary builders to prevent memo tables from growing unbounded.
+func (b *RecordBuilder) Reset() {
+	for _, f := range b.fields {
+		if lb, ok := f.(*ListBuilder); ok {
```
asubiotto, line 124: Do we also want to do this on dictionary builders that are not wrapped by a list builder?

pqarrow/arrow.go
thorfour: Your call. I don't have a preference

pqarrow/builder/recordbuilder.go
thorfour: So, we only see this issue with the repeated dictionaries, whereas the other dictionaries probably benefit from not a full reset, but I don't know how to express that.

pqarrow/arrow.go
asubiotto: I think so. This could come back to bite us if we ever reuse the builder

asubiotto (approved):

pqarrow/arrow.go
thorfour: We already reuse the builder. This just causes it to rebuild the dictionaries

thorfour: Made reset an explicit call.

PR #647, opened 2023-12-21
Add support for int32 columns

nicolastakashi:
Adding support to int32 columns.


nicolastakashi:
cc @metalmatze

query/physicalplan/binaryscalarexpr.go
```diff
@@ -148,6 +148,23 @@ func BinaryScalarOperation(left arrow.Array, right scalar.Scalar, operator logic
 		default:
 			panic("something terrible has happened, this should have errored previously during validation")
 		}
+	case arrow.PrimitiveTypes.Int32:
+		switch operator {
+		case logicalplan.OpEq:
+			return Int32ArrayScalarEqual(left.(*array.Int32), right.(*scalar.Int32))
```
thorfour, line 154: Couldn't all these just be one funciton that uses an operator to change how it performs the comparison? I think that might be fewer lines of code

nicolastakashi: Just double checking!
Do you mean encapsulate the logic we have in different functions in only one and reduce the amount of functions?

thorfour: I mean that instead of the `Int32ArrayScalarEqual` 
we just have a `Int32ArrayScalarComparison(op, left,right` where it changes the comparison it performs based on the Op

brancz: I don't think so, the code as is is what we want. We don't want to select the comparison in the hot loop, which is much slower (it's why the rest of the code is written this way). More of a side thing, but this also prepares us well for vectorizing these functions.

thorfour: Of note, when we've put selection in the hot loop before we haven't been able to notice during benchmarks any performance different. 

But since it's semantic I'm fine with merging this as is.

thorfour (approved):
Thanks for the contribution!

PR #657, opened 2023-12-25
Query concurrency option

thorfour:
Previously the query engine concurrency was hardcoded to GOMAXPROCS(0). This adds a physical plan option to manually set the concurrency in the query engine.

This exposed a bug with the query engine where if the concurrency was set to 1, aggregations would not work. So this also updates the aggregation code support not having multiple stages with a syncronizer and instead just having a single final aggregation step.

### DIFF
--- a/query/physicalplan/aggregate.go
+++ b/query/physicalplan/aggregate.go
@@ -343,31 +343,21 @@ func (a *HashAggregate) Callback(_ context.Context, r arrow.Record) error {
 
 		if _, ok := aggregate.dynamicAggregationsConverted[field.Name]; !ok {
 			for _, col := range aggregate.dynamicAggregations {
-				if a.finalStage {
-					if col.expr.MatchColumn(field.Name) {
-						// expand the aggregate.aggregations with a final concrete column aggregation.
-						columnToAggregate = append(columnToAggregate, nil)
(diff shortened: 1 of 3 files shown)


query/physicalplan/physicalplan.go
```diff
@@ -277,6 +277,12 @@ func WithOrderedAggregations() Option {
 	}
 }
 
+func WithConcurrency(concurrency int) Option {
```
gernest, line 280: Can you add a comment here? We also need to mention default value.

thorfour: Good call!

gernest (approved):
I can only do eyeball test on this part of the codebase. So far things look good to me.

query/physicalplan/aggregate.go
```diff
@@ -25,6 +25,7 @@ func Aggregate(
 	tracer trace.Tracer,
 	agg *logicalplan.Aggregation,
 	final bool,
+	only bool,
```
asubiotto, line 28: I'm not 100% certain what this parameter does or why is it necessary. It might be cleaner to define an enum and merge final and only if necessary (since `only -> final`).

thorfour: I ran into some weird behavior when trying this out in prod anyways. Gonna backburner this PR for now

PR #658, opened 2023-12-26
Support repeated fields of all base types on recordToRows

gernest:
Previously we were operating under assumptions that repeated fields were dictionary of `*array.Binary`.

This commit implements conversion on all arrow arrays of repeated base types `int64|float64|bool|string`

I have also added `*array.Binary` in the mix although it is not categorized as base type (Based on the use, It will make sense to introduce binary type in our schema proto)


thorfour (approved):
Awesome! Thanks

PR #659, opened 2023-12-26
Move (*Build[T]) to internal package

gernest:
Reasons

- It is useful in tests. Tying it to dynparquet introduces cyclic dependency
- It is not directly related to dynparquet
- Discourage direct use and promote `GenericTable`

I got tired constructing arrow records in tests, its messy and error prone. This way we can dog food the implementation fostering evolution and more use case coverage.

I had to move `dynparquet.Sample` to `samples` package in order to move the tests.


thorfour (approved):
Lgtm

PR #662, opened 2023-12-27
fix handling of *array.String in ParquetConverter

### LINKED ISSUE #661: pqarrow/arrow.go  assert *array.Dictionary only support *array.Binary type
I use record_builder to create arrow.Record, than insert into frostdb.  When i run aggregation query，returns an error: unsupported dictionary type: *array.String.  record builder support rle_dict to use “arrow.BinaryTypes.String”,"arrow.PrimitiveTypes.Int64","arrow.PrimitiveTypes.Float64","arrow.FixedWidthTypes.Boolean",

but pqarrow/arrow.go only support *array.Binary
case *array.Dictionary:
		b := builder.(*array.BinaryDictionaryBuilder)
		switch dict := arr.Dictionary().(type) {
		case *array.Binary:
			for i := 0; i < toCopy; i++ {
				if arr.IsNull(i) {
					b.AppendNull()
				} else {
					if err := b.Append(dict.Value(arr.GetValueIndex(i))); err != nil {
						panic("failed to append to dictionary")
					}
				}
			}
		default:
			panic(fmt.Sprintf("unsupported dictionary type: %T", dict))

gernest:
This commit adds `*array.String` handling inside `copyArrToBuilder` cases for both plain `*array.String` and dictionary of `*array.String` are handled.

Fixes #661

### DIFF
--- a/pqarrow/arrow.go
+++ b/pqarrow/arrow.go
@@ -993,6 +993,17 @@ func copyArrToBuilder(builder builder.ColumnBuilder, arr arrow.Array, toCopy int
 				b.Append(arr.Value(i))
 			}
 		}
+	case *array.String:
+		b := builder.(*array.BinaryBuilder)
+		for i := 0; i < toCopy; i++ {
+			if arr.IsNull(i) {
+				// We cannot use unsafe appends with the binary builder
+				// because offsets won't be appended.
+				b.AppendNull()
+			} else {
+				b.AppendString(arr.Value(i))
+			}
+		}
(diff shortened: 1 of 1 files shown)


thorfour (approved):
Sweet!

jicanghaixb:
Not completely fixed, arrowutils/groupranges.go,merge.go and builder/utils.go has the same issue.

gernest:
@jicanghaixb see my reply for you on the closed issue.

PR #664, opened 2023-12-27
Support all base types for hashListArray

gernest:
This is continuation to our efforts to Support all base types for repeated fields.

- For `int64` we use big endian binary encoding to pass to xxhash digest
- For `float64` we use `math.Float64bits` to obtain `uint64` representation that we get big endian binary encoding to pass to the xxhash digest
- For `bool` we use `byte(2)` for true and `byte(1)` for false(This was taken from existing `hashBooleanArray` function) to pass to xxhash digest


dynparquet/hashed.go
```diff
@@ -105,19 +105,59 @@ func HashArray(arr arrow.Array) []uint64 {
 func hashListArray(arr *array.List) []uint64 {
 	res := make([]uint64, arr.Len())
 	digest := xxhash.New()
-
-	dictionaryList, binaryDictionaryList, err := arrowutils.ToConcreteList(arr)
```
thorfour, line 109: Wondering if the removal of this has performance implications? I recall it being put in to reduce the number of allocations that were performed

gernest: > Wondering if the removal of this has performance implications?

I don't think so, we really didn't remove it. We just extended it to cover other cases. We still support dictionaries of `*array.Binary`, we just don't mendate it.

- For list of binary dictionaries nothing has changed (before this commit and now)
- For allocations , in  `int64` ,`float64` and `bool` we reuses buffers on the stack so allocation wise they are okay.

dynparquet/hashed.go
```diff
@@ -105,19 +105,59 @@ func HashArray(arr arrow.Array) []uint64 {
...
+	case *array.String:
+		hv = func(i int, d *xxhash.Digest) {
+			_, _ = d.WriteString(e.Value(i))
+		}
+	case *array.Dictionary:
+		switch dict := e.Dictionary().(type) {
+		case *array.Binary:
+			hv = func(i int, d *xxhash.Digest) {
+				_, _ = d.Write(dict.Value(e.GetValueIndex(i)))
+			}
```
gernest, line 145: @thorfour here is where we are covering the case before this change( list of binary dictionary).

dynparquet/hashed.go
gernest: So I ran benchmarks. there is a regression. I am fixing it now.

gernest (commented):
There is performance regression in this change

```
goos: darwin
goarch: amd64
pkg: github.com/polarsignals/frostdb/dynparquet
                │   old.txt   │               new.txt               │
                │   sec/op    │   sec/op     vs base                │
HashListArray-8   286.6n ± 2%   365.1n ± 1%  +27.41% (p=0.000 n=10)

                │  old.txt   │               new.txt                │
                │    B/op    │    B/op      vs base                 │
HashListArray-8   48.00 ± 0%   128.00 ± 0%  +166.67% (p=0.000 n=10)

                │  old.txt   │               new.txt               │
                │ allocs/op  │ allocs/op   vs base                 │
HashListArray-8   1.000 ± 0%   2.000 ± 0%  +100.00% (p=0.000 n=10)
```

gernest:
I fixed regression (sort of)

```
goos: darwin
goarch: amd64
pkg: github.com/polarsignals/frostdb/dynparquet
                │   old.txt   │              new.txt               │
                │   sec/op    │   sec/op     vs base               │
HashListArray-8   286.6n ± 2%   293.0n ± 0%  +2.27% (p=0.014 n=10)

                │  old.txt   │            new.txt             │
                │    B/op    │    B/op     vs base            │
HashListArray-8   48.00 ± 0%   48.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                │  old.txt   │            new.txt             │
                │ allocs/op  │ allocs/op   vs base            │
HashListArray-8   1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

@thorfour any ideas why we see ` +2.27%` ?. Everything is pretty much identical thats why the other two stats have equal samples.

thorfour:
> I fixed regression (sort of)
> 
> ```
> goos: darwin
> goarch: amd64
> pkg: github.com/polarsignals/frostdb/dynparquet
> cpu: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
>                 │   old.txt   │              new.txt               │
>                 │   sec/op    │   sec/op     vs base               │
> HashListArray-8   286.6n ± 2%   293.0n ± 0%  +2.27% (p=0.014 n=10)
> 
>                 │  old.txt   │            new.txt             │
>                 │    B/op    │    B/op     vs base            │
> HashListArray-8   48.00 ± 0%   48.00 ± 0%  ~ (p=1.000 n=10) ¹
> ¹ all samples are equal
> 
>                 │  old.txt   │            new.txt             │
>                 │ allocs/op  │ allocs/op   vs base            │
> HashListArray-8   1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=10) ¹
> ¹ all samples are equal
> ```
> 
> @thorfour any ideas why we see ` +2.27%` ?. Everything is pretty much identical thats why the other two stats have equal samples.

I am unsure, but I can take a look this week.

thorfour:
Apologies I ended up getting sick this week and didn't get to it. I'll prioritize next week

gernest:
> Apologies I ended up getting sick this week and didn't get to it. I'll prioritize next week

Sorry to hear that, this is nothing urgent. Get well soon.

thorfour (approved):
Okay, took a look. My theory would be that it's because we've increased the number of comparisons being performed in the type switch statement. So previously we just assumed it was a dictionary type and moved on, but now we have to perform 6 type comparisons before we know that it's a dictionary, and I think that's the reason for the small discrepancy. 

I think it's a worthwhile tradeoff

PR #666, opened 2023-12-29
Optimize dynparquet.MergeDynamicColumnSets

gernest:
This commit

- Adds test (There were no test covering this function)
- Introduce a new implementation for merging column sets that can be reused.
- Use `*sync.Pool` to reuse the new dynamic column set merger

Benchmark result

```
goos: darwin
goarch: amd64
pkg: github.com/polarsignals/frostdb/dynparquet
                         │   old.txt   │               new.txt               │
                         │   sec/op    │   sec/op     vs base                │
MergeDynamicColumnSets-8   2.220µ ± 1%   1.330µ ± 1%  -40.09% (p=0.000 n=10)

                         │   old.txt   │              new.txt               │
                         │    B/op     │    B/op     vs base                │
MergeDynamicColumnSets-8   1104.0 ± 0%   544.0 ± 0%  -50.72% (p=0.000 n=10)

                         │   old.txt   │              new.txt               │
                         │  allocs/op  │ allocs/op   vs base                │
MergeDynamicColumnSets-8   16.000 ± 0%   6.000 ± 0%  -62.50% (p=0.000 n=10)
```

### DIFF
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -1351,11 +1351,10 @@ func mergeDynamicRowGroupDynamicColumns(rowGroups []DynamicRowGroup) map[string]
 	return MergeDynamicColumnSets(sets)
 }
 
-func MergeDynamicColumnSets(sets []map[string][]string) (o map[string][]string) {
+func MergeDynamicColumnSets(sets []map[string][]string) map[string][]string {
 	m := newDynamicColumnSetMerger()
-	o = m.Merge(sets)
-	m.Release() // avoid defer penalty by releasing manually
-	return
+	defer m.Release()
+	return m.Merge(sets)
 }
 
 type dynColSet struct {
--- a/dynparquet/schema.go
+++ b/dynparquet/schema.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


brancz:
Dynamic columns are sorted intentionally, so we can do a k-way merge, we just hadn't implemented doing that yet.

gernest:
@brancz 
> Dynamic columns are sorted intentionally, so we can do a k-way merge, we just hadn't implemented doing that yet.

Yes, I think this is described in commit message but didn't go into godoc for `MergeRecords `  #293

I know this is a statement, but I was wondering if you could provide some context relative to this PR.

**_note_** : in this PR behaviour is the same ,only implementation changed (dynamic columns are sorted as before)

dynparquet/schema.go
```diff
@@ -1350,47 +1351,67 @@ func mergeDynamicRowGroupDynamicColumns(rowGroups []DynamicRowGroup) map[string]
...
-			if !seen {
-				dynamicColumns[k] = [][]string{}
-			}
-			dynamicColumns[k] = append(dynamicColumns[k], v)
-		}
-	}
+func MergeDynamicColumnSets(sets []map[string][]string) (o map[string][]string) {
+	m := newDynamicColumnSetMerger()
+	o = m.Merge(sets)
+	m.Release() // avoid defer penalty by releasing manually
```
asubiotto, line 1357: defer is pretty cheap in these new versions of go, have you seen it be a bottleneck? I would just `defer` this to also execute on e.g. panic (if that is a possiblity)

asubiotto (commented):
I agree with Frederic, If we can implement a k-way merge that would be wonderful.

dynparquet/schema.go
gernest: > have you seen it be a bottleneck?

nope, the overhead is negligible.  There is no panic ,so we don't need to recover.

done f05d317

gernest:

dynparquet/schema.go
```diff
@@ -1351,46 +1352,65 @@ func mergeDynamicRowGroupDynamicColumns(rowGroups []DynamicRowGroup) map[string]
...
-func mergeStrings(str [][]string) []string {
-	result := []string{}
-	seen := map[string]struct{}{}
-	for _, s := range str {
-		for _, n := range s {
-			if _, ok := seen[n]; !ok {
-				result = append(result, n)
-				seen[n] = struct{}{}
+func (c *dynColSet) Merge(sets []map[string][]string) (o map[string][]string) {
+	o = make(map[string][]string)
```
asubiotto, line 1386: The nice thing about doing a k-way merge is that you can avoid allocations of "seen" data points. I would implement this using `container/heap` as a crutch. I don't think it's necessary to implement this in this PR, but could you add a TODO?

asubiotto (approved):

dynparquet/schema.go
gernest: done c3bfff1

I just realised I completely misunderstood @brancz comment.

PR #669, opened 2024-01-03
*: Add mathematical projections

brancz:
After this, we can remove the average aggregation and instead have a logical optimizer that replaces an average expression with `sum(col) / count(col) as avg(col)`.

This is also some groundwork, so we can perform `count * period` calculations in Parca, which is required so we can compare profiling data with different periods.

@thorfour @asubiotto

### DIFF
--- a/query/physicalplan/project.go
+++ b/query/physicalplan/project.go
@@ -14,7 +14,9 @@ import (
 )
 
 type columnProjection interface {
+	// Name returns the name of the column that this projection will return.
 	Name() string
+	// Projects one or more columns. Each element in the field list corresponds to an element in the list of arrays.
 	Project(mem memory.Allocator, ar arrow.Record) ([]arrow.Field, []arrow.Array, error)
 }
 
@@ -34,7 +36,7 @@ func (a aliasProjection) Project(mem memory.Allocator, ar arrow.Record) ([]arrow
 		if err != nil {
 			return nil, nil, err
 		}
-		fields, array, err := binaryExprProjection{boolExpr: boolExpr}.Project(mem, ar)
+		fields, array, err := boolExprProjection{boolExpr: boolExpr}.Project(mem, ar)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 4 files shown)


query/physicalplan/project.go
```diff
@@ -53,20 +55,187 @@ func (a aliasProjection) Project(mem memory.Allocator, ar arrow.Record) ([]arrow
...
+		case logicalplan.OpSub:
+			return []arrow.Field{leftFields[0]}, []arrow.Array{SubInt64s(leftArray, rightArray.(*array.Int64))}, nil
+		case logicalplan.OpMul:
+			return []arrow.Field{leftFields[0]}, []arrow.Array{MulInt64s(leftArray, rightArray.(*array.Int64))}, nil
+		case logicalplan.OpDiv:
+			return []arrow.Field{leftFields[0]}, []arrow.Array{DivInt64s(leftArray, rightArray.(*array.Int64))}, nil
+		default:
+			return nil, nil, fmt.Errorf("unsupported binary expression: %s", b.expr.String())
+		}
+	case *array.Float64:
```
thorfour, line 108: Since we support Int32's now should we also add support for that here?

thorfour (approved):
LGTM

jicanghaixb:
It would be even better if hybrid computing could be supported in the future.  example: a * b / c

brancz:
Can you elaborate what you mean by that and how that's different from this PR?

jicanghaixb:
now forstdb only support simple mathematical operations，not support complex calculations，taking aggregation as an example:
select sum(value) as value_sum, t count(value) as value_count , avg(value) as value_avg group by labels.label2
the above query is not allowed, because avg implicitly converts to sum and count functions, resulting in aggregator conflicts（column name）

Perhaps frost db needs to have its own Data Definition Language to analyze data in order to be applied to more data analysis fields, similar to the following SQL expression: 
SELECT 
    SUM(IF(status != 200, 1, 0)) AS errors, 
    SUM(IF(status = 200, 1, 0)) AS success, 
    errors / COUNT(server) AS error_rate, 
    success / COUNT(server) AS success_rate, 
    SUM(response_time) / COUNT(server) AS load_avg, 
    MIN(response_time), 
    MAX(response_time), 
    path, 
    server
GROUP BY 
    server, 
    path
HAVING errors > 0
ORDER BY 
    server ASC, 
    load_avg DESC

brancz:
We definitely want to do this!

PR #670, opened 2024-01-04
Remove manual record building for Samples

gernest:
This commit uses `internal/records.Build` to build `arrow.Record`  for `Samples`. The plan is
 to eventually remove all manual `arrow.Record` generation in tests.

Also, testing package was changed from `records` to `records_test` because we uses Samples to test the builder which causes cyclic dependency. Testing package change is basically a NOOP change(everything stays the same when testing just a different test binary name.)

### DIFF
--- a/internal/records/record_builder.go
+++ b/internal/records/record_builder.go
@@ -14,7 +14,6 @@ import (
 	"github.com/google/uuid"
 
 	schemapb "github.com/polarsignals/frostdb/gen/proto/go/frostdb/schema/v1alpha1"
-	"github.com/polarsignals/frostdb/samples"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/records/record_builder.go
```diff
@@ -739,11 +738,21 @@ func newUUIDSliceField(mem memory.Allocator, name string) (f *fieldBuilderFunc)
 	}
 	bd := b.(*array.BinaryDictionaryBuilder)
 	f.buildFunc = func(v reflect.Value) error {
-		return bd.Append(samples.ExtractLocationIDs(v.Interface().([]uuid.UUID)))
+		return bd.Append(ExtractLocationIDs(v.Interface().([]uuid.UUID)))
 	}
 	return
 }
 
+func ExtractLocationIDs(locs []uuid.UUID) []byte {
```
gernest, line 746: This was copied to avoid importing `sample` package. Eventually `[]uuid.UUID` handling will go away together with this. Let's leave it here for now.

thorfour (approved):
Love it!

PR #673, opened 2024-01-05
buf format -w

gernest:
I was working on protobuf files and noticed my vscode was formatting the files differently.

We already use buf for generating, it is reasonable to use buf for formatting as well. Strictly for consistency.

### DIFF
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -1,0 +1,19 @@
+name: Buf
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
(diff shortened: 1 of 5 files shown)


thorfour (commented):
I think I'm fine with this, but I think we need to add a buf fmt check to our ci to ensure this is maintained

gernest:
@thorfour thanks for the review. I added a workflow that will fail if a PR touches `proto/**/*.proto` and didn't apply `buf format -w`. PTAL

thorfour (approved):
Great thanks!

PR #675, opened 2024-01-08
Return transaction on (*GenericTable[T]).Write

gernest:
We are wrapping `(*Table).InsertRecord` call but were ignoring the transaction . This commit ensures we match `(*Table).InsertRecord`  return signature.

I also added godoc on the method

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -293,7 +293,7 @@ func TestDurationAggregation(t *testing.T) {
 			Value:      3,
 		},
 	}
-	err = table.Write(context.Background(), records...)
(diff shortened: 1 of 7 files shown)


asubiotto:
I think this is a breaking API change. I would backburner this unless we find a clear need for it cc @thorfour

gernest:
> I think this is a breaking API change. I would backburner this unless we find a clear need for it cc @thorfour

Agreed..

While we are on this subject, any reason we don't have tagged releases ?

asubiotto:
Nope

thorfour (approved):

PR #676, opened 2024-01-08
Return sorting metadata on (*records.Build[T]).NewRecord

gernest:
This commit wraps returned `arrow.Record` with sorting column information. We already have sorting implementation for `arrow.Record`.

The reason we need this is to eventually allow `GenericTable` to automatically build and sort `arrow.Record` before inserting them to the `lsm`, removing the requirement that values needs to be sorted before being written.

### DIFF
--- a/internal/records/record_builder.go
+++ b/internal/records/record_builder.go
@@ -20,6 +20,18 @@ const (
 	TagName = "frostdb"
 )
 
+type Record struct {
+	arrow.Record
+	SortingColumns []SortingColumn
+}
+
+type SortingColumn struct {
+	ColumnIndex int
+	Direction   schemapb.SortingColumn_Direction
+	NullFirst   bool
+	Order       int
+}
+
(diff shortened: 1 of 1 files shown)


internal/records/record_builder.go
```diff
@@ -20,6 +20,18 @@ const (
 	TagName = "frostdb"
 )
 
+type Record struct {
+	arrow.Record
+	SortingColumns []SortingColumn
+}
+
+type SortingColumn struct {
```
asubiotto, line 28: This seems to be very close to the `arrowutils.SortingColumn` struct. If we can, I would reuse that here to avoid redefining similar structs in different packages (we might want to extract it?) since long term this creates technical debt and confusion.

gernest: done 3755706

gernest:

asubiotto (approved):

PR #677, opened 2024-01-09
Add BenchmarkSortRecord

gernest:
This was supposed to be part of #635

I will work on benchmarking `ReorderRecord` separately, I am not satisfied with `ReorderRecord` api, will think more about it( we pass `memory.Allocator` together with `context.Context`, we need to only pass `context.Context` like what `compute` apis do).


gernest:
Something is wrong with `github.com`. Checks are failing not because of the changes.

thorfour (approved):

PR #681, opened 2024-01-10
Reduce memory allocation in arrowutils.SortRecord

gernest:
This is my last pass before I move on to introduce the changes in `parca`

Benchmark result

```
goos: darwin
goarch: amd64
pkg: github.com/polarsignals/frostdb/pqarrow/arrowutils
             │   old.txt    │           new.txt            │
             │    sec/op    │   sec/op     vs base         │
SortRecord-8   7.492µ ± 10%   6.972µ ± 7%  ~ (p=0.093 n=6)

             │  old.txt   │              new.txt              │
             │    B/op    │    B/op     vs base               │
SortRecord-8   808.0 ± 0%   400.0 ± 0%  -50.50% (p=0.002 n=6)

             │   old.txt   │              new.txt              │
             │  allocs/op  │ allocs/op   vs base               │
SortRecord-8   15.000 ± 0%   8.000 ± 0%  -46.67% (p=0.002 n=6)
```

I really missed `errdefer` in `zig` manual release during `error` returns in annoying.

### DIFF
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
+	"sync"
 
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/apache/arrow/go/v14/arrow/array"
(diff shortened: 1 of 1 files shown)


pqarrow/arrowutils/sort.go
```diff
@@ -197,15 +187,47 @@ func newMultiColSorter(
 					},
 				}
 			default:
+				ms.Release()
```
thorfour, line 190: We could do named returns (which I don't particularly love) but then just have the defer func check if this function is about to return an error and then release ms if it is.

thorfour (approved):

PR #682, opened 2024-01-10
query/physicalplan: Use arrow compute package for binary scalar expressions

metalmatze:
Instead of implementing individual functions for each BinaryScalar operation, we'll now use the arrow `compute` package. It already implements most of the heavy lifting for us. 

Once we pass in the array and the scalar, we can convert the data back to a `*array.Boolean`. Walking this array, we construct the bitmap like we did before. 
It cleans up a lot of duplicate code and removes the need to add new functions for new Arrow types. 

Given the quality of live improvement, the performance impact is fine:
```
name                         old time/op    new time/op    delta
BinaryScalarOperation/==-24    3.39ms ± 2%    3.76ms ± 7%  +11.00%  (p=0.008 n=5+5)
BinaryScalarOperation/!=-24    11.2ms ± 3%    11.4ms ± 1%   +1.64%  (p=0.032 n=5+4)
BinaryScalarOperation/<-24     6.00ms ± 6%    6.25ms ± 3%     ~     (p=0.056 n=5+5)
BinaryScalarOperation/<=-24    7.07ms ± 2%    7.04ms ± 1%     ~     (p=0.841 n=5+5)
BinaryScalarOperation/>-24     7.02ms ± 6%    6.92ms ± 5%     ~     (p=0.690 n=5+5)
BinaryScalarOperation/>=-24    7.78ms ± 3%    7.95ms ± 5%     ~     (p=0.310 n=5+5)

name                         old alloc/op   new alloc/op   delta
BinaryScalarOperation/==-24     509kB ± 0%     643kB ± 0%  +26.23%  (p=0.008 n=5+5)
BinaryScalarOperation/!=-24     532kB ± 0%     665kB ± 0%  +25.12%  (p=0.008 n=5+5)
BinaryScalarOperation/<-24      532kB ± 0%     532kB ± 0%     ~     (all equal)
BinaryScalarOperation/<=-24     532kB ± 0%     532kB ± 0%     ~     (all equal)
BinaryScalarOperation/>-24      532kB ± 0%     532kB ± 0%     ~     (all equal)
BinaryScalarOperation/>=-24     532kB ± 0%     532kB ± 0%     ~     (all equal)

name                         old allocs/op  new allocs/op  delta
BinaryScalarOperation/==-24       263 ± 0%       309 ± 0%  +17.49%  (p=0.008 n=5+5)
BinaryScalarOperation/!=-24       267 ± 0%       313 ± 0%  +17.23%  (p=0.008 n=5+5)
BinaryScalarOperation/<-24        267 ± 0%       267 ± 0%     ~     (all equal)
BinaryScalarOperation/<=-24       267 ± 0%       267 ± 0%     ~     (all equal)
BinaryScalarOperation/>-24        267 ± 0%       267 ± 0%     ~     (all equal)
BinaryScalarOperation/>=-24       267 ± 0%       267 ± 0%     ~     (all equal)
```

### DIFF
--- a/query/physicalplan/binaryscalarexpr.go
+++ b/query/physicalplan/binaryscalarexpr.go
@@ -2,11 +2,13 @@ package physicalplan
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/apache/arrow/go/v14/arrow/array"
+	"github.com/apache/arrow/go/v14/arrow/compute"
 	"github.com/apache/arrow/go/v14/arrow/scalar"
 
 	"github.com/polarsignals/frostdb/query/logicalplan"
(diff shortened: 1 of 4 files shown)


thorfour (approved):

gernest (commented):
This looks like can be delegated to `compute` package . Scalar comparisons are supported.

 `compute.CallFuncion ( ... "equal",left,right)` which returns `*array.Boolean` that we can use to construct the `Bitmap`

**_you can completely ignore my comment, I have no clue what is happening on this part of the code base yet._**

asubiotto:
Interesting. I guess you can even get rid of the type switch in this case and just delegate that to `compute.CallFunction`? I wonder what else in our execution engine we could replace with these compute kernels.

metalmatze:
I'm working on an update with these exact changes. It's going to delete _a lot_ of code. :joy:

metalmatze:
Pushed and updated the title and description of the PR.

gernest:
@metalmatze @asubiotto @thorfour  I take this opportunity to thank you for being awesome and giving me a chance to hack on `frostdb` uninterrupted.

I have a minor surgery on Monday that will take me a couple of weeks to recover. So my silence won't mean I jumped ship, but rather I won't be able to work on my laptop.

I will pickup the slack the moment I get back, I'm quite experienced with `compute` and whole `arrow/**`  stuff, so I will probably focus there and our parquet converter ro see how much improvement we can get.

I have some pending local branches, that I will be glad to share the details with anyone touching these topics maybe they can be helpful.

- Building direct arrow records in logictest
- Integrating `GenericTable` in `parca` Ingester
- Benchmarking `arrowutils.ReorderRecord`  


 shameless plug 😃 **_Please don't forget about about me If you ever need to expand your team, our collaboration has been awesome._**

metalmatze:
@gernest, thank you for all your contributions in the last weeks. They've been totally awesome!
Good luck with the surgery! I wish you a fast recovery! 

We'll definitely let you know if we hire someone (you'll probably see it quickly anyway).

query/physicalplan/binaryscalarexpr.go
```diff
@@ -201,9 +100,52 @@ func BinaryScalarOperation(left arrow.Array, right scalar.Scalar, operator logic
 		panic("TODO: list comparisons unimplemented")
 	}
 
+	switch operator {
+	case logicalplan.OpEq:
+		return ArrayScalarCompute("equal", left, right)
```
thorfour, line 105: Are these consts anywhere in the compute pkg?

gernest: nope, they are hard coded in each kernel when functions are being dynamically registered. We can do better by resolving to these kernel function names `i.e` `(logicalplan.OpEq).String() =="equal"` so we can have constants of our own,  completely  removing those switch case.

metalmatze: I was looking for those constants. You're right that we don't even need those. 

Wow. 😄

thorfour (approved):

PR #688, opened 2024-01-18
rewrote compaction to write to separate files for each part

thorfour:
There is a data loss scenario with the persistent LSM's. 

1. Snapshot happens.
2. WAL is truncated.
3. Another compaction is triggered
     While this compaction is being written to the index file the node crashes/shutsdown and the file becomes corrupted. 
 
This corruption of the single file means we cannot recover any of the other parts in the file which results in data loss because we cannot recreate those parts from the WAL due to truncation.

To fix this I've split the index files up to have a file per part. Now if the above scenario happens we can recover the remaining parts in the level, and only the new writes that caused the compaction need to be recovered from the WAL.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1,7 +1,7 @@
 package frostdb
 
 import (
-	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -11,6 +11,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/oklog/ulid"
 	"github.com/parquet-go/parquet-go"
+	"golang.org/x/exp/slices"
 
 	"github.com/polarsignals/frostdb/dynparquet"
 	"github.com/polarsignals/frostdb/parts"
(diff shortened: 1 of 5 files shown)


compaction.go
```diff
@@ -85,30 +64,38 @@ func (f *FileCompaction) writeRecordsToParquetFile(compact []parts.Part, options
...
-	prevOffset := f.offset
-	f.offset += accountant.n + 8
-	f.ref++
-
-	// Record the file size for recovery.
-	size := make([]byte, 8)
-	binary.LittleEndian.PutUint64(size, uint64(accountant.n))
-	if n, err := f.file.Write(size); n != 8 {
-		return nil, 0, 0, fmt.Errorf("failed to write size to file: %v", err)
+	file, err := os.Open(filepath.Join(f.dir, fmt.Sprintf("%v.parquet", tx)))
```
thorfour, line 93: I don't know if this is going to be tenable to maintain open files. I think we'll hit the open file descriptor limit.

Instead I think we need to change the Part to just maintain the filename and only open when it's read.

thorfour: But also trying to track the lifecycle during a read is going to be challenging

thorfour: So maybe we just leave it as open fd and then control for open fd by adjusting level sizing. wdyt @asubiotto?

asubiotto: I think that LSMs like RocksDB and pebble keep file descriptors around up to a pre-configured limit. IIUC, these file descriptors are kept in a table cache (see https://github.com/facebook/rocksdb/wiki/RocksDB-Overview#table-cache and https://github.com/cockroachdb/pebble/blob/8ce58a8116e95f2db3c2c77f797c1ea40d57d536/table_cache.go#L271).

I think it might be simpler to start by having a well defined descriptor lifecycle. Can we somehow piggy back off of the parquet `Rows` interface to do this? I would have to take a closer look.

asubiotto: Additionally I think we should also take a closer look at level sizing now since as you say this affects the number of files. Specifically, taking RocksDB as inspiration, I think there's a setting called `target_file_size_base` which defaults to 64MB and specifies the size of a single SST file (can also configure this size per level with a `target_file_size_base_multiplier` or something). We could add this option and change compactions to merge overlapping files to limit the number of files per level.

thorfour: Yea I think we can wrap row groups with a reference counter to open and close files accordingly to control the lifecycle of fd.

thorfour: I think I'll leave managing the number of files to a future PR should we feel the need.

thorfour:
Currently reconsidering this. Instead of file per part instead just rotate the level file on snapshot. Thus reducing the number of files and file descriptors. But also adding the ability to recover after snapshot

PR #691, opened 2024-01-20
Pre hash columns before committing to wal

### LINKED ISSUE #685: Move prehash function to happen before write to WAL
Right now we support pre-hashing columns to speed up aggregation by storing the hash of the column directly.

However we perform this prehashing after writing to the WAL, which means during replay we no longer prehash the column.

We should prehash the data before writing to the WAL, that way the WAL has the prehashed data and theirs no problem with directly inserting afterwards.

gernest:
Closes #685

### DIFF
--- a/table.go
+++ b/table.go
@@ -676,11 +676,14 @@ func (t *Table) InsertRecord(ctx context.Context, record arrow.Record) (uint64, 
 	tx, _, commit := t.db.begin()
 	defer commit()
 
-	if err := t.wal.LogRecord(tx, t.name, record); err != nil {
+	preHashedRecord := dynparquet.PrehashColumns(t.schema, record)
+	defer preHashedRecord.Release()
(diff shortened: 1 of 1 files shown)


thorfour (approved):
Thanks!

asubiotto:
I think there's also a TODO we can delete in db.go. I don't see any prehashing in the recovery path though, am I missing something?

gernest:
> I think there's also a TODO we can delete in db.go. I don't see any prehashing in the recovery path though, am I missing something?

@asubiotto can you share the link to this so I can take a look?

asubiotto:
https://github.com/polarsignals/frostdb/blob/67e37c6b4582b0a8b7f592d909aadca299b7b9c7/db.go#L913

gernest:
thanks

Issue #672, opened 2024-01-04
Use compute.FilterRecord instead of bitmask and manually filtering


garrensmith:
A lot of the filters use a bitmask and then manually filtering the record batch. I think this could be simplified by using the built in [`compute.FilterRecord`](https://pkg.go.dev/github.com/apache/arrow/go/v12@v12.0.1/arrow/compute#FilterRecordBatch) to do all the work. I'm not 100% if it would improve performance or not, but I think it should give a memory improvement.

thorfour:
Sorry, apparently fat fingered the close button.

I played around with this today
```go
 func filter(ctx context.Context, pool memory.Allocator, filterExpr BooleanExpression, ar arrow.Record) (arrow.Record, bool, error) {
      bitmap, err := filterExpr.Eval(ar)
      if err != nil {
          return nil, true, err
      }

      if bitmap.IsEmpty() {
          return nil, true, nil
      }

      // Construct filter array
      // NOTE: this is intermediary right now. The Eval function should return a boolean array instead so we can directly pass it in.
      bldr := array.NewBooleanBuilder(pool)
      defer bldr.Release()
      for i := 0; i < int(ar.NumRows()); i++ {
          bldr.Append(bitmap.Contains(uint32(i)))
      }
      filterArr := bldr.NewArray()
      defer filterArr.Release()

      result, err := compute.FilterRecordBatch(ctx, ar, filterArr, compute.DefaultFilterOptions())
      if err != nil {
          return nil, true, err
      }
      return result, false, nil
  }
```

And ended up with a panic 
```bash
--- FAIL: Test_DB_All (0.00s)
    db_test.go:2085:
        	Error Trace:	/Users/thor/go/src/github.com/polarsignals/frostdb/db_test.go:2085
        	Error:      	Received unexpected error:
        	            	not implemented: function 'array_take' has no kernel matching input types (dictionary<values=utf8, indices=uint32, ordered=false>, uint16)
        	Test:       	Test_DB_All
```

Looks like it's not implemented for all types.

garrensmith:
That is unfortunate. Should we open an issue on apache arrow for this?

thorfour:
Yea and link it here so we know if/when we can actually implement this.

gernest:
> Yea and link it here so we know if/when we can actually implement this.

No need for this. We can in fact implement it now. Basically, this is how filter on records works

- build array of indices containing `rows` you wan't to choose
[...]

PR #692, opened 2024-01-20
Rename ReorderRecord to Take

gernest:
The new name reflects what the function does, we are not ordering anything we are just taking specific rows.


gernest:
I almost made a mistake to recommend duplicate implementation on ticket #672 because it was impossible to discover.

thorfour (approved):

PR #695, opened 2024-01-21
Fix memory leak in arrowutils.Take

gernest:
`r` and `indices` are externally managed resources. Owning them increases ref
 count ,since we never releases them they will leak when `r` has no dictionary field.

 There is no need to own these resources inside `Take`.

 This is part of #672


thorfour (approved):

metalmatze:
Thanks for finding and fixing this! :+1:

PR #700, opened 2024-01-22
Remove TODO since we prehash before wal now

gernest:
After reading the `TODO` and inspecting the code. I think it is safe to remove it now. From my understanding the `TODO` reminds us that the record added to `lsm` is not pre hashed and recommends we prehash before writing to wal which we currently do.

cc @asubiotto


asubiotto (approved):
LGTM. But if we were previously not prehashing before writing to the WAL and not prehashing here in recovery, doesn't that imply that the record inserted here would not have a prehashed column? No problem, just trying to understand the previous behavior.

Issue #208, opened 2022-09-23
Move aggregation alias out into projections


metalmatze:
_Originally posted by @asubiotto in https://github.com/polarsignals/frostdb/pull/202#discussion_r976278924_

brancz:
I'm working on a PR that actually fixes this.

asubiotto:
@brancz did you end up fixing this?

brancz:
Yes! Was fixed a few months ago.

Issue #687, opened 2024-01-17
unsupport avg on a single dynamic column 


jicanghaixb:
Thank you for fixing the issue #686 

but I found new issue，when i use avg aggregation on dynamic column  return new message:
<details><summary>Details</summary>
<p>
select avg(value) group by name

sum column for average projection for column value not found

</p>
</details> 

I found the problem lies in the internal conversion of avg to sum and count, there have two problem: 
issue 1:
when we use sum(value),count(value), than finalStage is true, count cover sum, because 
 when executing count, aggregate.dynamicAggregationsConverted[field.Name] is ok.
physicalplan/aggregation.go
```go
if _, ok := aggregate.dynamicAggregationsConverted[field.Name]; !ok {
			for _, col := range aggregate.dynamicAggregations {
				if a.finalStage {
					if col.expr.MatchColumn(field.Name) {
						// expand the aggregate.aggregations with a final concrete column aggregation.
						columnToAggregate = append(columnToAggregate, nil)
						aggregate.aggregations = append(aggregate.aggregations, Aggregation{
							expr:       logicalplan.Col(field.Name),
							dynamic:    true,
							resultName: resultNameWithConcreteColumn(col.function, field.Name),
							function:   col.function,
						})
						aggregate.dynamicAggregationsConverted[field.Name] = struct{}{}
					}
				} else {
					// If we're aggregating the raw data we need to find the columns by their actual names for now.
					if col.expr.MatchColumn(field.Name) {
						// expand the aggregate.aggregations with a concrete column aggregation.
						columnToAggregate = append(columnToAggregate, nil)
						aggregate.aggregations = append(aggregate.aggregations, Aggregation{
							expr:       logicalplan.Col(field.Name),
							dynamic:    true,
							resultName: field.Name, // Don't rename the column yet, we'll do that in the final stage. Dynamic aggregations can't match agains't the pre-computed name.
							function:   col.function,
						})
						aggregate.dynamicAggregationsConverted[field.Name] = struct{}{}
					}
				}
			}
		}
``` 
issue 2:
physicalplan/project.go
averageProjection need special handling of the matching rules for columnSum and columnCount
```go
func (a *averageProjection) Project(mem memory.Allocator, r arrow.Record) ([]arrow.Field, []arrow.Array, error) {
	columnName := a.expr.Name()
	resultName := "avg(" + columnName + ")"
	if avgExpr, ok := a.expr.(*logicalplan.AverageExpr); ok {
[...]

PR #702, opened 2024-01-23
Nested arithmetic projections

brancz:
This allows us to perform things like sum(value)/count(value) or any arbitrary combination of aggregations and arithmetic expressions.

To get this to work there were a couple of changes necessary:

1) Implement #208 
2) Make query plans dumber, and query optimizations only optimizations and not required for correctness.
3) Make query builders (the dataframe and SQL APIs) smarter to insert logical query plans correctly for the desired behaviors. For example, when there are arithmetic expressions that use aggregations a projection query plan is inserted after the aggregation to perform that calculation. Likewise, if there is a column that is used in a group by or as input to an aggregation, then a projection to perform that calculation needs to be inserted before the aggregation layer. In the SQL layer this was implemented, in the dataframe API it is expected that the user builds the query layers appropriately.

I wanted to open this early so it's not just on my computer and get some early feedback while I finish it up. A couple of things are still missing:

- [x] Not all tests are passing yet and there is some non-correct behavior for expressions like `(timestamp/1000)*1000 as timestamp_bucket`
- [x] Average queries

Side note: What's nice about these changes is that we could now definitively determine the input and output schema of a logical plan query layer which we could use eg. for validations.


jicanghaixb:
I'm really looking forward to this feature, sum(value) maybe is  float64, count(value)  is int64, so sum(value)/count(value) need to support mathematical calculations of different types. 

I have also made some attempts, and below are my test cases,the modified code is quite ugly, and last two test cases have leaked memory, can't pass require.Equal(t, 0, fdb.allocator.Allocated()). I haven't found the reason yet:
```
createtable schema=custom
----

insert cols=(name, duration,  labels.code, labels.instance, labels.method, labels.status, labels.upstream_code, timestamp, trace,  value.body_send, value.header_time,message, value.request)
cdn 1 200 127.0.0.1 GET HIT  - 1699871519098 legacyplatforms.info 1664782 0.057256 /Triple-buffered.htm 1
cdn 1 200 127.0.0.1 GET MISS  200  1699871520098 productinnovative.com 1769821  3.836798 /solution-oriented.php 1
cdn 1 200 127.0.0.1 HEAD MISS 200 1699871521098 regionalone-to-one.biz  1581608 1.655158 /intangible-local.jpg 1
cdn 1 200 127.0.0.1 GET MISS 200 1699871522098 legacysticky.info  1786570 4.581297 /monitoring-Virtual.jpg 1
cdn 1 200 127.0.0.1 GET HIT  - 1699871523098 legacyback-end.com 120018 0.156603 /projection-Versatile.js 1
cdn 1 200 127.0.0.1 GET HIT - 1699871524098 dynamicweb-enabled.com  621713 0.154511 /incremental/core/Cross-platform/conglomeration.css 1
cdn 1 200 127.0.0.1 GET HIT  - 1699871525098 futureembrace.org  2130660 0.286799 /full-range-4th%20generation/Customizable_interface%20Business-focused.htm 1
cdn 1 200 127.0.0.1 PUT HIT  - 1699871526098 districtenable.org  195706 0.171824 /Face%20to%20face.gif 1
cdn 1 200 127.0.0.1 GET MISS 200 1699871527098 globalreal-time.biz 1970713 4.200371 /Total_Implemented-budgetary%20management.js 1
cdn 1 200 127.0.0.1 HEAD MISS 200 1699871528098 internationalmission-critical.name  2836182 2.633382 /Re-contextualized.gif 1
cdn 1 200 127.0.0.1 GET MISS 200 1699871529098 futuremindshare.org  1379273 2.028894 /archive/Implemented.htm 1
cdn 1 200 127.0.0.1 GET HIT - 1699871530098 futurerobust.io  825589 0.384408 /methodology.jpg 1
cdn 1 200 127.0.0.1 PUT MISS 200 1699871531098 futuresystems.biz  2320748 4.660421 /initiative.css 1
cdn 1 200 127.0.0.1 GET MISS 200 1699871532098 futuresynergize.name  166355 2.627923 /4th%20generation-national/emulation.js 1
cdn 1 200 127.0.0.1 GET HIT  - 1699871533098 principalsynthesize.name  1775497 0.435607 /Versatile/4th%20generation/Devolved-policy/core.htm 1
cdn 1 200 127.0.0.1 POST HIT - 1699871534098 centralsticky.org  2669659 0.436545 /24%20hour/responsive-Graphical%20User%20Interface.svg 1
cdn 1 200 127.0.0.1 GET MISS 200 1699871535098 principalcommunities.com  2939218 3.630523 /Re-engineered-Multi-tiered-Operative.svg 1
cdn 1 200 127.0.0.1 GET HIT - 1699871536098 districtempower.name  1668660 0.258968 /Profound-bottom-line.png 1
cdn 1 200 127.0.0.1 DELETE MISS 200 1699871537098 humane-business.name  2926877 0.279401 /flexibility/website.php 1
cdn 1 200 127.0.0.1 GET MISS 200 1699871538098 dynamicimplement.com  1240638  1.376430 /context-sensitive-Re-contextualized.htm 1
----

exec
select sum(value), avg(value) group by labels.instance,labels.status
----
labels.instance  labels.status  sum(value.body_send)  sum(value.header_time)  sum(value.request)  avg(value.body_send)  avg(value.header_time)  avg(value.request)
127.0.0.1        MISS           20918003.00           31.51                   11.00               1901636.64            2.86                    1.00
127.0.0.1        HIT            11672284.00           2.34                    9.00                1296920.44            0.26                    1.00

exec
select sum(value.body_send) as body_send, sum(value.request) as request, body_send / request as body_send_rate  group by labels.instance,labels.status
----
labels.instance  labels.status  body_send    request  body_send_rate
127.0.0.1        MISS           20918003.00  11.00    1901636.64
127.0.0.1        HIT            11672284.00  9.00     1296920.44

exec
select sum(value.request), avg(value.body_send),max(value.header_time) group by labels.instance,labels.status
----
labels.instance  labels.status  sum(value.request)  avg(value.body_send)  max(value.header_time)
127.0.0.1        MISS           11.00               1901636.64            4.66
127.0.0.1        HIT            9.00                1296920.44            0.44

exec
select min(value.body_send), max(value.body_send), sum(value.body_send), count(value.body_send), avg(value.body_send) group by labels.instance,labels.status
----
labels.instance  labels.status  min(value.body_send)  max(value.body_send)  sum(value.body_send)  count(value.body_send)  avg(value.body_send)
127.0.0.1        MISS           166355.00             2939218.00            20918003.00           11.00                   1901636.64
127.0.0.1        HIT            120018.00             2669659.00            11672284.00           9.00                    1296920.44

exec
select labels.status, min(value.body_send), max(value.body_send), sum(value.body_send), count(value.body_send), avg(value.body_send) group by labels.instance,labels.status
----
labels.status  min(value.body_send)  max(value.body_send)  sum(value.body_send)  count(value.body_send)  avg(value.body_send)
MISS           166355.00             2939218.00            20918003.00           11.00                   1901636.64
HIT            120018.00             2669659.00            11672284.00           9.00                    1296920.44

exec
select labels.status, min(value.body_send), max(value.body_send), sum(value.body_send)/count(value.body_send) as avg, avg(value.body_send) group by labels.instance,labels.status
----
labels.status  min(value.body_send)  max(value.body_send)  avg         avg(value.body_send)
MISS           166355.00             2939218.00            1901636.64  1901636.64
HIT            120018.00             2669659.00            1296920.44  1296920.44

exec
select labels.status, min(value.body_send), max(value.body_send), sum(value.body_send)/count(value.body_send) as avg, avg(value.body_send)  having avg > 1296921 group by labels.instance,labels.status
----
labels.status  min(value.body_send)  max(value.body_send)  avg         avg(value.body_send)
MISS           166355.00             2939218.00            1901636.64  1901636.64

exec
select labels.status, min(value.body_send), max(value.body_send), sum(value.body_send), avg(value.body_send) * count(value.body_send) as sum group by labels.instance,labels.status
----
labels.status  min(value.body_send)  max(value.body_send)  sum(value.body_send)  sum
MISS           166355.00             2939218.00            20918003.00           20918003.00
HIT            120018.00             2669659.00            11672284.00           11672284.00

exec
select  avg(value.body_send) as avg_1x , avg(value.body_send) * 2 as avg_2x, avg(value.body_send) * 3 as avg_3x  group by labels.instance,labels.status
----
labels.instance  labels.status  avg_1x      avg_2x      avg_3x
127.0.0.1        MISS           1901636.64  3803273.27  5704909.91
127.0.0.1        HIT            1296920.44  2593840.89  3890761.33

exec
select sum(value) group by labels
----
labels.code  labels.instance  labels.method  labels.status  labels.upstream_code  sum(value.body_send)  sum(value.header_time)  sum(value.request)
200          127.0.0.1        DELETE         MISS           200                   2926877.00            0.28                    1.00
200          127.0.0.1        GET            HIT            -                     8806919.00            1.73                    7.00
200          127.0.0.1        GET            MISS           200                   11252588.00           22.28                   7.00
200          127.0.0.1        HEAD           MISS           200                   4417790.00            4.29                    2.00
200          127.0.0.1        POST           HIT            -                     2669659.00            0.44                    1.00
200          127.0.0.1        PUT            HIT            -                     195706.00             0.17                    1.00
200          127.0.0.1        PUT            MISS           200                   2320748.00            4.66                    1.00

exec
select sum(value) group by labels.method, second(5)
----
labels.method  timestamp      sum(value.body_send)  sum(value.header_time)  sum(value.request)
DELETE         1699871537098  2926877.00            0.28                    1.00
GET            1699871519098  1664782.00            0.06                    1.00
GET            1699871523098  4298122.00            8.73                    4.00
GET            1699871525098  5480646.00            6.52                    3.00
GET            1699871530098  2767441.00            3.45                    3.00
GET            1699871536098  5848516.00            5.27                    3.00
HEAD           1699871521098  1581608.00            1.66                    1.00
HEAD           1699871528098  2836182.00            2.63                    1.00
POST           1699871534098  2669659.00            0.44                    1.00
PUT            1699871526098  195706.00             0.17                    1.00
PUT            1699871531098  2320748.00            4.66                    1.00

exec
select topk(sum(value.body_send),3) group by labels.method, second(300)
----
labels.method  timestamp      topk(sum(value.body_send)_3)
GET            1699871519098  20059507.00
HEAD           1699871521098  4417790.00
DELETE         1699871537098  2926877.00

exec
select name,topk(value.body_send,3) group by labels
----
name    topk(value.body_send_3)
cdn     2939218.00
cdn     2926877.00
cdn     2836182.00

exec
select name,bottomk(value.body_send,3), topk(value.header_time,3) as top group by labels
----
name    bottomk(value.body_send_3)  name    top
cdn     120018.00                   cdn     4.66
cdn     166355.00                   cdn     4.58
cdn     195706.00                   cdn     4.20

exec
select name,labels.method,IF(labels.method = "GET",1,2) as result
----
name    labels.method  result
cdn     DELETE         2.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     GET            1.00
cdn     HEAD           2.00
cdn     HEAD           2.00
cdn     POST           2.00
cdn     PUT            2.00
cdn     PUT            2.00

exec
select sum(value.body_send) > 20000000  as body_send,labels group by labels.instance,labels.status
----
body_send  labels.instance  labels.status
true       127.0.0.1        MISS
false      127.0.0.1        HIT

exec
select IF(sum(value.body_send) > 20000000, "warning", "normal") as body_send,labels group by labels.instance,labels.status
----
labels.instance  labels.status  body_send
127.0.0.1        MISS           warning
127.0.0.1        HIT            normal

```

jicanghaixb:
The following are the test cases I implemented after modifying the code, hoping that frost db can support them natively:
```
createtable schema=default
----

insert cols=(labels.label1, timestamp, value)
value1 1 2
value1 3 4
value1 5 6
----

exec
select value * timestamp
----
2
12
30

exec
select value * timestamp limit 0,2
----
2
12

exec
select value * timestamp limit 1,2
----
2
12

exec
select value * timestamp limit 3,2
----

exec
select value * timestamp limit 2,2
----
12

exec
select value * 2
----
4
8
12

exec
select 2 * 3
----
6
6
6

exec
select value * (2 * 3)
----
12
24
36

exec
select value * (timestamp * 2)
----
4
24
60

exec
select value * timestamp * 2
----
4
24
60

exec
select value * timestamp + 2 - 1
----
3
13
31

exec
select value * timestamp * (2 - 1)
----
2
12
30

exec
select value, value * timestamp * (2 - 1)
----
2       2
4       12
6       30

exec
select timestamp, value * timestamp * (2 - 1) where value > 2
----
3       12
5       30

exec
select value, timestamp  * value * (2 - 1)  where value > 2
----
4       12
6       30

exec
select value , value  * timestamp * (2 - 1) as result where value > 2
----
4       12
6       30

exec
select value as result, value  * timestamp * (2 - 1)  where value > 2
----
4       12
6       30

exec
select value , value  * timestamp * (2 - 1) as result  having  value > 4
----
6       30

exec
select value as result, value  * timestamp * (2 - 1)  having result > 4
----
6       30

exec
select sum(value), count(value) group by labels.label1
----
value1  12      3

exec
select   max(value) - min(value) , sum(value) / count(value)  group by labels.label1
----
4       4

exec
select  labels.label1, max(value) - min(value) , sum(value) / count(value)  group by labels.label1
----
value1  4       4

exec
select  labels.label1, (max(value) - min(value)) /2  , sum(value) / count(value)  group by labels.label1
----
value1  2       4

exec
select  labels.label1, (max(value) - min(value)) /2  , sum(value) / count(value) , sum(value) / count(value) * 2 group by labels.label1
----
value1  2       4       8

```

logictest/logic_test.go
```diff
@@ -20,7 +20,7 @@ const testdataDirectory = "testdata"
 
 type frostDB struct {
 	*frostdb.DB
-	allocator *query.LimitAllocator
```
gernest, line 23: nice, maybe we should fix this on `main` branch as. well?

logictest/logic_test.go
```diff
@@ -20,7 +20,7 @@ const testdataDirectory = "testdata"
...
+	allocator *memory.CheckedAllocator
```
gernest, line 23: nice! maybe we should fix this on `main` branch too ?

jicanghaixb: I tested this test case,  I found PredicateFilter has a memory leak, but I still haven't found the cause:

TableScan [concurrent] - Projection (2 * 3 as va) - PredicateFilter (va > 6) - Synchronizer

```
        exec [0 args]
        select 2 * 3 as va having va > 6
        ----
    checked_allocator.go:186: LEAK of 192 bytes FROM
        	github.com/apache/arrow/go/v14/arrow/array.(*Int64Builder).newData+2f
        	github.com/asubiotto/arrow/go/v14@v14.0.0-20231129090719-b321865d34e9/arrow/array/numericbuilder.gen.go:181
    checked_allocator.go:186: LEAK of 64 bytes FROM
        	github.com/apache/arrow/go/v14/arrow/array.(*builder).init+af
github.com/asubiotto/arrow/go/v14@v14.0.0-20231129090719-b321865d34e9/arrow/array/builder.go:142
    logic_test.go:140: invalid memory size exp=0, got=256
```

thorfour: If you can run that again with `ARROW_CHECKED_MAX_RETAINED_FRAMES=30` that way we'll have more of a stacktrace dumped to determine where the leak is coming from.

gernest: > I think #714 is already the investigation of this

Yes, we need the logic test that trigger it, the reasoning on the ticket seems sensible to me but I will wait for @jicanghaixb PR to be sure.

I don't think it is directly related to this PR but rather a general case so It shouldn't block anything for the current PR.

gernest: @jicanghaixb thanks, can you do this in your local modification

```bash
git diff >memory_leak
```

And share with me the content of `memory_leak` file so I can check locally ?


It will be nice  to move our discussion to #714 .  As I'm sure this exists in `main` and the leak was not a result of changes introduced here(Still I will confirm after seeing your `diff`)

brancz:
This ended up being a way bigger PR than I thought initially, but it's now ready for review!

logictest/testdata/exec/aggregate/math
metalmatze, line 1: Fantastic to see! 🎉

query/logicalplan/builder.go
```diff
@@ -118,21 +127,86 @@ func (b Builder) Distinct(
...
+
+			var (
+				countExpr Expr = count
+				aggType   arrow.DataType
+			)
+			aggType, err = agg.Expr.DataType(b.plan)
+			// intentionally not handling the error here, as it will be handled
+			// in the build function.
+			if !arrow.TypeEqual(aggType, arrow.PrimitiveTypes.Int64) {
+				countExpr = Convert(countExpr, aggType)
```
metalmatze, line 158: This is the only place where a type conversion is happening implicitly, right? 
This only happens for averaging.

metalmatze (commented):
I'm really looking forward to this landing!

query/logicalplan/builder.go
brancz: Correct, it's just an artifact of average expressions being resolved already by the builder of the logical plan.

query/physicalplan/arithmeticexpr.go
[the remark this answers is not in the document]
brancz: nope, good catch

brancz:
I'm testing everything with Parca, and there seem to be a few things that need fixing, and I'll write some tests for those things along the way.

query/logicalplan/builder.go
jicanghaixb: if value is dynamic float64 field,  below code has problem, because there is no specific dynamic column information in the schema
example:
select avg(value.request) group by name

```go
							t, err := agg.Expr.DataType(v.schema)
							if err != nil {
								v.err = err
								return nil, true
							}

							var typedCount logicalplan.Expr = count
							if !arrow.TypeEqual(t, arrow.PrimitiveTypes.Int64) {
								typedCount = logicalplan.Convert(count, arrow.PrimitiveTypes.Float64)
							}
```

jicanghaixb: BinaryOperation may also require a type conversion to support the following test cases:

```
exec
select value * timestamp as int64, floatvalue * timestamp as float64
----
int64   float64
2       2.000000
12      12.000000
30      30.000000
``` 

```go
case *ast.BinaryOperationExpr:
		// Note that we're resolving exprs as a stack, so the last two
		// expressions are the leaf expressions.
		rightExpr, newExprs := pop(v.exprStack)
		leftExpr, newExprs := pop(newExprs)
		v.exprStack = newExprs

		var frostDBOp logicalplan.Op
		switch expr.Op {
		case opcode.GT:
			frostDBOp = logicalplan.OpGt
		case opcode.LT:
			frostDBOp = logicalplan.OpLt
		case opcode.GE:
			frostDBOp = logicalplan.OpGtEq
		case opcode.LE:
			frostDBOp = logicalplan.OpLtEq
		case opcode.EQ:
			frostDBOp = logicalplan.OpEq
		case opcode.NE:
			frostDBOp = logicalplan.OpNotEq
		case opcode.Plus:
			frostDBOp = logicalplan.OpAdd
		case opcode.Minus:
			frostDBOp = logicalplan.OpSub
		case opcode.Mul:
			frostDBOp = logicalplan.OpMul
		case opcode.Div:
			frostDBOp = logicalplan.OpDiv
		case opcode.LogicAnd:
			v.exprStack = append(v.exprStack, logicalplan.And(leftExpr, rightExpr))
			return nil
		case opcode.LogicOr:
			v.exprStack = append(v.exprStack, logicalplan.Or(leftExpr, rightExpr))
			return nil
		}
		left, _ := leftExpr.DataType(v.schema)
		right, _ := rightExpr.DataType(v.schema)
		if left != nil && right != nil {
			if !arrow.TypeEqual(left, right) {
				if left.ID() == arrow.PrimitiveTypes.Int64.ID() {
					leftExpr = logicalplan.Convert(leftExpr, arrow.PrimitiveTypes.Float64)
				}
				if right.ID() == arrow.PrimitiveTypes.Int64.ID() {
					rightExpr = logicalplan.Convert(rightExpr, arrow.PrimitiveTypes.Float64)
				}
			}
		}

		v.exprStack = append(v.exprStack, &logicalplan.BinaryExpr{
			Left:  leftExpr,
			Op:    frostDBOp,
			Right: rightExpr,
		})
```

jicanghaixb:
physicalplan/project.go function: DivInt64s, DivFloat64s,DivInt32s 
```go
func DivInt32s(mem memory.Allocator, left, right *array.Int32) *array.Int32 {
	res := array.NewInt32Builder(mem)
	defer res.Release()

	res.Resize(left.Len())

	for i := 0; i < left.Len(); i++ {
		res.Append(left.Value(i) / right.Value(i))
	}

	return res.NewInt32Array()
}
``` 
does the code need to check if the denominator is 0? Avoid panic
```go
func DivInt32s(mem memory.Allocator, left, right *array.Int32) *array.Int32 {
	res := array.NewInt32Builder(mem)
	defer res.Release()

	res.Resize(left.Len())

	for i := 0; i < left.Len(); i++ {
		if right.Value(i) == 0 {
			res.AppendNull()
		} else {
			res.Append(left.Value(i) / right.Value(i))
		}
	}

	return res.NewInt32Array()
}
```

jicanghaixb:
testdata/plan/aggregate/aggregate
```
exec
explain select labels.label1, (max(value) - min(value)) / 2, sum(value) / count(value), (sum(value) / count(value)) * 2 group by labels.label1
----
TableScan [concurrent] - Projection (labels.label1, value) - HashAggregate (max(value),min(value),sum(value),count(value),sum(value),count(value) by labels.label1) - Synchronizer - HashAggregate (max(value),min(value),sum(value),count(value),sum(value),count(value) by labels.label1) - Projection (labels.label1, max(value) - min(value) / 2, sum(value) / count(value), sum(value) / count(value) * 2)

``` 
HashAggregate existence of duplicate aggregation，we need perform deduplication steps to reduce resource consumption

sqlparse/visitor.go
```go
					for _, agg := range aggCollector.aggregations {
						if _, ok := includedPreprojections[agg.Expr.Name()]; !ok {
							preProjections = append(preProjections, agg.Expr)
							// The same expression can be used in multiple
							// aggregations, but we only need to project it
							// once.
							includedPreprojections[agg.Expr.Name()] = struct{}{}
						}
						if _, ok := includedAggregations[agg.Name()]; !ok {
							aggregations = append(aggregations, agg)
							includedAggregations[agg.Name()] = struct{}{}
						}
					}
``` 

logicalplan/builder.go
resolvedAggExpr may also need to consider deduplication 
```go
return Builder{
		err: err,
		plan: &LogicalPlan{
			Projection: &Projection{
				Exprs: append(groupExprs, projectExprs...),
			},
			Input: &LogicalPlan{
				Aggregation: &Aggregation{
					GroupExprs: groupExprs,
					AggExprs:   resolvedAggExpr,
				},
				Input: b.plan,
			},
		},
	}
```

jicanghaixb:
Can we solve the #687  issue synchronously? when i use dynamic value, there is an issue of aggregation coverage
```
exec
select labels.instance,sum(value) group by labels.instance
----
labels.instance  sum(value.body_send)  sum(value.request)
127.0.0.1        19.800000             18.000000
127.0.0.2        24.200000             22.000000

exec
select labels.instance,sum(value),count(value) group by labels.instance
----
labels.instance  sum(value.body_send)  sum(value.request)  count(value.body_send)  count(value.request)
127.0.0.1        19.800000             18.000000           8                       8
127.0.0.2        24.200000             22.000000           8                       8

``` 

second test case can't pass, return blow result, count(value) cover sum(value)
```
        labels.instance  sum(value.body_send)  sum(value.request)  count(value.body_send)  count(value.request)
        127.0.0.1        8                     8                   8                       8
        127.0.0.2        8                     8                   8                       8
``` 

The following code is my temporary solution to this coverage issue, resultName add col.function.String()
physicalplan/aggregate.go
```go
for _, col := range aggregate.dynamicAggregations {
				if a.finalStage {
					if col.expr.MatchColumn(field.Name) {
						// expand the aggregate.aggregations with a final concrete column aggregation.
						if strings.HasSuffix(field.Name, col.function.String()) {
							columnToAggregate = append(columnToAggregate, nil)
							aggregate.aggregations = append(aggregate.aggregations, Aggregation{
								expr:       logicalplan.Col(field.Name),
								dynamic:    true,
								resultName: resultNameWithConcreteColumn(col.function, strings.TrimSuffix(field.Name, col.function.String())),
								function:   col.function,
							})
							aggregate.dynamicAggregationsConverted[field.Name] = struct{}{}
						}
					}
				} else {
					// If we're aggregating the raw data we need to find the columns by their actual names for now.
					if col.expr.MatchColumn(field.Name) {
						// expand the aggregate.aggregations with a concrete column aggregation.
						columnToAggregate = append(columnToAggregate, nil)
						aggregate.aggregations = append(aggregate.aggregations, Aggregation{
							expr:       logicalplan.Col(field.Name),
							dynamic:    true,
							resultName: field.Name + col.function.String(), // Don't rename the column yet, we'll do that in the final stage. Dynamic aggregations can't match agains't the pre-computed name.
							function:   col.function,
						})
						aggregate.dynamicAggregationsConverted[field.Name] = struct{}{}
					}
				}
			}
```

brancz:
I would prefer fixing the bug in a separate PR, this is already quite large.

> HashAggregate existence of duplicate aggregation，we need perform deduplication steps to reduce resource consumption

What do you mean? The first one is executed concurrently, the second on merges the concurrent ones into a single result.

---

Good call about the zero division, I believe the correct thing is `-1/0 = -Inf`, `1/0 = +Inf`, and `0/0 = NaN`. Not sure if arrow has representations for these though.

jicanghaixb:
> HashAggregate

Ok, no problem, users can optimize it because HashAggregates are in an intermediate state in some queries and not the final result. Therefore, HashAggregates can be deduplicated to avoid duplicate aggregates with the same columns

query/physicalplan/project.go
```diff
@@ -380,6 +498,96 @@ func (p plainProjection) Project(_ memory.Allocator, ar arrow.Record) ([]arrow.F
...
+		default:
+			return nil, fmt.Errorf("unsupported conversion from %s to %s", c.DataType(), p.expr.Type)
+		}
+	default:
+		return nil, fmt.Errorf("unsupported conversion from %s to %s", c.DataType(), p.expr.Type)
+	}
+}
+
+func convertInt64ToFloat64(mem memory.Allocator, c *array.Int64) *array.Float64 {
+	defer c.Release()
```
asubiotto, line 529: nit: the caller should release this

asubiotto (commented):
Overall looks really good

query/physicalplan/project.go
jicanghaixb: > nit: the caller should release this

need release, example:  select value * timestamp * 2 , floatvalue * convert(timestamp * 2, float64)

because timestamp * 2 is new array from binaryExprProjection, caller can't release it, so convertInt64ToFloat64 need to release it.

jicanghaixb:
Frostdb will become an excellent project in the field of data analysis. The current implementation is already perfect, and there are still some issues with dynamic value aggregation (as mentioned in the comments above). 
Although the aggregation function supports dynamic column aggregation, the following code needs to be easily modified to support the following test cases
physicalplan/project.go
```go
	case *logicalplan.AggregationFunction:
		switch e.Expr.(type) {
		case *logicalplan.DynamicColumn:
			return dynamicProjection{
				expr: logicalplan.DynCol(fmt.Sprintf("%s(%s", e.Func.String(), e.Expr.Name())),
			}, nil
		default:
			return plainProjection{
				expr: logicalplan.Col(e.Name()),
			}, nil
		}
``` 


```
exec
select labels.instance,sum(value) group by labels.instance
----
labels.instance  sum(value.body_send)  sum(value.request)
127.0.0.1        22.000000             20.000000
127.0.0.2        22.000000             20.000000
```

dynparquet/schema.go
```diff
@@ -135,15 +136,46 @@ func (s *Schema) IsDynamicColumn(column string) bool {
...
+	if !colDef.Dynamic {
+		return ColumnDefinition{}, false
 	}
-	return s.columns[s.columnIndexes[column[0:periodPosition]]].Dynamic
+
+	return colDef, true
+}
+
+// FindColumn returns a column definition for the column passed.
+func (s *Schema) FindColumn(column string) (ColumnDefinition, bool) {
```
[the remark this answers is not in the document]
brancz: they're each slightly different, I added some comments .. while it's probably possible to "abstract" I find that pretty hard to reason about and the code "duplication" is ~8 lines .. are you ok with keeping it as is?

sqlparse/visitor.go
```diff
@@ -19,20 +22,26 @@ type astVisitor struct {
...
 	exprStack []logicalplan.Expr
 }
 
 var _ ast.Visitor = &astVisitor{}
 
-func newASTVisitor(builder query.Builder, dynColNames []string) *astVisitor {
+func newASTVisitor(
+	builder query.Builder,
+	dynColNames []string,
+	schema *parquet.Schema,
```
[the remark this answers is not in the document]
brancz: ah yes, good catch!

pqarrow/arrow.go
```diff
@@ -61,7 +61,15 @@ func ParquetSchemaToArrowSchema(ctx context.Context, schema *parquet.Schema, opt
...
-			dataType, err := distinctExpr.DataType(schema)
+			// Usually we would pass the logical query plan as the data type
+			// finder, but we're here because of an intended layering
+			// violation, which is pushing distinct queries down to the scan
+			// layer. In this case there are no other possible physical types
+			// other than the actual schema, so we can just implement a
+			// simplified version of the type finder that doesn't need to
+			// traverse the logical plan, since this is already the physical
+			// scan layer execution.
+			dataType, err := distinctExpr.DataType(&exprTypeFinder{s: s})
```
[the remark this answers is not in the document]
brancz: Unfortunately, when implementing it, I realized why a parquet schema is insufficient here. The problem arises when a concrete parquet schema does not have any instances of a dynamic column, then we can't infer the type as there are no concrete columns to infer the type from.

query/physicalplan/project.go
asubiotto: Not sure I understand. This array passed into `convertInt64ToFloat64` is retained somewhere (I believe in the call to `Project` in `convertProjection.Project`. I'm saying that the ownership for that array belongs to the `convertProjection` operator, so it should `Release` it in a defer (i.e. here: https://github.com/polarsignals/frostdb/pull/702/files/8296f351dfd060c9dbb4bc055e6f9d4b054ea184#diff-9cc015062538372988d7c41909b5a31a465ffe061d4d70bf5f24e41fdc8044afR543)

pqarrow/arrow.go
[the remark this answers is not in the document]
asubiotto: Hmm, I guess this would be something to solve at a later date when implementing https://github.com/polarsignals/frostdb/issues/366

dynparquet/schema.go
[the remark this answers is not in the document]
asubiotto: Sure

query/physicalplan/project.go
brancz: I agree, also prevents us from making mistakes in the "lower" functions.

query/physicalplan/project.go
```diff
@@ -95,39 +106,123 @@ func (b binaryExprProjection) Project(mem memory.Allocator, ar arrow.Record) ([]
 	case *array.Int64:
 		switch b.expr.Op {
 		case logicalplan.OpAdd:
-			return []arrow.Field{leftFields[0]}, []arrow.Array{AddInt64s(mem, leftArray, rightArray.(*array.Int64))}, nil
+			return []arrow.Field{{
+					Name:     b.expr.Name(),
```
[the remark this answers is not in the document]
brancz: Oh good catch, the field is exactly the same in each case! I think I was focusing so much on the other parts having to be different that I didn't pay attention to this.

pqarrow/arrow.go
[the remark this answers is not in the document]
brancz: agreed

asubiotto (approved):

jicanghaixb:
> The following code needs to be optimized and implemented with exception protection: physicalplan/project.go
> 
> ```go
> func (a aliasProjection) Project(mem memory.Allocator, ar arrow.Record) ([]arrow.Field, []arrow.Array, error) {
> 	fields, arrays, err := a.p.Project(mem, ar)
> 	if err != nil {
> 		return nil, nil, err
> 	}
> 
> 	if len(fields) != len(arrays) {
> 		return nil, nil, fmt.Errorf("invalid projection for alias: fields and arrays must be of equal length")
> 	} else if len(fields) == 0 {
> 		return nil, nil, nil
> 	}
>        return []arrow.Field{{
> 		Name:     a.name,
> 		Type:     fields[0].Type,
> 		Nullable: fields[0].Nullable,
> 		Metadata: fields[0].Metadata,
> 	}}, arrays, nil
> ```
> 
> because plainProjection can return fields and arrays is nil. so we need to make a judgment( if len(fields) == 0) to avoid panic later on.

add test case
```
createtable schema=default
----

createtable schema=default
----

insert cols=(labels.label1, stacktrace, timestamp, value, floatvalue)
value1 stack1 1 2 2.0
value1 stack1 3 4 4.0
value1 stack2 5 6 6.0
----

exec
select value, sum(value2) as aaa
----
value
2
4
6

``` 
this test case can't pass, return
```
        testdata/exec/projection/math_projection:10:
         select value, sum(value2) as aaa
        expected:
        value
        2
        4
        6
        
        found:
        runtime error: index out of range [0] with length 0
```

brancz:
@jicanghaixb I think this is something we should fix, but doesn't have to be part of this PR. For what it's worth I think this should already error at the planning stage, as we can statically verify that this column doesn't exist.

PR #704, opened 2024-01-23
Snapshot index rotation

thorfour:
Added a design doc in the designs directory with more detailed information.

TL;DR this fixes some data loss scenarios with the persistent lsm index files that were present with snapshots.

Previously if a snapshot occurred that mean the only copy of those now truncated writes that were in the WAL were now only found in an index file. If that index file was compacted into/out of and the node shutdown during this period that compaction could cause file corruption. This meant the only copy of the data was lost, and data loss occurred. 

Instead of relying solely on the index files we now have the snapshot hard link the index files into the snapshot directory. This is still much faster than copying the data into the snapshot file, and can be restored on startup by hard linking the snapshotted files back into the index.

This also adds a couple of QoL changes around the LSM where now the LSM configures its own levels from the settings instead of having to call configureLSM before calling NewLSM


db_test.go
```diff
@@ -2280,75 +2280,267 @@ func Test_DB_PersistentDiskCompaction(t *testing.T) {
...
+					op: func(table *Table, dir string) {
+						tx := table.db.highWatermark.Load()
+						require.NoError(t, table.db.snapshotAtTX(context.Background(), tx, table.db.snapshotWriter(tx)))
+					},
+				},
+				{
+					op: func(table *Table, dir string) { // save the L1 files to restore them
+						l1Dir := filepath.Join(table.db.indexDir(), "test", table.active.ulid.String(), "L1")
+						saveDir := filepath.Join(dir, "save")
+						require.NoError(t, os.MkdirAll(saveDir, 0755)) // Create a directory to save the files
```
thorfour, line 2485: I have no idea why this is a linkt error

index/compaction.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/go-kit/log/level"
+	"github.com/parquet-go/parquet-go"
+
+	"github.com/polarsignals/frostdb/dynparquet"
+	"github.com/polarsignals/frostdb/parts"
+)
+
+const (
+	IndexFileExtension     = ".idx"
+	ParquetCompactionTXKey = "compaction_tx"
```
asubiotto, line 22: Do we still need this compaction tx logic for index files? Since we now hard link files in snapshots can we just rely on the snapshot txn to truncate the WAL as we previously did?

index/compaction.go
```diff
@@ -0,0 +1,319 @@
...
+		return nil, err
+	}
+
+	return f, nil
+}
+
+func (f *FileCompaction) MaxSize() int64 { return f.maxSize }
+
+// Snapshot will create a new index file and all future compacitons are written to that index file.
+func (f *FileCompaction) Snapshot(dir string) error {
```
asubiotto, line 62: Can there be some concurrency issues in cases where we snapshot this file while we're writing to it?

index/lsm.go
```diff
@@ -146,6 +233,31 @@ func (l *LSM) LevelSize(t SentinelType) int64 {
...
+func (l *LSM) Snapshot(dir string) error {
+	for !l.compacting.CompareAndSwap(false, true) {
+		// TODO: Maybe don't use a tight loop?
+		runtime.Gosched()
+	}
+	defer l.compacting.Store(false)
+
+	// call snapshot on all the levels.
+	for i, lvl := range l.levels {
+		if lvl != nil {
```
asubiotto, line 246: nit:
```
if lvl == nil {
   continue
}
```
to make this easier to read

snapshot.go
```diff
@@ -458,7 +464,7 @@ func WriteSnapshot(ctx context.Context, _ uint64, db *DB, w io.Writer, offline b
 
 					// If file compaction is enabled we do not need to snapshot the parquet parts.
 					for _, cfg := range db.columnStore.indexConfig {
-						if cfg.Type == CompactionTypeParquetDisk {
+						if cfg.Type == index.CompactionTypeParquetDisk {
 							return ErrSkipPart
```
asubiotto, line 468: I feel like we're treating snapshots and index snapshots as two separate things (i.e. the index `Snapshot` method and then this `WriteSnapshot` that skips parts) which is confusing to me from a first glance. Can we unify these two snapshot paths?

asubiotto (commented):
I don't see the design doc, did you maybe forget to commit it?

index/compaction.go
thorfour: Yes for a couple reasons. First is we maintain compaction tx on parts in general and while this isn't strictly used since read_tx have no function it's outside of the purview of this PR to change that behavior. Second is we still replay the index when there are no snapshots available, and require the compaction tx to correctly replay the index.

If we got rid of index replay and just fell back on the wal we could probably remove them.

index/lsm.go
thorfour: done

index/compaction.go
thorfour: No because we call this from the LSM Snapshot function which ensures that no compactions are running.

snapshot.go
thorfour: I think I would like to refactor this to have the snapshot function just call the `index.Snapshot(...)` and have it correctly snapshot itself. But I think that this PR is already a bit too large as is. So I'm going to defer a change like that.

asubiotto: Sounds good, yes I think delegating the snapshotting to the index sounds like the solution here (i.e. removing part iteration from this file).

index/compaction.go
asubiotto: OK! I don't think we should get rid of index replay but I agree this doesn't belong in this PR.

design/index_files.md
```diff
@@ -0,0 +1,64 @@
...
+|   <ParquetFile>    |
+|   <uint64 size>    |
+|   <ParquetFile>    |
+|   <uint64 size>    |
+|   <ParquetFile>    |
+|   <uint64 size>    |
+|--------------------|
+
+Where Parquet files are appending togehter with a uint64 that represents the number of bytes in the previous ParquetFile.
+Each Parquet file represents a single Part in the LSM index. The Part's transaction id is stored in the Parquet file with a metadata key of 'compaction_tx'
```
asubiotto, line 28: "Each parquet file represents a single part in the LSM index" this is no longer true, right?

asubiotto (approved):
LGTM

design/index_files.md
thorfour: It's still true. A index file has many Parquet files/Parts in it, but each Parquet file is a part.

PR #705, opened 2024-01-24
Add support for sorting *array.Binary columns

gernest:
This commit adds sorting `*array.Binary` and dictionary of `*array.Binary`. I wanted `GenericTable` to automatically sort records before write. This is the only blocker.

### DIFF
--- a/pqarrow/arrowutils/sort.go
+++ b/pqarrow/arrowutils/sort.go
@@ -1,6 +1,8 @@
 package arrowutils
 
 import (
+	"bytes"
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


gernest:
We are getting closer to a smoother `frostdb` `ux`. Using `GenericTable` will  make sure the `arrow.Record` is sorted before being inserted into the  db.

asubiotto:
Nice! Could you add test coverage for this?

gernest:
@asubiotto I addressed your comment PTAL.

pqarrow/arrowutils/sort_test.go
```diff
@@ -206,6 +207,47 @@ func TestSortRecord(t *testing.T) {
 	}
 }
 
+func TestSort_Binary(t *testing.T) {
```
asubiotto, line 210: Wouldn't it be more ergonomic to add this to the existing `TestSortRecord` test?

gernest: @asubiotto 

Not possible for now. With the record builder , we map `string` to `*array.String`. There is no `binary` type in the `schema` proto yet so we can't build `*array.Binary` columns or  dictionary of `*array.Binary`

The reason we are discussing `v1alpha1` and `v1alpha2` is I wanted to introduce `binary` type .

This PR came without tests because it is a supplementary PR.  I didn't mind your test recommendation thats why I didn't bring this up.

Personally I would prefer we omit the test here, so I can make sure we can represent and build `*array.Binary` column. The test coverage for this case will be on that work  by extending `TestSortRecord` test case matrix. By which point I will again have to delete `TestSort_Binary`

I'm fine with whatever option you pick to be happy with this PR .

asubiotto: ok, in that case let's omit.

gernest:
@asubiotto thanks for the review, I addressed your comments PTAL.

asubiotto (approved):

PR #706, opened 2024-01-24
Add parca schema sample

gernest:
This commit adds a `Parca` struct sample defining the current `parca` database schema.

This is for use in tests, as well as giving perspective of production outlook for some features. For instance we mostly focus on `*array.String` or `*array.Binary` dictionary, but we see `*array.Int64` dictionary in this schema.

The plan is to eventually remove `parquet` -> `arrow` conversion during ingestion without breaking backward compatibility, this seems like a good start.

### DIFF
--- a/samples/parca.go
+++ b/samples/parca.go
@@ -1,0 +1,17 @@
+package samples
+
+type Parca struct {
+	Duration       int64             `frostdb:"duration,rle_dict"`
+	Labels         map[string]string `frostdb:"labels,rle_dict,asc(3),null_first"`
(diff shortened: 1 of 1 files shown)


asubiotto:
OOC, how do you define sort order of these columns with the struct?

gernest:
@asubiotto 

`asc(4)` sort in ascending, sort order `4`  or `desc(4)` sort in descending, sort order `4` . If you look closer at the `asc(n)` fields you will notice they align with the order of sorting column in `parca` schema `proto`

PR #707, opened 2024-01-25
Add support for building binary columns with records.Build[T]

gernest:
This commit adds `TYPE_BINARY` for schema `v1alpha1` that represents columns of go type `[]byte`.

`[]byte` columns are mapped to `*array.Binary`. All other options i.e `encoding`, `compression` etc works just like other supported column types.


gernest:
Will do a followup PR to add test matrix for sorting binary columns after this lands.

thorfour (approved):
LGTM

PR #708, opened 2024-01-25
Format markdown table for index_lsm design doc

gernest:
I was reading the rendered rfc and this table was not visible. I had to switch to RAW to read it.

cc @thorfour


thorfour (approved):
Huh weird. I blame vim. Thanks!

PR #711, opened 2024-01-26
Fix index deadlocks

thorfour:
Snapshot's and Compaction could deadlock. This fixes that.

Compactions wait for the watermark to reach the newest transaction in the L0 level.
Snapshots were creating a write tx, but where then calling index.Snapshot which acquires the index.Compaction lock.

This could result in a deadlock if a write came in that was newer than the snapshot tx and the compaction was waiting for the watermark to get to that, but the snapshot was waiting to commit it's tx waiting for the compaction lock.

Instead of acquiring a write tx we need only to get a read tx so we don't block the watermark. We also do not need to log to the WAL for snapshots because we do nothing with that information on replay.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1408,7 +1408,6 @@ func TestDBRecover(t *testing.T) {
 		require.NoError(t, err)
 
 		// Insert 3 txns.
-		var lastWriteTx uint64
 		for i := 0; i < numInserts; i++ {
 			samples := dynparquet.NewTestSamples()
 			for j := range samples {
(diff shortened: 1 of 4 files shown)


asubiotto:
Maybe I'm misreading your comment but the reason to log to the WAL is that *any* `begin` call which increments the tx needs to log some record to the WAL (regardless of whether or not it's used during replay), otherwise subsequent transactions will never be logged since the WAL will wait for that transaction to be logged. If snapshots use read transactions then this is no longer necessary.

PR #713, opened 2024-01-29
hedge against concurrent table creation

thorfour:
We recently hit a duplicate registration panic with table creation. 

I added some other db locks around the only other call to `newTable` that didn't already have one. Feels like there should be a safer way to do this than what we have though.


thorfour:
Nevermind. This didn't fix the duplicate registration

PR #715, opened 2024-01-30
Validate index files

thorfour:
Added a `ValidateIndexDir` function for when we're validating snapshots

Added this to the find valid snapshot function as well as before we restore an index from snapshot which keeps us from panicing when we have a corrupted snapshot.

Also added a compaction file seek to the offset when an error during write occurs. Should make compactions more resilient.

### DIFF
--- a/index/compaction.go
+++ b/index/compaction.go
@@ -316,3 +316,47 @@ func (l *inMemoryLevel) Compact(toCompact []parts.Part, options ...parts.Option)
 	postCompactionSize := int64(b.Len())
 	return []parts.Part{parts.NewParquetPart(0, buf, options...)}, preCompactionSize, postCompactionSize, nil
 }
+
+// ValidateIndexDir will validate each index file in the given directory. It will read the last part in each file and ensure that it can open it.
+func ValidateIndexDir(dir string) error {
+	return filepath.WalkDir(dir, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return fmt.Errorf("failed to walk snapshot index directory: %w", err)
+		}
+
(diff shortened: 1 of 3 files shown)


index/compaction.go
```diff
@@ -316,3 +329,48 @@ func (l *inMemoryLevel) Compact(toCompact []parts.Part, options ...parts.Option)
...
+		offset := info.Size() - 8
+		size := make([]byte, 8)
+		if n, err := idxfile.ReadAt(size, offset); n != 8 {
+			return fmt.Errorf("failed to read index file %s size: %w", path, err)
+		}
+
+		parquetSize := int64(binary.LittleEndian.Uint64(size))
+		offset -= parquetSize
+
+		_, err = parquet.OpenFile(io.NewSectionReader(idxfile, offset, parquetSize), parquetSize)
```
gernest, line 369: I was wondering do we only care that the snapshot is a `parquet` file and not the actual file we actually snapshotted ?

I can be wrong, but when validating backup, the need to know you are working with the actual data you backup is important.

I'm talking about having a checksum of the written data something like `crc32` checksum.

You can completely ignore this comment, it is random question rather than a review.

gernest:
The benefits of using checksums is we can completely separate snapshot validation with recovery process. There is no need to open `parquet` files when validating snapshots.

Things like bit rot or file tampering can be handled by checksums.

index/compaction.go
thorfour: We're less concerned that this is the file we expect and more concerned that it's a valid file. That said I'm going to close this for now because I want to refactor snapshots as they are right now as it's a little convoluted between the table indices and the snapshot file.

PR #717, opened 2024-01-30
Fix deadlock

thorfour:
Fixes a deadlock that can occur if the reading goroutine exits but strands the collect row groups routine.


metalmatze (approved):

asubiotto:
We had this select in a previous version of the code, I remember having to add it. It seems like it got removed in https://github.com/polarsignals/frostdb/pull/711/files#diff-134c48f173480a9772719053c1b3e967756a71c75f84ed4f94f12c71cd91ed30L1195.

Could you please add a regression test?

PR #719, opened 2024-02-01
Fix missing compaction tx

thorfour:
Fixes compaction being able to write files without the `compaction_tx` metadata.

Also no longer treats a missing `compaction_tx` as an error, and simply downgrades the part tx.


metalmatze (approved):
SGTM. More reviews would be great.

asubiotto:
Is there some more integration testing we can do here that would've caught this issue?

Issue #722, opened 2024-02-05
Snapshot refactor


thorfour:
Snapshots are in a bit of a goofy place logically wise. The database walks the list of tables and writes the L0 portion of the index into the snapshot file, and then calls `index.Snapshot(dir)` for the rest of the index. Which just feels logically wrong. It would be nice if a snapshot could be contained entirely by the index API. Where we need only to call `index.Snapshot(...)` on each table and it would do the right thing :tm: 

It would be nice to change snapshots to only include tx <= snapshot_tx but it's going to require pausing all table compactions which may or may not be worth it. Right now snapshots can include > TX but during replay the index won't insert duplicate or lower writes so it's still technically correct.

PR #730, opened 2024-02-07
snapshot: refactor to simplify

thorfour:
Part one of #722 

This moves more of the snapshot logic into the index itself instead of in the snapshot layer. 

Previously the snapshot would know to walk `L0` parts and write them to the file, and then depending on how the index was configured it would either write the `L1+` parts to the file or it would skip them and call `index.Snapshot` at the end.

This moves the handling of levels all into the index where it belongs. Now Snapshot provides a writer function for parts, and a directory for the snapshot and just calls `index.Snapshot` for each table. 

Next part will be to move the snapshot into it's own package.

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -414,31 +414,15 @@ func WriteSnapshot(ctx context.Context, tx uint64, db *DB, w io.Writer, offline 
 				if err := ctx.Err(); err != nil {
 					return err
 				}
-				schema := t.schema
 
 				if record := p.Record(); record != nil {
 					partMeta.Encoding = snapshotpb.Part_ENCODING_ARROW
-					recordWriter := ipc.NewWriter(
-						w,
-						ipc.WithSchema(record.Schema()),
-					)
-					defer recordWriter.Close()
-					if err := recordWriter.Write(record); err != nil {
-						return err
-					}
 				} else {
 					partMeta.Encoding = snapshotpb.Part_ENCODING_PARQUET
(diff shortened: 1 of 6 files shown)


snapshot.go
```diff
@@ -406,43 +404,30 @@ func WriteSnapshot(ctx context.Context, tx uint64, db *DB, w io.Writer, offline
...
 					CompactionLevel: uint64(p.CompactionLevel()),
 				}
-				err := func() error {
-					if err := ctx.Err(); err != nil {
+				if err := ctx.Err(); err != nil {
+					return err
+				}
+				schema := t.schema
+
+				if record := p.Record(); record != nil {
```
asubiotto, line 419: One other thing I would do to clean this up is maybe adding a `Write(w io.Writer)` to the part interface to avoid having the snapshot code figure this out.

asubiotto (commented):
The more we move out of the snapshot file, the more I wonder what exactly the "purpose" will be of an independent snapshot package. What will it be?

thorfour:
> The more we move out of the snapshot file, the more I wonder what exactly the "purpose" will be of an independent snapshot package. What will it be?

It still contains the file format and all the logic of reading/writing that file. I don't think the index should know how a snapshot file is written or read.

snapshot.go
thorfour: oh I like that idea!

thorfour: done

asubiotto (approved):

PR #731, opened 2024-02-08
query/physicalplan: For special substring regexp filter run substr matching directly

metalmatze:
Once a regexp filter is run it checks for .% at the beginning and end.
If it is true, then using bytes.Contains or strings.Contains is way more performant than running regexp.

Here's the result of running BenchmarkBinaryScalarRegexp:
```
name                      old time/op  new time/op  delta
BinaryScalarRegexp/=~-24   122ms ± 3%     1ms ±26%  -99.34%  (p=0.000 n=10+9)
BinaryScalarRegexp/!~-24   123ms ± 2%     1ms ±14%  -99.31%  (p=0.000 n=10+9)
```

### DIFF
--- a/query/physicalplan/filter.go
+++ b/query/physicalplan/filter.go
@@ -1,10 +1,12 @@
 package physicalplan
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/RoaringBitmap/roaring"
 	"github.com/apache/arrow/go/v14/arrow"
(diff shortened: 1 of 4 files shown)


query/physicalplan/filter.go
```diff
@@ -91,16 +93,38 @@ func binaryBooleanExpr(expr *logicalplan.BinaryExpr) (BooleanExpression, error)
 
 		switch expr.Op {
 		case logicalplan.OpRegexMatch:
-			regexp, err := regexp.Compile(string(rightScalar.(*scalar.String).Data()))
+			rightString := string(rightScalar.(*scalar.String).Data())
+
+			// This is a simple optimization to avoid using regex for simple substring matches.
+			if strings.HasPrefix(rightString, ".*") && strings.HasSuffix(rightString, ".*") {
```
brancz, line 99: This is not correct, there can be other wildcards in the regex and then this doesn't work.

Issue #721, opened 2024-02-05
Proper transaction support


thorfour:
Transactions were originally added to FrostDB because we used to split writes up and sort each row of a write into different granules. The scan layer excludes transactions that have a higher tx number than the read is being performed with to avoid reading a partial write.

However today we do not split up writes so reads excluding `tx > read_tx` doesn't make as much as sense because it's still a consistent read we're just now artificially limiting the amount of data we read. 

I propose that we add proper database transactions to FrostDB such that we can atomically write to multiple tables at once. Potentially in the future support other database transactions as well such as queries, table creation/deletion etc.

API Proposal

```go
type Table interface {
    InsertRecord(context.Context, arrow.Record) error
}

type Transaction interface {
    GetTable(string) (Table, error)
    Commit()
    Abort()
}
  
type DB interface {
    ...
    Begin() Transaction
}
```

Which would allow for something like

```go

tx := db.Begin()
defer tx.Commit()
if err := tx.GetTable("table1").InsertRecord(ctx, r1); err != nil {
   tx.Abort()
   return
}
if err := tx.GetTable("table2").InsertRecord(ctx, r2); err != nil {
   tx.Abort()
   return
}
```

Multi-table transactions could be implemented quite easily by getting a write tx which already blocks reads from reading it as well as compactions from compacting it until it has been committed. In the event of an abort the transaction can walk back the writes and remove them from the L0 level of each table that they were written to before marking the transaction as complete. Thus still providing a consistent view.

asubiotto:
I think multi-table transactions would only be beneficial in cases where we allow joining multiple tables. I'm very wary of adding joins to the DB (you introduce a world of complexity), so I'm simply not sure that explicit transactions would give us any benefits, but I might be missing something.

thorfour:
> I think multi-table transactions would only be beneficial in cases where we allow joining multiple tables. I'm very wary of adding joins to the DB (you introduce a world of complexity), so I'm simply not sure that explicit transactions would give us any benefits, but I might be missing something.

What about the writes that are causally related? Like we have in production today where we write two tables and it has resulted in missing samples if one of the writes completes but the other does not?

asubiotto:
True. That's what I was missing then.

thorfour:
One thing required for this change to to change snapshots to respect the tx. Right now they snapshot whatever is in the index.

PR #733, opened 2024-02-08
[DO NOT MERGE] Multi-table transaction support

thorfour:
Took a stab at #721 

There's a few rough edged I need to address around locking on aborts, and if we need a write block during aborts. But it's a start for a reference implementation.

### DIFF
--- a/db.go
+++ b/db.go
@@ -749,7 +749,7 @@ func (db *DB) recover(ctx context.Context, wal WAL) error {
 						filepath.Join(table.db.indexDir(), table.name, table.ActiveBlock().ulid.String()), // Any index files are found at <db.indexDir>/<table.name>/<block.id>
 						table.schema,
 						table.IndexConfig(),
-						db.Wait,
+						db.HighWatermark,
 						index.LSMWithMetrics(table.metrics.indexMetrics),
 						index.LSMWithLogger(table.logger),
 					)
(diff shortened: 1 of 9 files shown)


asubiotto:
How are you avoiding reading incomplete transactions? I'm also curious as to what happens in case of a crash before a transaction fully commits.

In case it's helpful, maybe what I remember from cockroach transaction coordination could be useful:

Each write that forms a part of a single transaction will be added to the LSM in a "pending" state with a pointer to a transaction intent (i.e. intent to commit) in that a central transaction coordinator manages. On reads, any part that points to one of these intents is only returned if the intent is in the "committed" state (that is an atomic flip once the `commit()` is called). This allows for all writes to have a linearization point where they are made available for reads. On transaction abort, the intent will be set to the "aborted" state and the participating writes will never have been returned for reads. Intent cleanup would involve first removing the pointers on the parts to the intent (implying a valid commit) in case of commit and simply removing the parts (or even just leaving them there since any reads e.g. for compaction should skip over uncommitted writes) in case of abort.

thorfour:
Read isolation is already taken care of with the watermark. The watermark can never increase past the incomplete transaction and therefore none of the writes can be included in the read. 

Crashing before commit is something that I still need to handle with the WAL (I believe there's  a TODO in there still). But I don't think our WAL implementation handles writes with the same tx so we'll need to have a prepare/commit message in the wal to indicate partial tx during replay and just drop them if we never got a commit message.

PR #734, opened 2024-02-09
query: Add Contains operator (rebased)

metalmatze:
Adding a `|=` operator to FrostDB.
It's basically `LIKE '%$foo%'` in SQL. It'll return true whenever the string contains the substring.

This is a rebased version of #634.
I'll be following up with the logicalplan optimizer for regexp substring matchers.


logictest/testdata/exec/filter/filter
```diff
@@ -156,3 +156,14 @@ select labels, timestamp, value where doesntexist >= 4
 exec
 select labels, timestamp, value where doesntexist <= 4
 ----
+
+exec
+select stacktrace, value where stacktrace LIKE 'ack'
```
metalmatze, line 161: FYI: These filter for dictionaries containing a substring.

logictest/testdata/exec/filter/filter_contains
```diff
@@ -0,0 +1,19 @@
...
+value3  value2  null    value4  3   baz
+----
+
+exec
+select labels, timestamp, value where timestamp = 2
+----
+value2  value2  value3  null    2       bar
+
+exec
+select labels, timestamp, value where value LIKE 'a'
```
metalmatze, line 16: FYI: This filters a array.Binary for substrings. No dictionary involved.

query/physicalplan/binaryscalarexpr.go
```diff
@@ -217,3 +229,69 @@ func DictionaryArrayScalarEqual(left *array.Dictionary, right scalar.Scalar) (*B
...
+			continue
+		}
+
+		switch dict := left.Dictionary().(type) {
+		case *array.Binary:
+			if bytes.Contains(dict.Value(left.GetValueIndex(i)), data) {
+				res.Add(uint32(i))
+			}
+		case *array.String:
+			// TODO: Add unsafe type cast if necessary.
```
brancz, line 289: Can we write a benchmark for this? I suspect this is already necessary right off the bat.

brancz (approved):
Nice!

thorfour (approved):
Agreed with adding string support right away, otherwise lgtm

query/logicalplan/expr.go
```diff
@@ -60,6 +61,8 @@ func (o Op) String() string {
 		return "*"
 	case OpDiv:
 		return "/"
+	case OpContains:
+		return "|="
```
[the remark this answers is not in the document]
metalmatze: Sure. I'm happy to do that, too. The `|=` seems more on par with the rest of the operators.

metalmatze:
Good call on benchmarking and making sure to add back the `unsafeStringToBytes`.
```
name                                old time/op    new time/op    delta
BinaryScalarContains/binary-10         442µs ± 0%     433µs ± 0%   -2.13%  (p=0.000 n=10+9)
BinaryScalarContains/string-10         433µs ± 0%     442µs ± 0%   +2.26%  (p=0.000 n=10+10)
BinaryScalarContains/dictionary-10    1.25ms ± 2%    0.44ms ± 1%  -64.53%  (p=0.000 n=9+10)

name                                old alloc/op   new alloc/op   delta
BinaryScalarContains/binary-10        25.1kB ± 0%    25.1kB ± 0%     ~     (all equal)
BinaryScalarContains/string-10        25.1kB ± 0%    25.1kB ± 0%     ~     (all equal)
BinaryScalarContains/dictionary-10    5.27MB ± 0%    0.03MB ± 0%  -99.52%  (p=0.000 n=9+10)

name                                old allocs/op  new allocs/op  delta
BinaryScalarContains/binary-10          18.0 ± 0%      18.0 ± 0%     ~     (all equal)
BinaryScalarContains/string-10          18.0 ± 0%      18.0 ± 0%     ~     (all equal)
BinaryScalarContains/dictionary-10     10.0k ± 0%      0.0k ± 0%  -99.82%  (p=0.000 n=10+10)
```

query/logicalplan/expr.go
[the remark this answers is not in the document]
brancz: I think `|=` is so uncommon that I didn’t realize first time reading it what it was, but again, no strong feelings, happy to merge as is.

PR #737, opened 2024-02-13
Fix logicalplan validations

brancz:


metalmatze (approved):

db_test.go
```diff
@@ -487,6 +491,9 @@ func Test_DB_ColdStart(t *testing.T) {
 			db, err = c.DB(context.Background(), sanitize(t.Name()))
 			require.NoError(t, err)
 
+			_, err = db.Table(sanitize(t.Name()), config)
```
thorfour, line 494: The point of the cold start test was to ensure we could read data after restarting without having to reconfigure the tables schema. This test should just be deleted as we no longer hold to that.

thorfour (approved):
Please remove cold start test

db_test.go
brancz: Makes sense!

PR #739, opened 2024-02-14
Set OptInt32Builder data to nil

metalmatze:
The reason we want to do this is so that when pooling the builder as part of the `MultiColSorter` we don't reuse the builder's data for a concurrent existing array.

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -342,7 +342,8 @@ func BenchmarkQuery(b *testing.B) {
 	// BenchmarkFilter benchmarks performing a simple filter. This is useful to get an idea of our scan speed (minus the
 	// execution engine).
 	b.Run("Filter", func(b *testing.B) {
-		table, err := db.Table(tableName, nil)
+		schema := dynparquet.SampleDefinition()
+		table, err := db.Table(tableName, NewTableConfig(schema))
 		require.NoError(b, err)
 		size := table.ActiveBlock().Size()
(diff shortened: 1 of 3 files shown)


asubiotto:
Can you add a regression test? You don't even need concurrency. I would just reuse a builder and check that a previously returned array is intact.

I'm a little concerned by how this will affect scan performance though since we will introduce more allocations. Could you run `BenchmarkQuery` on a prod db?

asubiotto:
I would also change this for all builders. I think this is a "contract-level" change. i.e. the caller of `NewArray` retains ownership of this array and the builder loses it, so any builder reuse should not affect existing arrays.

metalmatze:
Here are the results for the `BenchmarkQuery`:
```
name             old time/op  new time/op  delta
Query/Types-24    754µs ± 1%   773µs ± 2%  +2.53%  (p=0.032 n=5+5)
Query/Labels-24   292µs ± 1%   263µs ±10%  -9.79%  (p=0.016 n=4+5)
Query/Values-24   592µs ±20%   628µs ±16%    ~     (p=0.548 n=5+5)
Query/Merge-24    560ms ± 7%   568ms ±10%    ~     (p=1.000 n=5+5)
Query/Range-24   85.3ms ± 5%  87.0ms ±11%    ~     (p=0.310 n=5+5)
```
It doesn't seem too bad, is it?

asubiotto (approved):

PR #742, opened 2024-02-15
pqarrow: Ensure schema sorting

brancz:


thorfour (approved):

pqarrow/arrow.go
```diff
@@ -59,26 +59,36 @@ func ParquetSchemaToArrowSchema(ctx context.Context, schema *parquet.Schema, s *
...
+				})
 			}
-			fields = append(fields, arrow.Field{
-				Name:     distinctExpr.Name(),
-				Type:     dataType,
-				Nullable: true, // TODO: This should be determined by the expression and underlying column(s).
-			})
 		}
+
+		// Need to sort as the distinct columns are just appended, but we need
```
asubiotto, line 85: It's been a while since I looked at this code, but is this the right place to do it? Do `ParquetSchemaToArrowSchema` consumers expect the schema to be sorted? My hunch is that the actual issue is that the builder assumes field sorting when it shouldn't and so it should be the one to sort.

If a regression test is not too much trouble I think it would be good to add.

PR #745, opened 2024-02-15
Snapshot: Always acquire an ActiveWriteBlock

thorfour:
When we were closing the database we may take a snapshot on close. If we do, that snapshot was taking on "offline" snapshot which doesn't acquire the active writeblock.

Because it doesn't acquire an active write block it menat it could race with an active block rotation, which could close the index before the snapshot occurs resulting in a bad snapshot.

This changes the snapshot behavior to always be "online" and acquire the active write block.


asubiotto (approved):
Great find.

asubiotto:
Actually, isn't it enough to just move snapshotting before closing the tables?

thorfour:
No, because moving it to before would still cause it to race with the block rotation

PR #746, opened 2024-02-19
Add unique and "and" aggregations, and if and isNull projections

brancz:


thorfour:
May be worth adding some logic tests with these additions?

query/physicalplan/project.go
```diff
@@ -485,6 +485,132 @@ func (p convertProjection) Project(mem memory.Allocator, ar arrow.Record) ([]arr
...
+		return nil, nil, fmt.Errorf("invalid projection for if: then and else columns must be of the same type")
+	}
+
+	var res arrow.Array
+	switch thenCols[0].(type) {
+	case *array.Int64:
+		res = conditionalAddInt64(mem, condArr, thenCol.(*array.Int64), elseCol.(*array.Int64))
+	default:
+		return nil, nil, fmt.Errorf("unsupported if expression type: %s %T", thenCols[0].DataType(), thenCols[0])
+	}
```
jicanghaixb, line 587: because thenCol.Datatype equal elseCol.Datatype, we can support any dataType,not only int64.
example:
        builder := array.NewBuilder(mem, thenColFields[0].Type)
        defer builder.Release()

jicanghaixb: I think IfProjection can generate any type of new column based on conditional judgment, rather than just int64 column, so we can use thenCol or elseCol types to get new column dataType

brancz: The result must be the same on either side, which is ensured on line 577

jicanghaixb: yes, the result must be the same on either side, but  thenCol and elseCol must not be set int64 type.
This can adapt to more new column generation needs

Here is my previously ugly implementation code
```suggestion

type ifProject struct {
	filterColumn columnProjection
	trueColumn   columnProjection
	falseColumn  columnProjection
	expr         *logicalplan.IfElseExpr
}

func (p ifProject) Name() string {
	return p.expr.Name()
}

func (p ifProject) String() string {
	return p.Name()
}

func (p ifProject) Project(mem memory.Allocator, ar arrow.Record) ([]arrow.Field, []arrow.Array, error) {
	var (
		resultName   = p.Name()
		fields       = make([]arrow.Field, 0, 1)
		columnsArray = make([]arrow.Array, 0, 1)
	)

	filterFields, filterArrays, err := p.filterColumn.Project(mem, ar)
	if err != nil {
		return nil, nil, fmt.Errorf("filter projection for column failed: %v", err)
	}
	if len(filterFields) != 1 || len(filterArrays) != 1 {
		return nil, nil, fmt.Errorf("filter expression projection expected one field and one array for each side, got %d fields and %d arrays on left", len(filterFields), len(filterArrays))
	}
	defer filterArrays[0].Release()

	trueFields, trueArrays, err := p.trueColumn.Project(mem, ar)
	if err != nil {
		return nil, nil, fmt.Errorf("true projection for column failed: %v", err)
	}
	if len(trueFields) != 1 || len(trueArrays) != 1 {
		return nil, nil, fmt.Errorf("true expression projection expected one field and one array for each side, got %d fields and %d arrays on left", len(trueFields), len(trueArrays))
	}
	defer trueArrays[0].Release()

	falseFields, falseArrays, err := p.falseColumn.Project(mem, ar)
	if err != nil {
		return nil, nil, fmt.Errorf("false projection for column failed: %v", err)
	}
	if len(falseFields) != 1 || len(falseArrays) != 1 {
		return nil, nil, fmt.Errorf("true expression projection expected one field and one array for each side, got %d fields and %d arrays on left", len(falseFields), len(falseArrays))
	}
	defer falseArrays[0].Release()

	if trueFields[0].Type.ID() != falseFields[0].Type.ID() {
		return nil, nil, fmt.Errorf("trueColumn type: %s not equal falseColumn type: %s", trueFields[0].Type, falseFields[0].Type)
	}
	fields = append(fields, arrow.Field{
		Name: resultName,
		Type: trueFields[0].Type,
	})
	builder := array.NewBuilder(mem, trueFields[0].Type)
	defer builder.Release()
	for i := 0; i < int(ar.NumRows()); i++ {
		if filterArrays[0].ValueStr(i) == "true" {
			if err := builder.AppendValueFromString(trueArrays[0].ValueStr(i)); err != nil {
				return nil, nil, err
			}
		} else {
			if err := builder.AppendValueFromString(falseArrays[0].ValueStr(i)); err != nil {
				return nil, nil, err
			}
		}
	}
	columnsArray = append(columnsArray, builder.NewArray())
	return fields, columnsArray, nil
}
```

PR #749, opened 2024-02-20
help add the logic of creating arrow builder for the type uint64 in Go

elliotchenzichang:
Hi, long time no see. Wish you all the best. 
Recently I learned the base knowledge of Apache Arrow and read the code for creating Arrow Builder in your project. I noticed that you are only supporting some primitive types in Go, and these types are int64, float64, string, and bool. 
So I would like to help implement builders for other primitive types in Go, such as uint64, uint8, etc.
I also noticed that we should define the values of enums in protobuf, and the type uint64 has already been defined there. So this PR is just a case for what I want to do, and it implements the arrow builder for the type uint64. 
Would like to wait for your opinions or concerns, thanks!

### DIFF
--- a/internal/records/record_builder.go
+++ b/internal/records/record_builder.go
@@ -584,6 +584,17 @@ func newFieldBuild(dt arrow.DataType, mem memory.Allocator, name string, nullabl
 			}
 			return e.Append(v.Int())
 		}
+	case *array.Uint64DictionaryBuilder:
+		f.buildFunc = func(v reflect.Value) error {
+			if nullable {
+				if v.IsNil() {
+					e.AppendNull()
+					return nil
+				}
+				v = v.Elem()
+			}
+			return e.Append(v.Uint())
+		}
 	case *array.Float64Builder:
 		f.buildFunc = func(v reflect.Value) error {
 			if nullable {
@@ -653,6 +664,19 @@ func newFieldBuild(dt arrow.DataType, mem memory.Allocator, name string, nullabl
 				e.Append(true)
 				build.Reserve(v.Len())
 				return applyInt(v, func(i int64) error {
+					build.Append(i)
+					return nil
+				})
+			}
+		case *array.Uint64Builder:
+			f.buildFunc = func(v reflect.Value) error {
+				if v.IsNil() {
+					e.AppendNull()
+					return nil
+				}
+				e.Append(true)
+				build.Reserve(v.Len())
+				return applyUInt(v, func(i uint64) error {
 					build.Append(i)
 					return nil
 				})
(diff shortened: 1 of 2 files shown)


internal/records/record_builder.go
```diff
@@ -558,6 +561,18 @@ func newFieldBuild(dt arrow.DataType, mem memory.Allocator, name string, nullabl
 			e.Append(v.Int())
 			return nil
 		}
+	case *array.Uint64Builder:
```
gernest, line 564: If we support `*array.Uint64Builder` we also need to support `*array.Uint64DictionaryBuilder` . And I suggest the switch case for `*array.Uint64Builder` to be after `*array.Int64DictionaryBuilder`.

Also you need to add test case for `uint64` in  `record_builder_test.go` to make sure everything works as intended.

gernest (changes requested):
Looks promising to me. We already have `Uint64` in schema so It makes sense to support building it.

II left a comment. Please let me know if you need any help regarding record builder.

internal/records/record_builder.go
elliotchenzichang: Thanks for your suggestion, raised some commit for them. Please review it again when you have time.

elliotchenzichang: Besides, if I want to implement the arrow builders for other primitive types, what should I do after I add the enum values in the proto files?

gernest (approved):
Awesome work, LGTM!

> Besides, if I want to implement the arrow builders for other primitive types, what should I do after I add the enum values in the proto files?

Unfortunate, I'm not in a position to help you on this.

Normally, a team member will need to comment if those primitive types are actually needed.

Assumming they agree to to it then you will need

- update the schema proto with the type https://github.com/polarsignals/frostdb/blob/83dac77ff4287cba4cd9846b8c6674e9f2a71039/proto/frostdb/schema/v1alpha1/schema.proto#L34-L49
- Then come back the builder to implement them.

Note that, you should only work with schema `v1alpha1` they are deprecating `v1alpha2`

Querying other data types might be challenging, but for all primitive types `compute` package does the heavy lifting so, I think it is okay to add them without worrying about querying, as handling the edge cases can be easily added when needed.. 


Just ping me if you have any questions regarding the record builder. 

Thanks for your contribution and good luck on whatever your are up to.

elliotchenzichang:
Thanks a lot. @gernest

elliotchenzichang:
Hi @gernest could we merge this PR first if no concerns？ Thanks！

thorfour:
Looks good to me @elliotchenzichang thanks for the contribution!

PR #751, opened 2024-02-21
query: add global aggregation function pushdown optimization for Max

asubiotto:
When reading Particulates from blocks (either in-memory or object storage), we can drastically reduce the runtime of global aggregations with no filters. This commit implements the optimization pushdown as a filter.

In case of Max (implemented in this commit) for example, the TrueNegativeFilter memoizes the maximum value seen and filters out any Particulates that will definitely not contribute to the aggregation function (i.e. whose max value is less than the max value seen so far). This drastically reduces global Max queries. For example, on 1.5GB of data, a global Max on an unsorted int64 column went from 900ms to 50ms.

### DIFF
--- a/query/logicalplan/optimize.go
+++ b/query/logicalplan/optimize.go
@@ -15,6 +15,7 @@ func DefaultOptimizers() []Optimizer {
 		},
 		&FilterPushDown{},
 		&DistinctPushDown{},
+		&AggFuncPushDown{},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


query/logicalplan/optimize.go
```diff
@@ -154,3 +155,39 @@ func (p *DistinctPushDown) optimize(plan *LogicalPlan, distinctColumns []Expr) {
...
+}
+
+func (p *AggFuncPushDown) optimize(plan *LogicalPlan, filterExpr Expr) {
+	switch {
+	case plan.TableScan != nil:
+		if filterExpr != nil {
+			plan.TableScan.Filter = filterExpr
+		}
+	case plan.Aggregation != nil:
+		if plan.Aggregation.GroupExprs == nil && len(plan.Aggregation.AggExprs) == 1 {
```
brancz, line 180: nit: just to be more defensive here, maybe check the len of group exprs instead

brancz (approved):
Very cool optimization!

brancz (approved):

thorfour (approved):

PR #754, opened 2024-02-26
Update README.md

raphaelauv:
InfluxDB IOx is now part of InfluxDB

https://www.influxdata.com/blog/the-plan-for-influxdb-3-0-open-source/#heading3

also duckdb is an embedded  db


thorfour (approved):
Thanks for the fixes!

PR #761, opened 2024-03-04
Fix: double snapshot data loss

### LINKED ISSUE #753: snapshot: two snapshots at the same txn causes data loss
If a DB receives no writes between two snapshots, the second snapshot will error out, causing the first snapshot to be removed from disk and resulting in data loss:
```
level=error db=<...> name=storage ts=2024-02-22T08:24:04.677341856Z caller=snapshot.go:140 msg="failed to snapshot database" err="failed to write snapshot for tx 47318: failed to snapshot table <...> index: link /data/database/<...>/index/table/01HQ64SWCRHN5AR6RS01SXYN16/L1/00000000000000000000.idx /data/databases/<...>/snapshots/00000000000000047318/index/table/01HQ64SWCRHN5AR6RS01SXYN16/L1/00000000000000000000.idx: file exists"
```

The above error causes the following log line to be printed on recovery:
```
level=info db=<...> name=storage ts=2024-02-22T08:24:30.532034244Z caller=db.go:700 msg="failed to load latest snapshot" db=<...> err="no valid snapshots found"
```
The culprit seems to be this line:
https://github.com/polarsignals/frostdb/blob/d9f86325ab691f52e33d19b86ff9d78b6ea313c3/snapshot.go#L192

Couple of ways to fix this. Maybe the easiest is to us `reclaimDiskSpace` instead of `RemoveAll` which takes into account snapshot validity on disk. Another is to remove a snapshot with the same txn on disk before performing the new snapshot, or checking whether a valid snapshot already exists on disk (since if it is invalid I think we do want to perform it), but this might be more expensive.

thorfour:
Previously if we took a snapshot, stopped writing and shutdown the database we would take another snapshot. This snapshot would fail which would cause it to remove the valid snapshot directory.

This changes the behavior of snapshots to first check if a snapshot already exists. If it does validate it and if the snapshot is valid abort taking a new one. If the snapshot is invalid then remove it and perform the snapshot.

Closes #753

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -2853,8 +2853,6 @@ func Test_DB_SnapshotDuplicate(t *testing.T) {
 	require.NoError(t, err)
 	db, err = c.DB(context.Background(), "test")
 	require.NoError(t, err)
-	table, err = db.Table("test", config)
-	require.NoError(t, err)
 
 	validateRows := func(expected int64) {
 		pool := memory.NewCheckedAllocator(memory.DefaultAllocator)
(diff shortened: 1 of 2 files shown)


brancz (approved):
Nice!

PR #763, opened 2024-03-07
add script into makefile for easier upgrading dependencies

elliotchenzichang:
Hi, me again. 
For easier upgrading dependencies, I add a script into makefile. We can simply trigger it to upgrade all dependencies. This will bring more convenience to our development


thorfour:
I think instead of a script we want to add renovate https://github.com/renovatebot/renovate instead and have that automatically update our dependencies for us.

metalmatze:
It's probably possible to copy most of the config from Parca: https://github.com/parca-dev/parca/blob/main/renovate.json
That one, in turn, depends on the Parca org-wide config, which we can also add to polarsignals.
https://github.com/parca-dev/.github/blob/main/.github/renovate.json

Let us know if you have further questions!

PR #766, opened 2024-03-12
pqarrow/arrowutils: Use SortingColumns in MergeRecords

metalmatze:
This allows to merge records in arbitrary orders and columns.


asubiotto:
Weird, I thought we already did this?

PR #771, opened 2024-03-22
fix: uint64 persistence

thorfour:
We preiviously were not detecting uint64 arrow arrays and weren't casting them correctly to an int64 array for Parquet. This adds the switch statement to correctly perform this conversion.

We were also assuming that ONLY dynamic columns needed to have a definition level of 1, but we have other columns that also support nullable values. This removes the dynamic column check and correctly checks the parquet optionality instead.


metalmatze (approved):
🥳

PR #784, opened 2024-03-26
Add FrostDB gRPC Query service

metalmatze:
This allows applications (TypeScript, Go, or something else) to send arbitrary logical plans to an application implementing the FrostDBService Query endpoint.


proto/frostdb/storage/v1alpha1/storage.proto
```diff
@@ -0,0 +1,299 @@
...
+  // Query executes the query plan contained in the request and returns back
+  // the results as arrow records in IPC format.
+  rpc Query(QueryRequest) returns (QueryResponse) {}
+}
+
+// QueryRequest is the message sent to the Query gRPC endpoint.
+message QueryRequest {
+  // PlanRoot is the root of the query plan (i.e. a scan node).
+  PlanNode plan_root = 1;
+}
```
metalmatze, line 17: It's pretty hard to debug individual queries with three different ones already. 
What do you think about adding this? 

```suggestion
message QueryRequest {
  // PlanRoot is the root of the query plan (i.e. a scan node).
  PlanNode plan_root = 1;
  // optional description of the query to make debugging easier
  string description = 2;
}
```

proto/frostdb/storage/v1alpha1/storage.proto
```diff
@@ -0,0 +1,299 @@
...
+
+// QueryResponse is the message received from the Query gRPC endpoint.
+message QueryResponse {
+  // The resulting arrow record as bytes.
+  bytes record = 1;
+}
```
metalmatze, line 23: Similar to the above, I added more information about individual queries. One of which is always running `explain` on the query. This string is logged. 

I'm wondering what you would think about also (optionally) returning the explainer string to the caller.
```suggestion
// QueryResponse is the message received from the Query gRPC endpoint.
message QueryResponse {
  // The resulting arrow record as bytes.
  bytes record = 1;
  // explain string that shows the underlying executed physical plan
  string explain = 2;
}
```

proto/frostdb/storage/v1alpha1/storage.proto
nicolastakashi: I like the idea, just wondering how we will populate this?
This should be dynamically right?

proto/frostdb/storage/v1alpha1/storage.proto
nicolastakashi: Like the idea, especially in the future we can have a query planner 🔝

proto/frostdb/storage/v1alpha1/storage.proto
```diff
@@ -0,0 +1,301 @@
+syntax = "proto3";
+
+package frostdb.storage.v1alpha1;
+
+// FrostDBService is a service that allows clients to execute queries as logicalplan
+// against an instance of FrostDB.
+service FrostDBService {
+  // Query executes the query plan contained in the request and returns back
+  // the results as arrow records in IPC format.
+  rpc Query(QueryRequest) returns (QueryResponse) {}
```
brancz, line 10: this should return a stream, no?

query/exprpb/proto.go
```diff
@@ -0,0 +1,591 @@
...
+func protoAggFuncToLogicalAggFunc(f storagepb.AggregationFunction_Type) (logicalplan.AggFunc, error) {
+	switch f {
+	case storagepb.AggregationFunction_TYPE_SUM:
+		return logicalplan.AggFuncSum, nil
+	case storagepb.AggregationFunction_TYPE_MIN:
+		return logicalplan.AggFuncMin, nil
+	case storagepb.AggregationFunction_TYPE_MAX:
+		return logicalplan.AggFuncMax, nil
+	case storagepb.AggregationFunction_TYPE_COUNT:
+		return logicalplan.AggFuncCount, nil
```
brancz, line 199: the internal version of this has new options here, can you sync this? the proto will need a small update as well

PR #801, opened 2024-03-27
query/expr: Infer type from arrow.Scalar for parquet Value

metalmatze:
When comparing we want to use arrow's scalar.Scalar as the type supports a lot more underlying types, like unsigned integers.

Therefore, we want to convert parquet.Values to arrow scalar.Scalar and then compare. That way we can ensure that the unsigned integers, for example, are properly treated and don't overflow.

### DIFF
--- a/query/expr/binaryscalarexpr.go
+++ b/query/expr/binaryscalarexpr.go
@@ -1,11 +1,16 @@
 package expr
 
 import (
+	"bytes"
+	"cmp"
 	"errors"
 	"fmt"
 
+	"github.com/apache/arrow/go/v15/arrow"
+	"github.com/apache/arrow/go/v15/arrow/scalar"
 	"github.com/parquet-go/parquet-go"
 
+	"github.com/polarsignals/frostdb/pqarrow"
 	"github.com/polarsignals/frostdb/query/logicalplan"
 )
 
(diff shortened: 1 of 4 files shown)


metalmatze:
Some type conversions are missing, and we probably want to run benchmarks before merging this. 
For now, I'm curious about your thoughts.

query/expr/binaryscalarexpr.go
```diff
@@ -80,15 +84,30 @@ var ErrUnsupportedBinaryOperation = errors.New("unsupported binary operation")
...
 	}
 	numNulls := NullCount(leftColumnIndex)
 	fullOfNulls := numNulls == left.NumValues()
+	leftColumnMax := Max(leftColumnIndex)
+	leftColumnMin := Min(leftColumnIndex)
+
+	// Because parquet doesn't support some types that arrow does, we need to
+	// convert the parquet min/max values to arrow scalars to be able to compare,
+	// using the right scalar as the type reference.
+	leftColumnMaxScalar, err := pqarrow.ParquetValueToArrowScalar(leftColumnMax, right.DataType())
```
metalmatze, line 100: This is the most important change of this PR.
Given that we have the right side's DataType, we know exactly what type we want to compare it to. Using that information, we can be more precise when converting parquet to arrow.

thorfour: Hmm it feels bad to need to use the index to determine the scalar especially since the index is not a thing that is requried to exist.

metalmatze: I think it's the other way around. We use the scalar passed into the binary scalar operation to determine what type the max and min types from the index should have.

PR #805, opened 2024-03-28
internal/records: Create generic reader

metalmatze:
Similar to the generic builder I've added a reader.
It can read multiple records and returns the passed struct when calling Value(i).

This makes it super easy to read the records back to structs and compare against some expected values.

```go
type struct3 struct {
	Field1 int64 `frostdb:",asc(0)"`
	Field2 int64 `frostdb:",asc(1)"`
	Field3 int64 `frostdb:",asc(1)"`
}
expected := []struct3{
	// record1
	{Field1: 1, Field2: 2, Field3: 0},
	{Field1: 1, Field2: 3, Field3: 0},
	// record2
	{Field1: 1, Field2: 0, Field3: 2},
	{Field1: 1, Field2: 0, Field3: 3},
	// record3
	{Field1: 1, Field2: 1, Field3: 1},
	{Field1: 2, Field2: 2, Field3: 2},
}

reader := records.NewReader[struct3](recs...)
rows := reader.NumRows()
require.Equal(t, int64(len(expected)), rows)

actual := make([]struct3, rows)
for i := 0; i < int(rows); i++ {
	actual[i] = reader.Value(i)
}
require.Equal(t, expected, actual)
```

I didn't add tests as this reader is going to be used in tests a lot.

### DIFF
--- a/internal/records/record_reader.go
+++ b/internal/records/record_reader.go
@@ -1,0 +1,188 @@
+package records
+
+import (
+	"reflect"
+
+	"github.com/apache/arrow/go/v15/arrow"
+	"github.com/apache/arrow/go/v15/arrow/array"
+)
+
+type Reader[T any] struct {
+	records []arrow.Record
+}
+
+func NewReader[T any](records ...arrow.Record) *Reader[T] {
+	var a T
+	r := reflect.TypeOf(a)
+	for r.Kind() == reflect.Ptr {
+		r = r.Elem()
+	}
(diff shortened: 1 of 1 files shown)


thorfour (approved):
Would be nice if it handled repeated values as well.

metalmatze:
Agreed! 
This is merely the start of this helper. In my case, I only needed the `int64` support. :smile: 
Given it's used for testing, let's add whatever we need whenever we get there.

PR #806, opened 2024-03-28
pqarrow/arrowutils: Add EnsureSameSchema for records

metalmatze:
Sometimes, before using MergeRecords, we want to make sure that all records have the same schema. If they don't we need to add NullArrays for fields that don't exist. 

This is a modified version of what the OrderedSynchronizer already has.

### DIFF
--- a/pqarrow/arrowutils/schema.go
+++ b/pqarrow/arrowutils/schema.go
@@ -1,0 +1,95 @@
+package arrowutils
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/apache/arrow/go/v16/arrow"
+	"github.com/apache/arrow/go/v16/arrow/array"
+)
+
+// EnsureSameSchema ensures that all the records have the same schema. In cases
+// where the schema is not equal, virtual null columns are inserted in the
+// records with the missing column. When we have static schemas in the execution
+// engine, steps like these should be unnecessary.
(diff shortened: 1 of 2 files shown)


pqarrow/arrowutils/schema.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+	"sort"
+
+	"github.com/apache/arrow/go/v15/arrow"
+	"github.com/apache/arrow/go/v15/arrow/array"
+)
+
+// EnsureSameSchema ensures that all the records have the same schema. In cases
+// where the schema is not equal, virtual null columns are inserted in the
```
thorfour, line 12: Modifying the schema of an arrow record isn't supposed to be supported, and I think may even be removed in future releases of Arrow. 

Can we instead create new records with the unified schema instead?

metalmatze: Isn't this what this code is doing? 
We merged all columns both in the schemas and the arrays.

thorfour: Hmm okay I suppose it is, but it's modifying the slice records. Could you make it return a new slice instead?

metalmatze: Let me know if this is now closer to what you imagined.

thorfour: Great thanks!

thorfour (approved):

PR #807, opened 2024-04-04
Add proto CI runs

metalmatze:
The `.github/workflows/` are mostly copied from Parca.


metalmatze:
We expect the proto-pr to fail here. 
To satisfy the linter, I had to rename `ENCODING_UNKNOWN` to `ENCODING_UNSPECIFIED`, and now the other check is complaining about a breaking change.

PR #823, opened 2024-04-18
pqarrow/builder: Add OptFloat64Builder

metalmatze:
We mostly need this because it supports the Set and Add methods compared to the upstream Float64Builder.


asubiotto (approved):
I think it'd be nice to have some sort of generic test that verifies all the builders are correct, but in the absence of that test, this PR looks good.

metalmatze:
Agreed on the generic tests. :+1: 
I was already thinking the same when I posted the PR. I'll see what I can easily come up with before we merge this one.

metalmatze:
I'm going to merge this as the generic tests have been added. Let me know if we need to follow up on anything.

asubiotto:
Thanks!

PR #824, opened 2024-04-18
Iceberg storage

thorfour:
This adds a new storage implementation based on [Apache Iceberg](https://iceberg.apache.org)

The default storage for FrostDB performs a linear scan on all data files. This means that at minimum we have O(N) reads performed. Just to perform they query planning step.

Iceberg allows us to perform query planning down to the data file with O(1) reads. It extracts and groups metadata into a set of manifest files that allow us to perform fewer reads to determine which data files may contain useful data. 

This implementation currently performs data file pruning only at the manifest layer and does not yet support pruning at the manifest list layer. That will be a future addition once table partitioning is supported in the [iceberg-go](https://github.com/polarsignals/iceberg-go) library. However, being able to prune at the manifest layer should already be a marked improvement.

In addition to adding support for pruning at the manifest list level, future support could be added for utilities to prune/cleanup tables. As new snapshots are created, old snapshots and their metadata files are left around.


brancz (approved):
Very cool! lgtm

db_test.go
```diff
@@ -2962,3 +2963,86 @@ func Test_DB_SnapshotDuplicate_Corrupted(t *testing.T) {
...
+	validateRows(10)
+
+	require.NoError(t, table.RotateBlock(ctx, table.ActiveBlock(), false))
+	require.Eventually(t, func() bool {
+		info, err := bucket.Attributes(ctx, filepath.Join("test", "test", "metadata", "v1.metadata.json"))
+		if err != nil {
+			return false
+		}
+		return info.Size > 0
+	}, 30*time.Second, 100*time.Millisecond)
```
asubiotto, line 3025: This feels super slow for a unit test. Can we check in intervals of `10*time.Millisecond` with max time of `5*time.Second`?

asubiotto (approved):

asubiotto:
Also what is the long-term plan for our iceberg-go fork. Are we eventually going to upstream the changes? Seems like there's some stuff that's on the roadmap that might be interesting for us from a cursory view (e.g. scan planning)

thorfour:
> Also what is the long-term plan for our iceberg-go fork. Are we eventually going to upstream the changes? Seems like there's some stuff that's on the roadmap that might be interesting for us from a cursory view (e.g. scan planning)

In the medium term we'll definitely continue our fork as right now the Apache repo doesn't seem to be under active development ( most recent commit was 2 months ago). And ours has diverged quite a bit already. 

Long term yes it would be nice if the Apache repo implemented a lot of the things we need but no idea what that timeline looks like.

db_test.go
thorfour: Haha I'm pretty sure GitHub copilot picked those settings. I'll tune them down

PR #865, opened 2024-05-16
dst: add deterministic simulation test

asubiotto:
Running this test by itself is just a normal integration test. To run this test
deterministically use POLARSIGNALS_RANDOM_SEED, the WASM build target, and
github.com/asubiotto/go. For now, only the test is checked in to reduce the
change surface area and is skipped if POLARSIGNALS_RANDOM_SEED is not
specified.

There are still a couple of bugs to fix (#864 is the first PR here, and one of the commits in this PR contains another fix), so adding the running of this test to CI is still future work.

The idea is to merge this test as-is even though it is not final to provide a base to build on and fix bugs.


dst/dst_test.go
```diff
@@ -0,0 +1,547 @@
...
+				}
+				return nil
+			})
+		case restart:
+			// This is a hack to ensure some goroutines are scheduled before
+			// this restart.
+			// TODO(asubiotto): Figure out if we still need this.
+			time.Sleep(1 * time.Millisecond)
+			// Graceful shutdown.
+			require.NoError(t, c.Close())
```
thorfour, line 419: Would be nice to have an ungraceful shutdown as well

thorfour (approved):
LGTM once go build is happy

dst/dst_test.go
asubiotto: Yep! Definitely want to add that.

asubiotto:
The failure was a test in the `wal` package that assumed the previous (buggy) `nextTx = truncateTx + 1` behavior. I'm going to remove that fix in this PR to preserve the old behavior and look into the more comprehensive fix/testing changes as a followup.

PR #866, opened 2024-05-17
wal: correctly set nextTx on Truncate

asubiotto:
On Truncate(k), if k was logged to the WAL the behavior is unchanged. However, if k had not yet been logged to the WAL, the WAL performed a full reset of the WAL and set the next expected txn to k+1. This is incorrect, since the semantics of truncate are that k becomes the first entry in the WAL after truncation.


asubiotto:
This was found by DST and fixes a data loss bug (previously `k` would've been incorrectly dropped after a snapshot)

metalmatze (approved):
Interesting and good find!

thorfour (approved):

PR #878, opened 2024-05-22
Converter read all pages

thorfour:
Discovered some good performance gains during query time of Parquet files that if we read in all the Parquet values into the buffer at once instead of page by page it decreases query times.

```bash
thor@thors-MacBook-Pro ~/.../github.com/polarsignals/frostdb % benchstat before.txt after.txt
name                                        old time/op    new time/op    delta
_ParquetAggregation/paruqet_aggregation-10    16.0ms ± 0%     9.3ms ± 1%  -41.80%  (p=0.000 n=9+10)

name                                        old alloc/op   new alloc/op   delta
_ParquetAggregation/paruqet_aggregation-10    33.0MB ± 0%    37.9MB ± 0%  +14.99%  (p=0.000 n=9+10)

name                                        old allocs/op  new allocs/op  delta
_ParquetAggregation/paruqet_aggregation-10     1.42k ± 0%     1.40k ± 0%   -1.88%  (p=0.000 n=9+10)
```

Slightly more data allocated (due to the larger buffer) but I think it's a worthwhile trade off

### DIFF
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -480,7 +480,7 @@ func Benchmark_ParquetAggregation(b *testing.B) {
 	b.ResetTimer()
 	b.Run("paruqet_aggregation", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
-			engine.ScanTable("test").
+			require.NoError(b, engine.ScanTable("test").
 				Filter(
 					logicalplan.Col("timestamp").Gt(logicalplan.Literal(10)),
 				).
@@ -493,7 +493,7 @@ func Benchmark_ParquetAggregation(b *testing.B) {
 					},
 				).Execute(ctx, func(ctx context.Context, r arrow.Record) error {
 				return nil
-			})
+			}))
 		}
 	})
 }
--- a/aggregate_test.go
+++ b/aggregate_test.go
@@ -433,7 +433,7 @@ func BenchmarkAggregation(b *testing.B) {
 	}
 }
 
-// Benchmark_ParquetAggregation is similar to BenchmarkAggregation but uses uses Parquet as the storage format.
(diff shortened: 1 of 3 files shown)


pqarrow/arrow.go
```diff
@@ -796,21 +796,19 @@ func (c *ParquetConverter) writeColumnToArray(
...
+			if n := columnChunk.NumValues(); int64(cap(c.scratchValues)) < n {
 				c.scratchValues = make([]parquet.Value, n)
 			} else {
 				c.scratchValues = c.scratchValues[:n]
 			}
-
-			// We're reading all values in the page so we always expect an io.EOF.
 			reader := p.Values()
-			if _, err := reader.ReadValues(c.scratchValues); err != nil && err != io.EOF {
+			i, err = reader.ReadValues(c.scratchValues[i:])
```
asubiotto, line 805: Do we want to limit the buffering somehow?

aggregate_test.go
```diff
@@ -433,6 +433,71 @@ func BenchmarkAggregation(b *testing.B) {
 	}
 }
 
+// Benchmark_ParquetAggregation is similar to BenchmarkAggregation but uses uses Parquet as the storage format.
```
asubiotto, line 436: nit: double uses

thorfour: damnit copilot

pqarrow/arrow.go
thorfour: We didn't before but something to consider in the future potentiall

thorfour: But there was no limit to how many values were in the page just like with row groups so it was a virtual limit not one that we set.

thorfour:
Ooof I found that there was a bug in this change where if we had more than two pages the offset i was incorrect and we started to overwrite pages. The different when that's fixed is minimal. Going to close

PR #883, opened 2024-05-29
workflows: add DST workflow

asubiotto:


asubiotto:
Some things to note:
- DST is currently set to run for 10 iterations and takes 1 minute to trun. The first iteration takes around 30s, so I'm not sure reducing this number will affect runtime that much. We might need to investigate why the first run takes so long but for now this is fine.

brancz (approved):

PR #884, opened 2024-05-29
Reservoir sampler

thorfour:
Adds a new operator to the query engine to perform random sampling of the rows.

Initially this was implemented in the physical conversion layer but it required us to push down the filter into the physical layer which created a significant amount of overhead as filtering on parquet values is slow. That approach has been abandoned for now but can be revisited in the future.


query/logicalplan/builder.go
```diff
@@ -237,6 +237,22 @@ func resolveAggregation(plan *LogicalPlan, agg *AggregationFunction) ([]*Aggrega
 	}
 }
 
+func (b Builder) Sample(expr Expr) Builder {
```
brancz, line 240: We should make the type of sampler used configurable, while we currently just need the reservoir sampler, there can certainly be valid use cases for others.

query/physicalplan/sampler.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// sample implements the reservoir sampling algorithm found https://en.wikipedia.org/wiki/Reservoir_sampling.
+func (s *ReservoirSampler) sample(r arrow.Record) {
+	n := s.n + r.NumRows()
+	for i := float64(s.n); i < float64(n); {
+		i += math.Floor(math.Log(rand.Float64())/math.Log(1-s.w)) + 1
+		if i <= float64(n) {
+			// replace a random item of the reservoir with row i
+			s.reservoir[rand.Intn(int(s.size))] = r.NewSlice(int64(i)-s.n-1, int64(i)-s.n)
```
brancz, line 101: Using rand means a query over the same data is unlikely to return the same data?

query/physicalplan/sampler.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// fill will fill the reservoir with the first size records.
+func (s *ReservoirSampler) fill(r arrow.Record) arrow.Record {
+	if s.currentSize == s.size {
+		return r
+	}
+
+	if s.currentSize+r.NumRows() <= s.size { // The record fits in the reservoir
+		for i := int64(0); i < r.NumRows(); i++ { // For simplicity of implementation the reservoir is by row; This is probably not optimal
```
brancz, line 76: This concerns me scalability-wise, we're turning the columnar database into a row-based one with this.

brancz (commented):
Generally the implementation looks fine (except for the scalability concerns).

I'm unconvinced this will work for the purpose we want it for, both the use of randomness, as well as the unknown order of records passed through the sampler means that we cannot guarantee that the result of running the same query twice will return the same data.

brancz:
So while there are other ways to achieve this, the requirements in Parca/Polar Siganls Cloud that must continue to hold true are:

1) sharing a link must result in users seeing exactly the same result
2) data across visualizations must be consistent (eg. viewing data in an icicle graph must show the same values as tables)
3) the binary names returned from the new metadata call must be consistent with what's shown in the icicle graph

The only way I can see this continuing to be true would be if we cache results from a query indefinitely, and only render reports based on the cached result.

query/physicalplan/sampler.go
thorfour: Agreed, I think it definitely needs to change to not slice records during the initial fill at all.

query/physicalplan/sampler.go
thorfour: Correct, we're punting on the determinism for now since there's other issues that even using the same rand seed would break determinism. i.e Scan is concurrent and there's no guarantee of ordering right now.

If we can fix the non-determinism of scan than we can revisit adding a rand seed that's injected.

query/logicalplan/builder.go
thorfour: Yea I waffled on that since we didn't have a use case for it today I decided to punt on it and not add complexity yet.

thorfour:
```bash
thor@thors-MacBook-Pro ~/.../github.com/polarsignals/frostdb % benchstat before.txt after.txt
name                old time/op    new time/op    delta
Aggregation/sum-10     392µs ± 0%     390µs ± 0%  -0.40%  (p=0.000 n=9+9)

name                old alloc/op   new alloc/op   delta
Aggregation/sum-10     402kB ± 0%     403kB ± 0%  +0.27%  (p=0.000 n=8+9)

name                old allocs/op  new allocs/op  delta
Aggregation/sum-10       688 ± 0%       707 ± 0%  +2.72%  (p=0.000 n=10+10)
```

Aggregation benchmarks when sampling with a reservoir that is > total rows. So there should be little impact to queries that don't need to be sampled.

PR #895, opened 2024-06-05
dst: add hardRestart command

asubiotto:
This simulates a hard restart by calling the vfs_shutdown hook on the wasm
host module to error out all filesystem calls.

> [!NOTE]
> This PR is currently blocked on the resolution of #916, a bug that was found with DST but requires a more fundamental rearchitecture of the WAL.

Bugs found:
- [x] #914 
- [ ] #916

### DIFF
--- a/dst/dst_test.go
+++ b/dst/dst_test.go
@@ -52,7 +52,8 @@ const (
 	compact
 	snapshot
 	rotate
-	restart
+	gracefulRestart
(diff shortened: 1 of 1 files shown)


asubiotto:
I'm not very happy with this PR since it's not a complete solution given open FDs. Will keep this up while I see how we can simulate wasi syscalls.

PR #901, opened 2024-06-12
table: add waitgroup to optionally wait for block persistence

asubiotto:


table.go
```diff
@@ -605,12 +612,18 @@ func (t *Table) writeBlock(block *TableBlock, nextTxn uint64, skipPersist, snaps
 	}
 }
 
-func (t *Table) RotateBlock(_ context.Context, block *TableBlock, skipPersist bool) error {
+// RotateBlock rotates the given block and persists it asynchronously if
+// requested. If wg is non-nil, wg.Done will be called once the block has been
+// persisted.
+func (t *Table) RotateBlock(_ context.Context, block *TableBlock, skipPersist bool, wg *sync.WaitGroup) error {
```
thorfour, line 618: Feels unfortunate to have to pass nil everywhere. Could we instead either make this an option, or have the RotateBlock function return the waitgroup itself so callers can decide if they'd like to wait for it?

brancz (approved):

table.go
asubiotto: Yep, can make these options

asubiotto:
Made them options. PTAL

PR #904, opened 2024-06-18
*: add reusableRegistry

### LINKED ISSUE #716: panic: Duplicate registration
In a couple places in `db.go` we call `newTable(...)` which registers metrics under that table's name.

But before adding that table into the database we call `newTableBlock(...)` which can return an error. Which then leaves us in a state where that table isn't in the database but it's metrics have been registered and on the next attempt to create that table it panics.

https://github.com/polarsignals/frostdb/blob/ab6b4914523d5f1961c294f14e3af28ebed13376/db.go#L818-L836

We should prevent this scenario from happening. Either somehow call `newTableBlock` before `newTable` or split metrics registration out and call it after.

asubiotto:
This registry allows the re-registration of metrics when dropping and recreating DBs and tables. In this case, we do not want to panic and simply want to continue exporting the same metrics.

Fixes #716

### DIFF
--- a/db.go
+++ b/db.go
@@ -112,6 +112,11 @@ func New(
 		}
 	}
 
+	// Wrap the registry in a reusable registry to allow metrics to be
+	// registered multiple times in cases e.g. where a database is dropped and
+	// subsequently recreated, or a table is dropped and recreated.
+	s.reg = newReusableRegistry(s.reg)
+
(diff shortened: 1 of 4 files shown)


asubiotto:
Closing in favor of an upcoming PR that registers metrics at the store level.

PR #905, opened 2024-06-19
*: move all metrics registration to the store level

asubiotto:
At component instantiation time (e.g. db, wal, table), existing metrics are
provided to the components with the correct variable label set. This moves the
metric registration lifecycle to the store level and avoids duplicate
registration panics when e.g. reopening a closed db.


brancz (approved):

asubiotto:
Tested on local cloud and prometheus and verified some metrics are exported with the correct names and labels.

PR #906, opened 2024-06-20
*: remove read-only tables

asubiotto:
These were broken and mostly unused. Reading tables that only exist on object storage is now supported by using the WithProvideIfNotExists option on the DBTableProvider at query time.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -2,7 +2,6 @@ package frostdb
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"io/fs"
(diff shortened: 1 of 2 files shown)


db_test.go
```diff
@@ -537,141 +537,6 @@ func Test_DB_Filter_Block(t *testing.T) {
...
-	if e.delete != nil {
-		return e.delete(ctx, name)
-	}
-
-	return nil
-}
-
-func (e *ErrorBucket) Name() string { return "error bucket" }
-
-func Test_DB_OpenError(t *testing.T) {
```
asubiotto, line 641: @thorfour seems like this test was not doing much so I removed it. Is that ok?

thorfour: It was validating that if the DB encountered an error when opening that it didn't panic and that we could successfully open the DB in subsequent attempts. 

Why remove it?

thorfour (commented):
I don't care for this change as it makes FrostDB more frustrating to use. Now a consumer of the database has to know about an option to correctly open a database that should be in read-only mode. I think FrostDB should just do the right thing here. 

I think this is the wrong way to address read-only tables.

PR #910, opened 2024-06-20
vfs: add a VFS wrapper used in deterministic simulation tests

asubiotto:
This currently exposes the capability to simulate a hard shutdown. Given that this currently causes test failures, it is unused in this PR. A follow up PR will be created to use these hooks and fix the fallout.


asubiotto:
I'm going to go ahead and merge this without a review since it's a testing change

PR #915, opened 2024-06-25
[WIP/DNM] table: fix behavior on new table block error

asubiotto:
Table block creation can fail for a variety of reasons, but we were only
handling the happy case. This commit only overwrites the active block if the
new table block was successfully created, and logs a noop record to the WAL on
error to not block subsequent wal records logs.

Bug found with DST.


metalmatze (approved):
Once more very interesting. :+1:

